Issue #243, opened 2014-07-29
Refactor Logging


oberstet:
This depends on issue #242 
1. Add a new method `LoopMixin.log(msg)` to loop mixin.
2. Cleanup:
   - All code in `autobahn/twisted/*` uses Twisted logging
   - All code in `autobahn/asyncio/*` uses Python standard logging
   - All code in `test` and `autobahn/wamp1` folders remains untouched
   - All other code uses `LoopMixin.log`

wolfhechel:
When is this marked to be released?

kampka:
@wolfhechel Sorry, but AutobahnPython has changed so much since my patch that it can no longer be applied. I don't now if or when I will get back to this, so if want to see it merged, feel free to rebase the patch against a current version of AutobahnPython, minding @oberstet's comments on #303.

oberstet:
this depends on https://github.com/tavendo/txaio/issues/8

oberstet:
we have that now: via txaio. shiny, full featured, structured logging. used in Autobahn and Crossbar.io

Issue #242, opened 2014-07-29
Refactor FutureMixin


oberstet:
https://github.com/tavendo/AutobahnPython/blob/master/autobahn/autobahn/asyncio/wamp.py#L52
https://github.com/tavendo/AutobahnPython/blob/master/autobahn/autobahn/twisted/wamp.py#L50

1. Rename class to `LoopMixin`
2. Move to `autobahn.twisted.util` / `autobahn.asyncio.util`
3. Make static methods public
4. Add proper docstrings

meejah:
This can be closed I think since the move to https://github.com/tavendo/txaio

PR #303, opened 2014-12-13
Refactor logging

kampka:
This commit resolves the outstanding issues #242 and #243.
I strayed a little from the proposed solution in order to improve the possibilities of the API and leverage the control provided by the logging package.

### DIFF
--- a/autobahn/autobahn/asyncio/log.py
+++ b/autobahn/autobahn/asyncio/log.py
@@ -1,0 +1,48 @@
+###############################################################################
+#
+#  Copyright (c) 2014 Christian Kampka <christian@kampka.net>
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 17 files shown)


oberstet:
The PR can't be merged anymore, since of the recent refactoring of all routing code from AutobahnPython into Crossbar.io - I apologize for not getting to this PR before.

The change seems to address 2 things at once: `LoopMixin` and logging. Can we make that into 2 separate changes?

That is, create 2 new files:

* `autobahn/twisted/loop.py` with a single class `LoopMixin` having e.g. `LoopMixin.reject_future`
* `autobahn/asyncio/loop.py` with a single class `LoopMixin` having e.g. `LoopMixin.reject_future`

plus the changes to the rest of the library that result.

We will have to synchronize this change with Crossbar.io code also (as the routing code now lives there).

PR #329, opened 2015-02-06
_call_later method

Paranaix:
This pull request adds a _call_later method to the FutureMixin which might be required for proper timeout handling. 

Currently those methods are no class methods because they have to access either a reactor or an ioloop. Making  While those attributes could also be passed in the function I decided that this is would not be a nice solution as they might shadow parameters and make the invocation less intuitive and more verbose.

Please feel free to find a better solution and let me know. If you think that the current one is acceptable, I would propose to make all the other static functions in FutureMixin regular methods as well for the sake of consistency. In case we go with this solution we would also need a way to pass a reactor/loop to an ApplicationSession then.

### DIFF
--- a/autobahn/autobahn/asyncio/wamp.py
+++ b/autobahn/autobahn/asyncio/wamp.py
@@ -92,6 +92,22 @@ class FutureMixin:
    def _gather_futures(futures, consume_exceptions = True):
       return asyncio.gather(*futures, return_exceptions = consume_exceptions)
 
+   def _call_later(self, delay, callback, *args, **kwargs):
+      if self._loop:
+         loop = self._loop
+      else:
+         loop = asyncio.get_event_loop()
+
+      f = Future()
+
+      def wrapper():
+         try:
+            result = callback(*args, **kwargs)
+            f.set_result(result)
+         except Exception as e:
+            f.set_exception(e)
+
+      return (loop.call_later(delay, wrapper), f)
 
 
 class ApplicationSession(FutureMixin, protocol.ApplicationSession):
--- a/autobahn/autobahn/twisted/wamp.py
+++ b/autobahn/autobahn/twisted/wamp.py
@@ -71,6 +71,22 @@ class FutureMixin:
    def _gather_futures(futures, consume_exceptions = True):
       return DeferredList(futures, consumeErrors = consume_exceptions)
 
+   def _call_later(self, delay, callback, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


oberstet:
First, yes, this should be part of the mixing. And yes, it should be possible to provide the loop/reactor to the mixin.

With the specifics of call_later: what if the callback produces a deferred itself? What about cancelling?

Also, we already have it:

https://github.com/tavendo/AutobahnPython/blob/master/autobahn/autobahn/twisted/websocket.py#L210
https://github.com/tavendo/AutobahnPython/blob/master/autobahn/autobahn/asyncio/websocket.py#L232

Paranaix:
I will fix the issues mentioned by you. Cancelling should actually already be possible (notice that the methods return tuples), but wrapping this as a cleaner interface might not be a bad idea.

Regarding the factories:  I didn't know about them but I think this is an excellent example of what I deem very wrong with Autobahn and Crossbar (no offense intended). Based on your comment, I believe you would write code like this `self._factory._calllater(....)` if you would ever need that method inside a session in the future. I encountered such lines very frequently as I crawled through your code. Basically everything is interconnected this way. You have tons of classes (nearly too many in my opinion) but those are not clearly separated from each other. This is not good code in my humble opinion. As I fixed https://github.com/crossbario/crossbar/pull/235 it took me I believe 1-2h just to find where that session came from. The point that I didn't know about `_calllater` illustrates this issue as well. Why would somebody expect a factory to contain such a method besides that it also makes the method non re-useable? In my opinion its absolutely in the wrong place.

Please don't take this criticism too harsh or personal but rather as a honest insight from somebody who isn't familiar with the code in all the details. I know that this was only slightly relevant to this issue, but that was a thing I wanted to state for a long time and the mentioned classes illustrated it pretty well in my opinion.

oberstet:
In general, I agree about too much coupling inside the code base.

To my defense: you have to take into account: history (e.g. we managed to inject asyncio support _after_ the code base already existed), required support for many features (e.g. Crossbar.io supports 18 pluggable WAMP transports), support for a fuzzing WebSocket testsuite (AutobahnTestsuite), etc

The issue is more pressing with Crossbar.io than AutobahnPython. The latter has clear interfaces for WAMP internal stuff using ABCs.

oberstet:
One more note: I am currently running autopep8 .. it'll make PRs unmergable. https://github.com/tavendo/AutobahnPython/issues/331

PR #339, opened 2015-02-12
support multiple subscription for the same topic

tardyp:
fixes https://github.com/tavendo/AutobahnPython/issues/274

### DIFF
--- a/autobahn/autobahn/wamp/protocol.py
+++ b/autobahn/autobahn/wamp/protocol.py
@@ -21,6 +21,7 @@ from __future__ import absolute_import
 import inspect
 import six
 from six import StringIO
+import weakref
 
 from autobahn.wamp.interfaces import ISession, \
     IPublication, \
@@ -61,11 +62,32 @@ class Handler:
     """
     """
 
-    def __init__(self, obj, fn, topic, details_arg=None):
-        self.obj = obj
-        self.fn = fn
+    def __init__(self, session, subscriptionid, topic, details_arg=None):
+        self.session = weakref.proxy(session)
+        self.id = subscriptionid
         self.topic = topic
         self.details_arg = details_arg
+        self.subscriptions = []
+
+    def callWithMessage(self, msg):
+        for s in self.subscriptions:
+            s._callWithMessage(msg)
+
+    def subscribe(self, obj, fn):
+        s = Subscription(self, obj, fn)
+        self.subscriptions.append(s)
+        return s
+
+    def unsubscribe(self, subscription):
+        self.subscriptions.remove(subscription)
+        if not self.subscriptions:
+            return self.unsubscribeSession()
+
+    def unsubscribeSession(self):
+        """
+        Implements :func:`autobahn.wamp.interfaces.ISubscription.unsubscribe`
+        """
+        return self.session._unsubscribe(self)
 
 
 class Publication:
@@ -79,22 +101,53 @@ class Publication:
 
 IPublication.register(Publication)
 
-
 class Subscription:
     """
     Object representing a subscription.
+    Two methods of a same component can subscribe twice to the same topic
     This class implements :class:`autobahn.wamp.interfaces.ISubscription`.
     """
-    def __init__(self, session, subscriptionId):
-        self._session = session
-        self.active = True
-        self.id = subscriptionId
+
+    def __init__(self, handler, obj, fn):
+        self.obj = obj
+        self.fn = fn
+        self.handler = weakref.proxy(handler)
+
+    def _callWithMessage(self, msg):
+        try:
+            if self.obj:
+                if msg.kwargs:
+                    if msg.args:
+                        self.fn(
+                            self.obj, *msg.args, **msg.kwargs)
+                    else:
+                        self.fn(self.obj, **msg.kwargs)
+                else:
+                    if msg.args:
+                        self.fn(self.obj, *msg.args)
+                    else:
+                        self.fn(self.obj)
+            else:
+                if msg.kwargs:
+                    if msg.args:
+                        self.fn(*msg.args, **msg.kwargs)
+                    else:
+                        self.fn(**msg.kwargs)
+                else:
+                    if msg.args:
+                        self.fn(*msg.args)
+                    else:
+                        self.fn()
+
+        except Exception as e:
+            log.error(e, "While firing event handler {0} subscribed under '{1}' ({2}): {3}".format(
+                      self.fn, self.handler.topic, msg.subscription, e))
 
     def unsubscribe(self):
         """
         Implements :func:`autobahn.wamp.interfaces.ISubscription.unsubscribe`
         """
-        return self._session._unsubscribe(self)
+        return self.handler.unsubscribe(self)
 
 
 ISubscription.register(Subscription)
@@ -432,34 +485,7 @@ class ApplicationSession(BaseSession):
                             msg.kwargs = {}
                         msg.kwargs[handler.details_arg] = types.EventDetails(publication=msg.publication, publisher=msg.publisher)
 
-                    try:
-                        if handler.obj:
-                            if msg.kwargs:
-                                if msg.args:
-                                    handler.fn(handler.obj, *msg.args, **msg.kwargs)
-                                else:
-                                    handler.fn(handler.obj, **msg.kwargs)
-                            else:
-                                if msg.args:
-                                    handler.fn(handler.obj, *msg.args)
-                                else:
-                                    handler.fn(handler.obj)
-                        else:
-                            if msg.kwargs:
-                                if msg.args:
-                                    handler.fn(*msg.args, **msg.kwargs)
-                                else:
-                                    handler.fn(**msg.kwargs)
-                            else:
-                                if msg.args:
-                                    handler.fn(*msg.args)
-                                else:
-                                    handler.fn()
-
-                    except Exception as e:
-                        if self.debug_app:
-                            print("Failure while firing event handler {0} subscribed under '{1}' ({2}): {3}".format(handler.fn, handler.topic, msg.subscription, e))
-
+                    handler.callWithMessage(msg)
                 else:
                     raise ProtocolError("EVENT received for non-subscribed subscription ID {0}".format(msg.subscription))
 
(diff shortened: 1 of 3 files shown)


oberstet:
This one is worth some discussion first I think. There are multiple aspects:

1. A client might want to attach multiple handlers on a single subscription object retrieved via one call to `ApplicationSession.subscribe`. A client might also want to just call `ApplicationSession.subscribe` multiple times (having the handler transparently being attached to the already existing subscription). Do we want both or only one (and which one)?

2. With the second approach in 1., we can't simply check a "by topic map" if a subscription is already there and skip talking to the router since: the subscription can be done with options, like prefix-match. The topic string will be identical, the subscription returned by the router different. We don't want to impose the burden on all client implementations to implement some involved logic.

3. What if the handler function provided is exactly the same in both situations above?

4. AutobahnJS does implement the second approach from 1. in above
    * https://github.com/tavendo/AutobahnJS/blob/master/package/lib/session.js#L243
    * https://github.com/tavendo/AutobahnJS/blob/master/package/lib/session.js#L1149

goeddea:
I think the first variant is much less important than the second one. With the first, if I there's a single place in the code where I attach the multiple handlers, I can do so via a single attached handler which passes on the received data to the other handlers. 

I doubt there are that many cases where I want to attach multiple handlers at different times in my code and then accept that a single unsubscribe anywhere stops them all receiving events.

What I do want to do is to be able to subscribe from anywhere in my code without having to worry about other subscriptions already existing, and similarly cancel subscriptions without having to worry about side effects elsewhere in the code. I think the way AutobahnJS implements this is good from the user view.

see also this [discussion on the WAMP mailing list](https://groups.google.com/forum/#!topic/wampws/kC878Ngc9Z0)

tardyp:
Hi,
My implementation is implementing something very similar to the JS version, and compatible with @goeddea 
The only difference is that the Handler class is acting as the list of subscription that the JS version has.

So the Handler class now handle both aspects of managing subscribers, and calling the functions/methods.
Maybe we should just rename Handler to SessionSubscription

When using ApplicationSession.subscribe, a list of ISubscription object is returned, one for each subscripted method.
We could wrap this list in another ISubscription object than just unregisters all the Subscription objects.

What this implementation does not do is to avoid the registration call to the router. The router will send the same registration id as before, *and then* we will add one subscription to the handler.
Avoiding this call requires to properly map handlers per topic/options rather than just registrationid.

oberstet:
> Avoiding this call requires to properly map handlers per topic/options rather than just registrationid.

Which is a pain, and the reason we do allow to send another `SUBSCRIBE`  message to the router, and have the router only return the (already existing) subscription ID.

What the PR does is fine in principle, and we will add this feature.

tardyp:
> What the PR does is fine in principle, and we will add this feature.

BTW, I can work on this, and fix my proposal to fit your design principles, just tell what's wrong..

oberstet:
Implemented in trunk now. E.g. consider the following test program:

```python
from twisted.internet.defer import inlineCallbacks

from autobahn.wamp.types import PublishOptions, SubscribeOptions
from autobahn.twisted.util import sleep


class MyComponent(ApplicationSession):

    @inlineCallbacks
    def onJoin(self, details):
        print("session ready")

        self.received = []

        topic1 = u"com.example.topic1"
        sub_options = SubscribeOptions(details_arg="details")
        pub_options = PublishOptions(acknowledge=True, exclude_me=False)
        delay = 0.1

        # subscribe handler1 on topic
        #
        def handler1(value, details=None):
            print("handler1 got event")
            self.received.append(("handler1", value, details.publication))

        sub1 = yield self.subscribe(handler1, topic1, options=sub_options)
        print("handler1 subscribed: {}".format(sub1), sub1)

        res = yield self.publish(topic1, 1, options=pub_options)
        print("event 1 published: {}".format(res))
        yield sleep(delay)

        # subscribe handler2 on same topic
        #
        def handler2(value, details=None):
            print("handler2 got event")
            self.received.append(("handler2", value, details.publication))

        sub2 = yield self.subscribe(handler2, topic1, options=sub_options)
        print("handler2 subscribed: {}".format(sub2), sub2)

        res = yield self.publish(topic1, 2, options=pub_options)
        print("event 2 published: {}".format(res))
        yield sleep(delay)

        # subscribe handler2 on same topic a second time
        #
        sub2b = yield self.subscribe(handler2, topic1, options=sub_options)
        print("handler2 subscribed 2nd: {}".format(sub2b), sub2b)

        res = yield self.publish(topic1, 3, options=pub_options)
        print("event 3 published: {}".format(res))
        yield sleep(delay)

        # unsubscribe subscription1
        #
        yield sub1.unsubscribe()
        print("handler1 unsubscribed: {}".format(sub1), sub1)

        res = yield self.publish(topic1, 4, options=pub_options)
        print("event 4 published: {}".format(res))
        yield sleep(delay)

        # unsubscribe subscription2
        #
        yield sub2.unsubscribe()
        print("handler2 unsubscribed: {}".format(sub2), sub2)

        res = yield self.publish(topic1, 5, options=pub_options)
        print("event 5 published: {}".format(res))
        yield sleep(delay)

        # unsubscribe subscription2b
        #
        yield sub2b.unsubscribe()
        print("handler2 unsubscribed 2nd: {}".format(sub2b), sub2b)

        res = yield self.publish(topic1, 6, options=pub_options)
        print("event 6 published: {}".format(res))
        yield sleep(delay)

        print("Done!", self.received)
        self.leave()

    def onLeave(self, details):
        print("session left")
        self.disconnect()

    def onDisconnect(self):
        print("transport disconnected")
        reactor.stop()


if __name__ == '__main__':
    runner = ApplicationRunner(url=u"ws://localhost:8080/ws", realm=u"realm1",
        debug=False, debug_wamp=False)
    runner.run(MyComponent)
```

Here is the output:

```console
(python279_1)oberstet@thinkpad-t430s:~/scm/crossbar/crossbarexamples/metaapi$ python client.py 
session ready
('handler1 subscribed: Subscription(id=1689659226, is_active=True)', <autobahn.wamp.protocol.Subscription object at 0x7f3ff5d779d0>)
event 1 published: Publication(id=1244413349)
handler1 got event
('handler2 subscribed: Subscription(id=1689659226, is_active=True)', <autobahn.wamp.protocol.Subscription object at 0x7f3ff5bdf490>)
event 2 published: Publication(id=1999161274)
handler1 got event
handler2 got event
('handler2 subscribed 2nd: Subscription(id=1689659226, is_active=True)', <autobahn.wamp.protocol.Subscription object at 0x7f3ff5bdf350>)
event 3 published: Publication(id=1452547618)
handler1 got event
handler2 got event
handler2 got event
('handler1 unsubscribed: Subscription(id=1689659226, is_active=False)', <autobahn.wamp.protocol.Subscription object at 0x7f3ff5d779d0>)
event 4 published: Publication(id=448187591)
handler2 got event
handler2 got event
('handler2 unsubscribed: Subscription(id=1689659226, is_active=False)', <autobahn.wamp.protocol.Subscription object at 0x7f3ff5bdf490>)
event 5 published: Publication(id=238286570)
handler2 got event
('handler2 unsubscribed 2nd: Subscription(id=1689659226, is_active=False)', <autobahn.wamp.protocol.Subscription object at 0x7f3ff5bdf350>)
event 6 published: Publication(id=592787387)
('Done!', [('handler1', 1, 1244413349), ('handler1', 2, 1999161274), ('handler2', 2, 1999161274), ('handler1', 3, 1452547618), ('handler2', 3, 1452547618), ('handler2', 3, 1452547618), ('handler2', 4, 448187591), ('handler2', 4, 448187591), ('handler2', 5, 238286570)])
session left
transport disconnected
```

oberstet:
@tardyp apart from automatic reconnection, are there more features you need?

tardyp:
Ok thanks, I'll try to see how it works with buildbot.

What I have been struggling on lately is how to catch properly disconnection from slaves. I understand now this can be done safely with latest crossbar.io using wamp.* namespace.
I need to make some experiement with this API.

Also, what we may need with wamp is the possibility to have persistent named queues, like with amqp.

And as a last feedback, I think there is a need for a status table to have clear information of what features are implemented in autobahn and crossbar, compared to the spec.
I spend too much time debugging stuff based on the spec, to find out that this is actually not yet implemented.
Its ok not to have everything implemented yet, but you should tell it to your users.

oberstet:
Rgd. detecting buildslaves (connected over WAN) disconnect. There are 2 important aspects:

* reliably detecting loss of TCP at network level .. it needs ping/pong .. which Crossbar.io implements (see WebSocket auto_ping_* options here http://crossbar.io/docs/WebSocket-Options/)
* you can hook into session lifecycle using WAMP meta API http://crossbar.io/docs/Session-Metaevents-and-Procedures/ => on_join/on_leave event)

Regarding features: absolutely. https://github.com/crossbario/crossbarwww/issues/50

>  persistent named queues

Could you expand a little on your use case / need within buildbot?

tardyp:
> Could you expand a little on your use case / need within buildbot?

We have schedulers and change sources in buildbot. Each of them is run only
once in the cluster of masters.
changesource are connected to the version control server, and generate
change events to the mq.
Schedulers watch the change events, and generate build request
appropriately.

If a master is stopped, then another master will take over for its
schedulers.
There is a period here where some events might be lost so build wouldn't be
triggered, so we figured out that this is a usecase for persistent queues.
Okok, this is a little bit far fetched. I think we can leave with it,
especially that this does not cover the change source disappearing,
@djmitche, do I understand well the use case for persistent queues?



Le mer. 11 mars 2015 à 19:22, Tobias Oberstein <notifications@github.com> a
écrit :

djmitche:
Yes.  I don't think it's a particularly unusual case: we have an event stream (changes) with multiple consumers (schedulers) - basically pub/sub, except that each consumer wants to be confident that it has received all events, even in the face of consumer downtime.  This is a cloud architecture, so downtime is common.

oberstet:
We don't want to have a queue per subscriber - transient or persistent. And we don't want transactional event processing, where a subscriber can consume events from queues within transactions. Reasons are mainly related to scalability.

We might have "event history" per topic (or subscription), even with persistence (surviving router restarts). With event history, a client can actively retrieve (limited) history of events dispatched on a topic (or generall subscription).

Hence, when subscriber looses it's connection, it can reconnect, resubscribe and retrieve any "missed" event history. The burder of remembering which events the client has already processed will be on the client. The latter is a consequence of avoiding "per client queues".

A rough sketch of the meta API from the WAMP advanced profile: https://github.com/tavendo/WAMP/blob/master/spec/advanced.md#event-history

Would that work for you?

PR #340, opened 2015-02-12
ReconnectingWampWebSocketClientFactory

tardyp:
Automatic reconnection in case of failure

fixes: https://github.com/tavendo/AutobahnPython/issues/328

### DIFF
--- a/autobahn/autobahn/twisted/wamp.py
+++ b/autobahn/autobahn/twisted/wamp.py
@@ -28,7 +28,6 @@ from twisted.internet.defer import Deferred, \
     DeferredList, \
     inlineCallbacks
 from twisted.internet.protocol import ReconnectingClientFactory
-from twisted.internet.error import ConnectionDone
 
 from autobahn.wamp import protocol
 from autobahn.wamp.types import ComponentConfig
@@ -95,8 +94,7 @@ class ApplicationSessionFactory(FutureMixin, protocol.ApplicationSessionFactory)
 class ReconnectingWampWebSocketClientFactory(WampWebSocketClientFactory, ReconnectingClientFactory):
 
     def clientConnectionFailed(self, connector, reason):
-        print("failed to connect", reason.getErrorMessage())
-        print("retrying...")
+        print("Failed to connect to router: {0}".format(reason.getErrorMessage()))
         ReconnectingClientFactory.clientConnectionFailed(self, connector, reason)
 
     def buildProtocol(self, addr):
@@ -110,8 +108,7 @@ class ReconnectingWampWebSocketClientFactory(WampWebSocketClientFactory, Reconne
         # dont reconnect in case of close by us
         if self._proto.closedByMe:
             return
-        print("disconnected", reason.getErrorMessage())
-        print("reconnecting...", self._proto._session)
+        print("Disconnected from router: {0}".format(reason.getErrorMessage()))
         ReconnectingClientFactory.clientConnectionLost(self, connector, reason)
 
 
--- a/autobahn/autobahn/twisted/wamp.py
+++ b/autobahn/autobahn/twisted/wamp.py
@@ -100,7 +100,6 @@ class ReconnectingWampWebSocketClientFactory(WampWebSocketClientFactory, Reconne
     def buildProtocol(self, addr):
         # only way of knowing that connection has been closed by us is to do this cross layer talks :-/
         self._proto = WampWebSocketClientFactory.buildProtocol(self, addr)
-        self._proto.reconnecting = True
         return self._proto
 
     def clientConnectionLost(self, connector, reason):
--- a/autobahn/autobahn/twisted/wamp.py
+++ b/autobahn/autobahn/twisted/wamp.py
@@ -99,12 +99,19 @@ class ReconnectingWampWebSocketClientFactory(WampWebSocketClientFactory, Reconne
         print("retrying...")
         ReconnectingClientFactory.clientConnectionFailed(self, connector, reason)
 
+    def buildProtocol(self, addr):
+        # only way of knowing that connection has been closed by us is to do this cross layer talks :-/
+        self._proto = WampWebSocketClientFactory.buildProtocol(self, addr)
+        self._proto.reconnecting = True
+        return self._proto
+
     def clientConnectionLost(self, connector, reason):
         WampWebSocketClientFactory.clientConnectionLost(self, connector, reason)
-        if reason.check(ConnectionDone):
+        # dont reconnect in case of close by us
+        if self._proto.closedByMe:
             return
         print("disconnected", reason.getErrorMessage())
-        print("reconnecting...")
+        print("reconnecting...", self._proto._session)
         ReconnectingClientFactory.clientConnectionLost(self, connector, reason)
 
 
@@ -496,6 +503,7 @@ class Service(service.MultiService):
         self.debug_wamp = debug_wamp
         self.debug_app = debug_app
         self.make = make
+        self.session = None
         service.MultiService.__init__(self)
         self.setupService()
 
@@ -507,10 +515,11 @@ class Service(service.MultiService):
 
         # factory for use ApplicationSession
         def create():
-            cfg = ComponentConfig(self.realm, self.extra)
-            session = self.make(cfg)
-            session.debug_app = self.debug_app
-            return session
+            if self.session is None:
+                cfg = ComponentConfig(self.realm, self.extra)
+                self.session = self.make(cfg)
+                self.session.debug_app = self.debug_app
+            return self.session
 
         # create a WAMP-over-WebSocket transport client factory
         transport_factory = ReconnectingWampWebSocketClientFactory(
@@ -527,3 +536,6 @@ class Service(service.MultiService):
 
         client = clientClass(host, port, transport_factory)
         client.setServiceParent(self)
+
+    def stopService(self):
+        return self.session.leave()
--- a/autobahn/autobahn/twisted/wamp.py
+++ b/autobahn/autobahn/twisted/wamp.py
@@ -27,6 +27,8 @@ from twisted.internet.defer import Deferred, \
     maybeDeferred, \
     DeferredList, \
     inlineCallbacks
+from twisted.internet.protocol import ReconnectingClientFactory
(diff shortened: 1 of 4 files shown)


oberstet:
Oh, thanks. Did you test it?

In general, I think it's better than nothing for now, though it generates log noise that can't be turned down, the parameters that automatic reconnection cannot be controlled (like jitter etc http://twistedmatrix.com/documents/current/api/twisted.internet.protocol.ReconnectingClientFactory.html).

AutobahnJS has a quite sophisticated logic here: it differentiates between cases like a connection was never successful in the first place vs the connection was lost, and then doing auto-reconnect (https://github.com/tavendo/AutobahnJS/blob/master/package/lib/connection.js#L93). And also allows the app code to quite auto-reconnecting manually etc.

But we can add that later. Main thing: did you test basic functionality?

tardyp:
Unit testing that is quite difficult, so I skipped that part :-}

I did manual tests by starting the router manually after starting the unit tests.

For reconnection after drop, I also did the test via stop + start the router in the middle of the tests.
This did one not work well. I need to look more thoroughly what can be done to delay the calls until reconnection is done. Need to see if the session id is kept after reconnection, and thus whether we need call onLeave/onJoin.

So indeed reconnection needs more work

tardyp:
New version that really works with reconnect upon disconnection

tests:

TEST_DISCONNECTION=1 trial autobahn.wamp.test.test_component.TestRpc.test_case2

then start and stop your router

oberstet:
haha, now that you got me hooked on pep8/pyflakes: https://travis-ci.org/tavendo/AutobahnPython/builds/50539087

trivial, I know, I just couldn't resist;)

I've made Crossbar.io also fully comply: https://github.com/crossbario/crossbar/issues/243

Now back to the topic at hand: will try! need to leave office, but then ...

tardyp:
in buildbot, we have some nice post-commit hook that does the checks automatically on the modified files. So we don't have to wait for travis for such trivial things as flake8.
https://github.com/buildbot/buildbot/blob/master/common/validate.sh

tardyp:
fixed pep8 and py3

oberstet:
When there is no transport, no WAMP action can be performed (as without a transport, there is no session). When a new transport is established (whether reconnecting or initial), this means there is a new WAMP session. E.g. any previous subscriptions or registrations are lost.The app must be able to cope with this. Queuing actions isn't what we want to do - this will seriously mix things up.

Adding auto-reconnect to WAMP-over-WebSocket transport (for Twisted) should not touch `wamp/protocol.py` at all.

I'll try to address the underlying issue (need auto-reconnect) with high prio and come up with something ..

tardyp:
Well event queing is really something that we need in buildbot.
We have lots of people complaining that the protocol is stateful i.e when
there is  a slight loss of network connectivity between the master and the
slave, then the build will just stop.

With this design, the pending transactions are just frozen until the
network come back in a better shape.

onJoin is still called *before* transactions are happening again, which
indeed is necessary to make subscription and registration come back
(provided those are static, which is the case in buildbot protocol)

oberstet:
I don't know enough about Buildbot. I would have expected a build - once started from a client - run independent of the client (even continues if the client dies). If this is about the Buildbot client that actually executes build steps, and that client looses connection, how would it know how to continue anyway if the current step is done? But I guess I don't know enough to even discuss this in a helpful way ..

In any case, event (or even call) queuing inside a WAMP client library is wrong from Autobahn's design perspective. As a starter: it mixes up ordering guarantees, what about acknowledged events, and so on. Note that without acknowledged publications, a router is allowed to silently drop your event anyway (be it because of lack of authorization to publish to the topic, or for other reasons). If you want event/call queuing, this has to be at app level. That could be trivial: instead of publishing, push events to a queue (not tied to a session), and pop and publish from there.

The difference is: the app is then expected to know how and what to do when reconnecting (purge the queue, or publish queued stuff, or whatever). A WAMP client library cannot know the right thing.

And finally: event queuing is orthogonal to auto-reconnect at the transport level. I will address the latter (but not the former). Sorry, but I am quite picky about the exact line of API: what should be done inside a WAMP client library, and what should *not* be done there, but at app level ...

oberstet:
One more note: when a transport is reconnected, a completely new `ApplicationSession` object should be created. A single instance of `ApplicationSession` (on a active transport) can leave a realm, and join another realm, but when the transport goes away, the session object should go away.

This is in accordance with the spec https://github.com/tavendo/WAMP/blob/master/spec/basic.md#transport-and-session-lifetime, and in accordance with how AutobahnJS works.

tardyp:
>One more note: when a transport is reconnected, a completely new ApplicationSession object should be created. A single instance of ApplicationSession (on a active transport) can leave a realm, and join another realm, but when the transport goes away, the session object should go away.

Yes, I made that on purpose, and I realize that this is not the way wamp was originally designed.

I though it was possible while originally looking at the AutobahnPython api.
There is an onConnect(), and and onJoin() on the session class. At that time, it did not make sense if it was not for implementing session persistance over disconnection (a feature that would solve several of issues for me, in the slave protocol, and the web ui protocol).

Then I looked at the actual implementation, and I see this is actually implementation details, and mixed separation of concerns. onConnect is for transport Layer, and onJoin for application Layer.

I am sorry that session persistance over disconnection is not part of advanced feature in wamp. I may not yet have all design details understanding, but I have the feeling that it is simpler to implement in the wamp layer rather than in the application layer.

I think for the moment wamp is already resolving a lot of problems for me, so this is good enough. I will implement recover on disconnection on the app layer, but probably not session persistance, as indeed after more thoughts there are some details that the router need to resolve so that it can really work.

For medium term, I'd like to propose an enhancement to the advanced protocol.
Feature: session persistance over reconnect.
* needs to be supported in the client, and in the broker
* if requested session would persist until GOODBYE is sent by one of the two end.
* broker would have to queue messages until connection is restored (or a timeout)
* client would have to queue message and RPCs until connection is restored
Let me know if this is completely unfeasible. I value a lot your expertise to help us making a good system.

> I don't know enough about Buildbot. I would have expected a build - once started from a client - run independent of the client (even continues if the client dies). If this is about the Buildbot client that actually executes build steps, and that client looses connection, how would it know how to continue anyway if the current step is done? But I guess I don't know enough to even discuss this in a helpful way ..
In buildbot, slaves are pretty dumb, they basically run commands, and forward logs to the master (more or less).
If the connection is dropped, what needs to be done is to queue logs of the current command, to send that to master when connection reappears. If the command is done, just wait until connection is restored, so that the master sends new command.
With current protocol, if connection is dropped, then master just interrupts the current build and restart it, which is potentially very harmful when you are running long lasting tests (some people in my team runs stability (aka MTBF) tests for a week)

oberstet:
> Then I looked at the actual implementation, and I see this is actually implementation details, and mixed separation of concerns. onConnect is for transport Layer, and onJoin for application Layer.

Yes, that is true. It is mixing concerns. Generally a bad thing. This particular design developed out of the motivation to have it all pluggable. There might be a better design.

FWIW, this is how it works today:

There are 3 core internal interfaces to all the WAMP stuff:

* `ITransport` and `ITransportHandler`: https://github.com/tavendo/AutobahnPython/blob/master/autobahn/autobahn/wamp/interfaces.py#L218 and https://github.com/tavendo/AutobahnPython/blob/master/autobahn/autobahn/wamp/interfaces.py#L175
* `ISession` https://github.com/tavendo/AutobahnPython/blob/master/autobahn/autobahn/wamp/interfaces.py#L249

The former 2 abstract away any transport specifics. This works and does the job. However, there needs to be hooking from transport into Sessions. E.g. when a transport is established, a session needs to start the WAMP opening handshake.

If you look at `ApplicationSession`, this class actually implements both `ITransportHandler` _and_ `ISession`.

E.g. https://github.com/tavendo/AutobahnPython/blob/master/autobahn/autobahn/wamp/protocol.py#L327 does the hooking.

In a way, the transport level `onOpen` is mirrored on the session (interface) in `onConnect` (same for `onDisconnect` and `onClose`).

App classes deriving of `ApplicationSession` should not be concerned of the detail that `ApplicationSession` _also_ implements `ITransportHandler`.

I am not completely convinced of this particular design myself. Specially, since AutobahnJS takes a slightly different approach internally (and API wise). I wrote AutobahnJS after AutobahnPython. Then, AutobahnCpp is yet different. However, _after_ a session is there, the APIs of all 3 are as close as language differences permit (I think).

> I'd like to propose an enhancement to the advanced protocol. Feature: session persistance over reconnect.

This sounds like a potentially powerful feature which, depending on situation and use case, could take away a lot of burden from apps. It also sounds complex;) Anyway, yes, please, I'd be interested in your thoughts here! File it on WAMP and we see where we can take it.

"Complex": in general, here is my current thinking. There a messaging queueing systems that maintain a queue per client. WAMP wasn't designed so. Deliberately. Performance! (sidenote: we also made sure that a dispatched message only needs to be serialized once inside a broker). If a client needs to catch up with stuff lost because it wasn't connected for some time, what the WAMP AP has (well, no text, but the heading;) is "event history". So instead of maintaining per-client queues (which come at great cost, the broker may maintain a (limited) history on a topic, and the client can just retrieve that. Cost for client: it needs to remember up to which point it already got all events. It needs to skip any dups if it can't do the former. With brokers that maintain per-client queues, this is done by the broker. However, as said, the price is high - very high. Now, whether it's worth nevertheless to add a "per-client queuing" feature we need to evaluate carefully. Also: what you propose goes even further - since of the client side queueing (essentially, both sides - client and broker - can queue). Further: in the context of WAMP, we not only need to care about "broker", but also "dealer". What does it mean to queue a call? What is the semantics?

Anyway: this is fascinating, so pls file it (though pls don't hold your breath .. it'll be likely too far into the future for Buildbot to depend on). But: I am taking this serious because you actually have a use-case driven itch to scratch! Overall, this has served WAMP design / evolution very well: not having academic, theoretical driven discussions / design, but be driven by user *needs*.

tardyp:
same here, I will remove my hacks and do a simpler propal

tardyp:
After twisting my head on how to make this working in a generic manner, I cannot figure it out.
So for the moment I close this PR, in favor of a new much more simple solution

oberstet:
No worries .. I will adress this (and your other PR) over the weekend.

PR #341, opened 2015-02-12
Handle connection errors

meejah:
ApplicationRunner.run() ignores the Deferred from .connect() which means if there's a connection problem, run() won't exit the event-loop and no exception will be raised to the caller. Minimal test-case:

```python

class Session(ApplicationSession):
    pass

runner = ApplicationRunner('ws://localhost:1', 'realm')
runner.run(Session) 
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
 install:
-  - pip install tox 
+  - pip install tox mock
 
 env:
   - TOX_ENV=flake8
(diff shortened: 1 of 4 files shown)


oberstet:
@meejah nice solution! I have no clue why Travis denies to install mock.

Maybe 2 lines work:

```
install:
  - pip install tox 
  - pip install mock
```

meejah:
I'm going to squash this branch and do another PR.

meejah:
@oberstet see #342

PR #352, opened 2015-03-05
Callback exception handling

meejah:
This adds several unit-tests around subscribe() handling, simplifies the callback-invoking code, handles an error-case with duplicate subscription IDs and suggests a way to handle exceptions coming from "user"/callback code (but only covers the case of such an error in a publish() callback).


Paranaix:
+1

meejah:
I re-squashed the branch and re-did this pull request as #353 after Travis-CI became green.

PR #353, opened 2015-03-05
Callback exception handling

meejah:
This adds several unit-tests around subscribe() handling, simplifies the callback-invoking code, handles an error-case with duplicate subscription IDs and suggests a way to handle exceptions coming from "user"/callback code (but only covers the case of such an error in a publish() callback).

Note this is a re-squashed version of #352 after I got Travis-CI to be happy.

### DIFF
--- a/autobahn/autobahn/twisted/wamp.py
+++ b/autobahn/autobahn/twisted/wamp.py
@@ -86,13 +86,15 @@ class ApplicationSession(FutureMixin, protocol.ApplicationSession):
     WAMP application session for Twisted-based applications.
     """
 
-    def onUserError(self, e):
+    def onUserError(self, e, msg):
         """
         Override of wamp.ApplicationSession
         """
         # see docs; will print currently-active exception to the logs,
         # which is just what we want.
         log.err()
+        # also log the framework-provided error-message
+        log.err(msg)
 
 
 class ApplicationSessionFactory(FutureMixin, protocol.ApplicationSessionFactory):
--- a/autobahn/autobahn/twisted/wamp.py
+++ b/autobahn/autobahn/twisted/wamp.py
@@ -86,6 +86,14 @@ class ApplicationSession(FutureMixin, protocol.ApplicationSession):
     WAMP application session for Twisted-based applications.
     """
 
+    def onUserError(self, e):
+        """
+        Override of wamp.ApplicationSession
+        """
+        # see docs; will print currently-active exception to the logs,
+        # which is just what we want.
+        log.err()
+
 
 class ApplicationSessionFactory(FutureMixin, protocol.ApplicationSessionFactory):
     """
--- a/autobahn/autobahn/wamp/protocol.py
+++ b/autobahn/autobahn/wamp/protocol.py
@@ -70,7 +70,8 @@ class Handler:
     """
     """
 
-    def __init__(self, obj, fn, topic, details_arg=None):
+    def __init__(self, subscription_id, obj, fn, topic, details_arg=None):
+        self.subscription_id = subscription_id
         self.obj = obj
         self.fn = fn
         self.topic = topic
@@ -94,10 +95,11 @@ class Subscription:
     Object representing a subscription.
     This class implements :class:`autobahn.wamp.interfaces.ISubscription`.
     """
-    def __init__(self, session, subscriptionId):
+    def __init__(self, session, subscriptionId, handler):
         self._session = session
         self.active = True
         self.id = subscriptionId
+        self.handler = handler
 
     def unsubscribe(self):
         """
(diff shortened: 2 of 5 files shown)


autobahn/autobahn/wamp/protocol.py
```diff
@@ -481,15 +478,20 @@ def error(err):
...
                 if msg.request in self._subscribe_reqs:
                     d, obj, fn, topic, options = self._subscribe_reqs.pop(msg.request)
-                    if options:
-                        self._subscriptions[msg.subscription] = Handler(obj, fn, topic, options.details_arg)
-                    else:
-                        self._subscriptions[msg.subscription] = Handler(obj, fn, topic)
-                    s = Subscription(self, msg.subscription)
-                    self._resolve_future(d, s)
+                    sub_id = msg.subscription
+                    # we should NEVER have our ID subscribed already
```
oberstet: No. We can receive the same subscription ID - if we are already subscribed. E.g. sending two `SUBSCRIBE` messages with same topic is fine. Router will answer with same subscription ID.

autobahn/autobahn/twisted/wamp.py
```diff
@@ -86,6 +86,14 @@ class ApplicationSession(FutureMixin, protocol.ApplicationSession):
     WAMP application session for Twisted-based applications.
     """
 
+    def onUserError(self, e):
```
oberstet: We need a default implementation for asyncio too.

autobahn/autobahn/wamp/protocol.py
```diff
@@ -432,41 +445,25 @@ def error(err):
...
-                                    handler.fn(*msg.args)
-                                else:
-                                    handler.fn()
+                        handler.fn(*invoke_args, **invoke_kwargs)
 
                     except Exception as e:
+                        self.onUserError(e)
                         if self.debug_app:
-                            print("Failure while firing event handler {0} subscribed under '{1}' ({2}): {3}".format(handler.fn, handler.topic, msg.subscription, e))
+                            traceback.print_exc()
```
oberstet: Not sure, but why isn't this in the default implemention of `onUserError`

autobahn/autobahn/wamp/protocol.py
```diff
@@ -432,41 +445,25 @@ def error(err):
...
-                                    handler.fn(**msg.kwargs)
-                            else:
-                                if msg.args:
-                                    handler.fn(*msg.args)
-                                else:
-                                    handler.fn()
+                        handler.fn(*invoke_args, **invoke_kwargs)
 
                     except Exception as e:
+                        self.onUserError(e)
```
oberstet: What if `onUserError` raises? Probably just eat any exception then without any logging or whatever would be fine ..

autobahn/autobahn/wamp/test/test_protocol.py
```diff
@@ -246,6 +248,208 @@ def on_event(*args, **kwargs):
             self.assertTrue(type(subscription.id) in (int, long))
 
         @inlineCallbacks
+        def test_publish_duplicate_subscription_id(self):
```
oberstet: See above. Crossbar.io would fail this, but this is how it should be.

oberstet:
Cool!! Most of the patch looks good and I'd like to merge ...

autobahn/autobahn/wamp/protocol.py
meejah: Oh, okay. I'll go back to the other code I had then -- which treats self._subscriptions as as list of Handlers. So then in the tests, subscription0.id should == subscription1.id if I used the same topic, right?

autobahn/autobahn/twisted/wamp.py
meejah: Yes, for sure. I wasn't sure what was going on for logging with asyncio.

meejah:
Okay, I'll fix the fact that subscription IDs should be the same for the same topic.

For logging, what if I just put "traceback.print_exc" as the default implementation of onUserError (better name here?) and then only Twisted gets a custom override?

autobahn/autobahn/wamp/protocol.py
oberstet: It should be a list of handlers. It should do the SUBSCRIBE, and when the subscription ID comes back look into the list and append the handler if it's not already in the list. When doing unsubscribe(handler) it should remove the given handler (if it's in the list) and if the list becomes empy, send out an UNSUBSCRIBE. If doing an unqualified unsubscribe() - that is one not providing a specific handler to be removed - it should just send out the UNSUBSCRIBE (and remove the list from the map).

The subscription ID will be the same for the pair (topic, match policy). Two subscribes for the same topic, but different match policies will return different subscription IDs.

meejah:
Not sure why the timestamps are old on that last commit.
Want me to re-squash?

PR #357, opened 2015-03-12
More error handling

meejah:
This covers a whole bunch more cases where "user" code can cause exceptions (mostly in ISession members). It also explicitly tests the cases where those methods return Deferreds instead of acting synchronously. That wasn't documented as allowed, but I don't see why they shouldn't be permitted to behave asynchronously.

Still needs some more test-coverage in protocol.py and should have same/similar tests run in the asyncio version as well, ideally.

This PR also adds coverage commands for at least the Twisted tox environment -- the others should probably get similar commands as well.

If the general idea here is sound, I'll squash and rebase off master.


autobahn/autobahn/wamp/types.py
```diff
@@ -434,7 +434,7 @@ def __init__(self,
         assert(timeout is None or (type(timeout) in list(six.integer_types) + [float] and timeout > 0))
         assert(disclose_me is None or type(disclose_me) == bool)
 
-        self.on_progress = on_progress
+        self.onProgress = on_progress
```
oberstet: should not be renamed

autobahn/tox.ini
```diff
@@ -62,10 +63,14 @@ setenv =
 [testenv:py27twisted]
 commands =
    sh -c "which python"
-   sh -c "which trial"
+   sh -c "which coverage"
```
oberstet: should still have "which trial"

autobahn/autobahn/asyncio/wamp.py
```diff
@@ -98,13 +100,19 @@ def _reject_future(future, value):
         future.set_exception(value)
 
     @staticmethod
-    def _add_future_callbacks(future, callback, errback):
+    def _add_future_callbacks(future, callback, errback,
```
oberstet: I can just use currying `add_future_callbacks(future, lambda x: callback("foo", x))` which is more flexible, since it allows me to chose the order between actual result and additional parameters to callback.

autobahn/autobahn/asyncio/wamp.py
```diff
@@ -98,13 +100,19 @@ def _reject_future(future, value):
...
                 res = f.result()
-                callback(res)
-            except Exception as e:
-                errback(e)
+                if callback:
+                    callback(res, *callbackArgs, **callbackKeywords)
+            except:
+                typ, exc, tb = sys.exc_info()
+                if errback:
+                    args = (typ, exc, tb) + errbackArgs
```
oberstet: No. I don't want to handle those extra args everywhere.

oberstet:
Well, I'm not gonna merge it. It does too many things I am not convinced of ..

autobahn/autobahn/wamp/types.py
meejah: There is other code that uses the camel-case, so wasn't sure what was correct. Also some examples. For example, here: https://github.com/tavendo/AutobahnPython/blob/master/autobahn/wamp/protocol.py#L660

So, everything should use on_progress instead?
I figured maybe the camelCase was the right way as it matched ISession-style API

meejah:
I made the following changes:

- rebased to master
- took out the Twisted-API args/kwargs callback/errback stuff
- ensure onUserError is always wrapped in try/except
- use on_progress intead of onProgress consistently

(Note I force-pushed the rebased branch over top of this one; not sure how nicely or not that plays with github + travis but appears to be working...)

One thing I wasn't sure about: are all the ISession methods (onJoin etc) allowed to be coroutines/return Deferreds? This change does ensure they can, and I clarified that in the documentation...

meejah:
I squashed this, and will do a new pull-request.

PR #358, opened 2015-03-12
More error handling

meejah:
This includes unit-tests and fixes for:

 - consistently handle async returns from onLeave etc methods
 - usable, consistent errback API between asyncio/Twisted
 - test-coverage for user-code raising errors
 - test-coverage for exceptions from invoked methods
 - on_progress used consistently everywhere
 - test-coverage for happy-path + exceptions in on_progress handlers

### DIFF
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -26,6 +26,8 @@
 
 from __future__ import absolute_import
 
+import sys
+
 from autobahn.wamp import protocol
 from autobahn.wamp.types import ComponentConfig
 from autobahn.websocket.protocol import parseWsUrl
@@ -102,9 +104,12 @@ class FutureMixin(object):
         def done(f):
             try:
                 res = f.result()
-                callback(res)
-            except Exception as e:
-                errback(e)
+                if callback:
+                    callback(res)
+            except:
+                typ, exc, tb = sys.exc_info()
+                if errback:
+                    errback(typ, exc, tb)
         return future.add_done_callback(done)
 
     @staticmethod
--- a/autobahn/twisted/wamp.py
+++ b/autobahn/twisted/wamp.py
@@ -30,6 +30,7 @@ import sys
 import inspect
 
 from twisted.python import log
+from twisted.python.failure import Failure
 from twisted.application import service
 from twisted.internet.defer import Deferred, \
     maybeDeferred, \
@@ -84,7 +85,29 @@ class FutureMixin(object):
 
     @staticmethod
     def _add_future_callbacks(future, callback, errback):
-        return future.addCallbacks(callback, errback)
+        # callback and/or errback may be None
+
+        # Note that errback takes 3 args: type, exception, tb our
+        # errbacks can be consistent between Twisted and asyncio.
+
+        # alternatively: what if we made a dummy Failure-like thing
+        # for asyncio so our "standard" errback in protocol.py could
+        # use the Failure API, or at least the bits we need...like
+        # collecting the traceback frames.
+
+        def _errback(fail):
+            # converting to common API between asyncio/Twisted
+            return errback(fail.type, fail.value, fail.tb)
+
+        if callback is None:
+            assert errback is not None
+            future.addErrback(_errback)
+        elif errback is None:
+            future.addCallback(callback)
+        else:
+            future.addCallbacks(callback, _errback)
+
+        return future
 
     @staticmethod
     def _gather_futures(futures, consume_exceptions=True):
@@ -96,15 +119,14 @@ class ApplicationSession(FutureMixin, protocol.ApplicationSession):
     WAMP application session for Twisted-based applications.
     """
 
-    def onUserError(self, e, msg):
+    def onUserError(self, typ, exc, tb, msg=None):
         """
         Override of wamp.ApplicationSession
         """
-        # see docs; will print currently-active exception to the logs,
-        # which is just what we want.
-        log.err()
-        # also log the framework-provided error-message
-        log.err(msg)
+        log.err(Failure(exc, typ, tb))
+        # also log the framework-provided error-message, if any
+        if msg:
+            log.err(msg)
 
 
 class ApplicationSessionFactory(FutureMixin, protocol.ApplicationSessionFactory):
--- a/autobahn/twisted/wamp.py
+++ b/autobahn/twisted/wamp.py
@@ -123,10 +123,10 @@ class ApplicationSession(FutureMixin, protocol.ApplicationSession):
         """
         Override of wamp.ApplicationSession
         """
-        log.err(Failure(exc, typ, tb))
-        # also log the framework-provided error-message, if any
         if msg:
-            log.err(msg)
+            log.err(Failure(exc, typ, tb), msg)
+        else:
+            log.err(Failure(exc, typ, tb))
 
 
 class ApplicationSessionFactory(FutureMixin, protocol.ApplicationSessionFactory):
--- a/autobahn/wamp/interfaces.py
+++ b/autobahn/wamp/interfaces.py
@@ -262,7 +262,10 @@ class ISession(object):
     @abc.abstractmethod
     def onConnect(self):
         """
-        Callback fired when the transport this session will run over has been established.
+        Callback fired when the transport this session will run over has
+        been established.
+
+        May return a Deferred/Future.
         """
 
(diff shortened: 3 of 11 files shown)


autobahn/asyncio/wamp.py
```diff
@@ -102,9 +104,12 @@ def _add_future_callbacks(future, callback, errback):
...
                 res = f.result()
-                callback(res)
-            except Exception as e:
-                errback(e)
+                if callback:
+                    callback(res)
+            except:
+                typ, exc, tb = sys.exc_info()
+                if errback:
+                    errback(typ, exc, tb)
```
oberstet: as said, I won't merge this

autobahn/twisted/wamp.py
```diff
@@ -84,7 +85,29 @@ def _reject_future(future, error):
 
     @staticmethod
     def _add_future_callbacks(future, callback, errback):
-        return future.addCallbacks(callback, errback)
+        # callback and/or errback may be None
+
+        # Note that errback takes 3 args: type, exception, tb our
+        # errbacks can be consistent between Twisted and asyncio.
```
oberstet: no, **both** should use **1** argument: the error that happened. is this not the case currently? I don't get it.

autobahn/wamp/interfaces.py
```diff
@@ -262,7 +262,10 @@ class ISession(object):
     @abc.abstractmethod
     def onConnect(self):
         """
-        Callback fired when the transport this session will run over has been established.
+        Callback fired when the transport this session will run over has
+        been established.
+
```
oberstet: *if* we decide to do that (which we need to discuss first since it is API changing), *then* we should do that in a different commit. In any case, looking at the impacts takes some thinking and time.

autobahn/twisted/wamp.py
meejah: No, currently it's inconsistent: if you're using Twisted, you get a Failure and if you're using asyncio you get just the Exception instance. In this case, you can't print out a traceback (and your error-handler has to do isinstance() or some tricks to figure out if you're in asyncio or Twisted).

So, I switched them to the same args that sys.exc_info() returns so that they're a) consistent, and b) you can actually print out a traceback if you need/want to. We could switch to having a single arg (the exception instance, for example) but that's a lot less flexible, IMO (and would still need a wrapper-errback for Twisted to extract the exception from the Failure).

autobahn/wamp/interfaces.py
meejah: You're talking about the ability for those ISession methods to be async?

There are some instances in Crossbar that decorate at least onJoin with inlineCallbacks -- so the try...catch around self.onJoin would be insufficient to catch exceptions from such methods...

autobahn/twisted/wamp.py
oberstet: Ah, I see. So that's the issue with the current code even: with Twisted, app code gets `Failure` instead of plain exception. It should be the latter for both. And yes, that requires a Twisted wrapper to unfold the plain exception value. Regardng naming: sorry, yes, the naming suggest private, but it's actually intended as incubation to be made public. Because if apps want to write source compatible code (between Twisted and asyncio), they would be pointed to that API - if we release it.

autobahn/wamp/interfaces.py
oberstet: Yes, that ability. The `onJoin` callback (should be named `on_join`, but that's a different story), can fire multiple times without the underyling transport to go away. Would that be ok with having all callback we allowed to return a deferred/future?

meejah: Yes, I believe so.

All I'm really changing in this PR is how the error-cases are handled. Currently, even with the ``try..catch`` and (recent) onUserError handler we still would simply never log an error *if* the handler happened to be async, and didn't have an error until after the first yield.

Now, we will always attach _swallow_error() which will at least log the error (and then ignore it). Of course, this would be a "last-chance" handler -- if user-code put their own error handling inside the onJoin or whatever, they would get called first.

At least, that's the theory ;)

I did add both "returns Deferred" and "simply throws" unit-tests for all the ISession handlers, *and* covered all the code-paths in protocol.py that lead to those handlers being called.

So: I don't expect *any* user-visible changes from this PR, except that errors now get logged.

autobahn/twisted/wamp.py
meejah: Looking at the code in Twisted's Failure, for example, my opinion would be that we should provide at least the 3 args from sys.exc_info() -- otherwise, I don't know of any way that you could print out a stack trace in your error handler.

All you'd be able to do is print out the Exception instance.

As I noted, another alternative would be to provide our own Failure-type object for asyncio (and then in Twisted you could get your Failure and in asyncio you'd get "some object that has many of the same methods"). Not sure if this is a "win" or not though ;)

autobahn/wamp/interfaces.py
meejah: Shall I add a unit-test for "multiple onJoin calls" just to be explict that's supported (and works)?

autobahn/twisted/wamp.py
oberstet: Providing 3 args: doesn't that mean us inventing even a 3rd way of "wrapping" the same information? The other being Twisted failure, and the missing asyncio way.

meejah: All 3 of those bits of info are in the Failure, and we have to provde a Twisted errback-wrapper for either the 3-arg case or the "just the Exception" case anyway.

For asyncio, we immediately have access to those args (we're in "except") and are directly calling the errback() anyway. So, it maps fairly well to how it was already working -- except that you'd get the same thing in the errback no matter if you're in Twisted or asyncio.

The "other" option (our own Failure-type thing for asyncio) means no wrapper-errback for Twisted, but probably more work (although, it *could* just be a simple data-container for now that just has self.type, self.value and self.tb I guess). It would have the advange for Twisted users of being a more-"normal" single-arg errback too.

autobahn/wamp/interfaces.py
oberstet: My main issue with this PR: it's actually 2 unrelated things in one (big) PR: about error handling and async callbacks (plus some smaller things;)

oberstet: Btw: I cherrypicked the Tox changes, and would merge the async callbacks change right away.

autobahn/twisted/wamp.py
meejah: I'd probably lean towards the "Failure-ish thing for asyncio case", but I'm probably really biased since I have more Twisted than asyncio experience ;)

autobahn/wamp/interfaces.py
meejah: Okay, I can split it if you want.
I actually could probably just squash to earlier on my other branch, as I had the error-handling stuff but then realized it wasn't handling async-errors properly either. And *then* I didn't have the right args to do tracebacks in the errbacks.

Anyway. I can split it to do "error handling stuff" (without any of the async-errors?) and then a diferent one with the errback API changes + the handling of async-errors? (Then the second one would depend on the first I guess).

oberstet: Yes, splitting it up would be awesome in this case. The asynch callbacks probably right away - you've convinced me on that one already.

oberstet: Another note: not sure, what do you think? Probably the mixin should be named "loop mixins" and be something to be used in various contexts, not just WAMP transport factories or something.

meejah: Okay, I'll make a new pull-request with just the error-handling things (including async/sync case) but not changing the errback or onUserError API.

Separately, I'll play with the errback API and see if I can make something that follows Twisted style (with the real Failures and no errback wrapper) but that also provides a usable API for asyncio case as well.

oberstet: awesome! sounds great.

meejah: Oh, while I'm thinking of it: do you prefer allowing ``None`` as callback or errback to ``_add_future_callbacks`` or should I mirror Twisted more and add ``_add_future_callback`` and ``_add_future_errback`` to the API as well?

For what it's worth, Twisted's ``addCallbacks`` allows a ``None`` errback but does not allow a ``None`` callback (so there'd be no way to add *just* an errback with such an API -- which is why I let them be ``None`` in the first place).

PR #362, opened 2015-03-16
Sync and async error handling

meejah:
This is a portion of #358 with just the sync and async error-handling and logging added (i.e. it doesn't change the ``onUserError`` or errback APIs).

Since there was a bug in onProgress versus on_progress, this in in here as well; I can take that out as well if you like (it's in a separate commit).

It does make a slight change to ``_add_future_callbacks()`` API so that you can pass ``None`` as a callback in order to provde a way to add *just* an errback.

Note that I re-based the unit-tests to (all) be in a separate commit from the code-changes. Is it better to have the unit-tests in the same commit as the code-changes?


oberstet:
awesome! thanks!

PR #368, opened 2015-03-21
Respect multiple decoration with @subscribe and @register. fixes #335

### LINKED ISSUE #335: Only first @subscribe decorator is acted upon
The `@subscribe` decorator accumulates subscriptions in an array, yet only the first entry in this array is actually being acted upon:

```python
# autobahn/wamp/protocol.py
  ..
  def subscribe(self, handler, topic = None, options = None):
    ..
    for k in inspect.getmembers(handler.__class__, test):
    proc = k[1]
    if "_wampuris" in proc.__dict__:
      pat = proc.__dict__["_wampuris"][0]
      if pat.is_handler():
        uri = pat.uri()
        dl.append(_subscribe(handler, proc, uri, options))
```

Seems to me this should be a `for pat in proc.__dict__["_wampuris"]` instead of just `proc.__dict__["_wampuris"][0]`.

rafzi:
The motivation for subscribe is pretty simple. There are cases where published topics are handled by the same function.

For register it's a bit harder to find a use-case, but maybe this could be a good solution for some authorization problems. Or maybe services of multiple tiers with a unified client, where the simplest service just registers the same function for many URIs.

### DIFF
--- a/autobahn/wamp/protocol.py
+++ b/autobahn/wamp/protocol.py
@@ -1062,11 +1062,11 @@ class ApplicationSession(BaseSession):
             for k in inspect.getmembers(handler.__class__, is_method_or_function):
                 proc = k[1]
                 if "_wampuris" in proc.__dict__:
-                    pat = proc.__dict__["_wampuris"][0]
-                    if pat.is_handler():
-                        uri = pat.uri()
-                        subopts = options or pat.subscribe_options()
-                        on_replies.append(_subscribe(handler, proc, uri, subopts))
+                    for pat in proc.__dict__["_wampuris"]:
+                        if pat.is_handler():
+                            uri = pat.uri()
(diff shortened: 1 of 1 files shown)


meejah:
This PR makes sense for the `subscribe`s for sure as-is, but it's currently an error if you `register()` two procedures on the same endpoint in the same session. I think (seperately) we should improve that behavior (I just filed #400 for these improvements).

That said, you *could* of course choose to register the same procedure at multiple (different) endpoints, so the code in this PR is correct. It might be ideal if it checked to ensure that there's not duplicate endpoints, but overall I think it's better to simply call the "real" `register()` as this code does and fix the bahavior of the underlying `register()` calls as per #400.

Sorry it took us so long to review this, and thanks for the patch!

meejah:
(@oberstet this can be merged)

PR #369, opened 2015-03-21
Added the ability to customize msgpack packing/unpacking arguments

agronholm:
This in particular lets the developer add hooks for (un)serializing types not natively supported by msgpack. The same thing could be done with JSON, except that ujson does not seem to support such hooks, while the standard library json module does.

### DIFF
--- a/autobahn/wamp/serializer.py
+++ b/autobahn/wamp/serializer.py
@@ -253,20 +253,28 @@ else:
       between strings and binary).
       """
 
-        def __init__(self, batched=False):
+        def __init__(self, batched=False, pack_kwargs=None, unpack_kwargs=None):
             """
             Ctor.
 
             :param batched: Flag that controls whether serializer operates in batched mode.
             :type batched: bool
+            :param pack_kwargs = Keyword arguments passed to msgpack.packb().
+            :type pack_kwargs dict
+            :param unpack_kwargs = Keyword arguments passed to msgpack.unpackb().
+            :type unpack_kwargs dict
             """
             self._batched = batched
+            self._pack_kwargs = pack_kwargs or {}
+            self._pack_kwargs.setdefault('use_bin_type', self.ENABLE_V5)
+            self._unpack_kwargs = unpack_kwargs or {}
+            self._unpack_kwargs.setdefault('encoding', 'utf-8')
 
         def serialize(self, obj):
             """
             Implements :func:`autobahn.wamp.interfaces.IObjectSerializer.serialize`
             """
-            data = msgpack.packb(obj, use_bin_type=self.ENABLE_V5)
+            data = msgpack.packb(obj, **self._pack_kwargs)
             if self._batched:
                 return struct.pack("!L", len(data)) + data
             else:
@@ -292,7 +300,7 @@ else:
                     data = payload[i + 4:i + 4 + l]
 
                     # append parsed raw message
-                    msgs.append(msgpack.unpackb(data, encoding='utf-8'))
+                    msgs.append(msgpack.unpackb(data, **self._unpack_kwargs))
 
                     # advance until everything consumed
                     i = i + 4 + l
@@ -302,7 +310,7 @@ else:
                 return msgs
 
             else:
-                return [msgpack.unpackb(payload, encoding='utf-8')]
+                return [msgpack.unpackb(payload, **self._unpack_kwargs)]
 
     IObjectSerializer.register(MsgPackObjectSerializer)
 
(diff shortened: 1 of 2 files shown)


oberstet:
This won't work since we need to be able to transparently translate between different serializations in router code. And a client speaking multiple serialization also doesn't know which one gets selected. Finally, app code should break when serialization format changes ...

agronholm:
Isn't there a configuration option in crossbar that makes it use msgpack?

By limiting the serializers only to msgpack, the application can make sure it gets selected.

I don't understand what you mean by the app code having to break when the serialization format changes.

oberstet:
When the custom msgpack serializer is in use, it might unserialize a datetime e.g.. For clients connected via JSON, Crossbar.io will then try to serialize the datetime into JSON, and that will fail. Crossbar.io has to be able to transparently translate between different serialization formats.

When an app assumes it can use a datetime in a call argument eg (because it relies on custom msgpack serializer), but then the client is forced to speak JSON, this will also fail.

agronholm:
This is a deployment/configuration problem. You wouldn't try to use custom types with clients that can't handle those in the first place. Also, crossbar.io would never unserialize anything to a datetime, because the custom serializers would be in the process containing the autobahn application, right?

agronholm:
I think this deserves a better explanation. Let's say I have an autobahn app that sends a message serialized with msgpack, containing a datetime object. Crossbar.io will unserialize the message and the datetime object is serialized into ExtType(n) where n is the identifier of the custom type assigned in the autobahn app. When crossbar passes the message along, it will use msgpack to reserialize the message (if I understood this right) and msgpack will happily serialize that ExtType instance. The receiving application then unserializes the message again, using its customized msgpack serializer, and properly receives the datetime object.

Paranaix:
Can't we solve on application level (opposed to transport) if Autobahn had a mechanism to register custom serializers and deserializers which convert a custom object into something JSON/msgpack serializable?

E.g:

``` python

def SerializeDate(obj):
    return str(obj)

def DeserializeDate(raw):
   return datetime.strptime(raw)

@register(u"com.example")
def ReturnDate():
    return datetime.now()
```

For returning values this is trivial; for input parameters and return values of RPC calls we would need a interface to map deserializers to parameters/return values.

agronholm:
> For returning values this is trivial; for input parameters and return values of RPC calls we would need a interface to map deserializers to parameters/return values.

Yeah well that's the hard part, isn't it? How do you propose we do that in a universally compatible fashion? I've done that with JSON by assigning values to magic keys like `__type__` but that won't work with arbitrary data since there could be any number of conflicts. For this very reason, version 2 of the msgpack specification added a special type for encapsulating arbitrary custom types.

Paranaix:
`in a universally compatible fashion` I believe this is your main logical error. We don't need a solution which is universally compatible. Such a feature is merely a candy served to you by a client library. WAMP only knows JSON and msgpack for now, how that data is processed is completely up to the client library. So the real question is, how can this be nicely implemented in AutobahnPython.

I think the easiest solution to register serializers for arbitrary types (not everything is a custom model written by the user, e.g `datetime`) would be something like this: `session.RegisterSerializer(datetime, MyDatetimeSerializer`. This is sufficient for return values and parameters in calls.

To automatically convert call parameters as callee something like this might work:
``` python
@register(u"com.example")
@param("dtObj", datetime)
@param(1, MyCustomType)
def SomeWampProcedure(regularParam, customType, dtObj = null):
    pass
```

That decorator could either return a closure which invokes the serialization framework and converts the params or it could similiar to `@register` just annotate the function and `ApplicationSession` will do the conversion upon calls.

Additionally there should be a low level access to the framework to allow code like this `converted = wamp.deserialize(jsonOrMsgpackSerializeable)` for those who don't want to use these annotations and want to have direct control (or who consider this semantically more correct).

Generally speaking: Python is perfectly suited for performing magic tricks. And while it might be need some effort until a nice and clean interface is found, its for sure not hard. Keep in mind that above snippets are only some crude examples I made up in minutes.

agronholm:
And how would you handle custom types appearing in complex data structures, like lists and dictionaries?

Paranaix:
By using the low-level serialization framework. E.g to deserialize a list:

```python
@register(...)
def SomeHandler(someList):
    deserialized = [session.Deserialize(obj, User) for obj in someList]
```

Still not perfect but IMO completely sufficient. Something more convenient (if you really need this...) could be implemented like this:

```python
class UserList:
    # marker class
    pass

def DeseriaizeUserList(someList):
    return [session.Deserialize(obj, User) for obj in someList]

session.RegisterDeserializer(DeserializeUserList, UserList)

@register(...)
@param(0, UserList)
def SomeHandler(userList):
    pass
```

agronholm:
Can you at least explain why you rejected this?

agronholm:
Paranaix, how would register() work? How would the framework know which deserializer to call?

Paranaix:
Sorry for the late reply, had plenty of work to do.

Well the most basic implementation of which I can think of would simply store the deserializer in a map with type -> deserializer, e.g `deserializers[UserList] = DeserializeUserList` (with substituted parameters of course). The low-level `Deserialize` method would then simply lookup the appropriate deserializer for the passed type and call it with the given object.

But please notice: All the code I posted was only meant as a simpe proof of concenpt, to demonstrate a way how it COULD be done.

agronholm:
The passed type? Given data of arbitrary structure, where would the type come from?

Paranaix:
By specifying them in the annotations/decorators as shown above...

agronholm:
How exactly would the decorators work?? Do they take dicts as input and somehow figure out if a dict should be converted to another type?

Do you use IRC? This conversation should've taken an hour MAX, but over github it's taken WEEKS.

Paranaix:
Hey, BB is really lacking the ability to send private messages. I got Skype and ofc email, but I rather don't want to post my email publicly. You can add me in Skype by the same name I'm using on BB.

PR #371, opened 2015-03-25
Ignore any unknown attributes in the RoleXFeatures constructors.

bb4242:
This prevents the client from crashing if the server returns features that the
client doesn't recognize.

### DIFF
--- a/autobahn/wamp/message.py
+++ b/autobahn/wamp/message.py
@@ -278,7 +278,6 @@ class Hello(Message):
             if u'features' in details_role:
                 check_or_raise_extra(details_role[u'features'], "'features' in role '{0}' in 'roles' in 'details' in HELLO".format(role))
 
-                # FIXME: skip unknown attributes
                 role_features = role_cls(**details_role[u'features'])
 
             else:
@@ -431,7 +430,6 @@ class Welcome(Message):
             if u'features' in details_role:
                 check_or_raise_extra(details_role[u'features'], "'features' in role '{0}' in 'roles' in 'details' in WELCOME".format(role))
 
-                # FIXME: skip unknown attributes
                 role_features = role_cls(**details_roles[role][u'features'])
 
             else:
--- a/autobahn/wamp/role.py
+++ b/autobahn/wamp/role.py
@@ -85,7 +85,8 @@ class RoleBrokerFeatures(RoleFeatures):
                  subscriber_blackwhite_listing=None,
                  publisher_exclusion=None,
                  subscription_revocation=None,
-                 event_history=None):
+                 event_history=None,
+                 **kwargs):
         self.publisher_identification = publisher_identification
         self.publication_trustlevels = publication_trustlevels
         self.pattern_based_subscription = pattern_based_subscription
@@ -110,7 +111,8 @@ class RoleSubscriberFeatures(RoleFeatures):
                  publication_trustlevels=None,
                  pattern_based_subscription=None,
                  subscription_revocation=None,
-                 event_history=None):
+                 event_history=None,
+                 **kwargs):
         self.publisher_identification = publisher_identification
         self.publication_trustlevels = publication_trustlevels
         self.pattern_based_subscription = pattern_based_subscription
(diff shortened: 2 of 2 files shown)


oberstet:
A server should not return attributes that aren't in the spec. So we do want to bail out on those ..

bb4242:
Hmm, I thought the presence of the fixme comments (`# FIXME: skip unknown attributes`) in `message.py` implied that you were planning on letting the client gracefully handle receiving unknown attributes.  I'm OK with your decision if that's not the case anymore, but can you explain those comments so I can better understand your intent?  Thanks! :)

Note that erwa currently returns a few attributes that cause the python client to crash.  For example, erwa returns `callee_blackwhite_listing=false`.  This attribute actually *is* part of the [spec](https://github.com/tavendo/WAMP/blob/385f45c4d09c2a559993376fdcc70c5bc0db367b/spec/advanced.md#remote-procedure-calls), but the python client has not been updated to include it yet.  So, in this case, it's actually the python client that's behind the spec, and the current implementation of the client prevents it from connecting to a server that's actually *more* compliant with the spec.

If our approach is going to be that we require all WAMP clients and servers to be on precisely the same version of the spec at all times, that seems like a big problem, in my opinion.  It'll require very tight control over every client and server in the ecosystem, and I think that's going to cause a lot of interoperability problems, especially given that the WAMP spec itself is still evolving.  I personally think it's going to be necessary to allow clients and servers that are written to slightly different versions of the spec to interoperate.

oberstet:
callee black-/white listing: ah yes, that is definitely a bug in Autobahn. Practically however, callee black-/whitelisting only makes sense if the router is able to forward a call to multiple callees. I don't think Erwa has that. But anyway, it's a bug in Autobahn, since "false" should be acceptable.

No doubt, the comment rgd. attributes is misleading;) Actually "skip unknown attributes" should read "skip custom attributes". Can't seem to find the spec text right now, but we want to allow implementation specific attributes in certain places (like options/details in messages .. and probably also in WELCOME). Those attributes should be distinguished from the "standard" ones by having a prefix like `x_` similar what HTTP headers `X-` do. So these should be ignored, while non-prefixed, unknown attributes should be treated more strictly. 

The last paragraph: I agree. Interoperability is maybe not only about being conservative/strict, but forgiving too. On the other hand, letting things run out of control is also a problem.

Thinking about it, there is interoperability between different implementations, but also different versions. And we want to allow progression on the spec, and I guess you are right, that means implementations will be at different versions.

How about this spec text: "Implementations SHOULD ignore (any) attributes they don't understand, MUST only use non-prefixed attributes which are described in the spec, and MAY use prefixed attributes for custom stuff"?

bb4242:
I like the idea prefix custom attributes with something like `x_`.

The spec text you proposed ("Implementations SHOULD ignore (any) attributes they don't understand, MUST only use non-prefixed attributes which are described in the spec, and MAY use prefixed attributes for custom stuff") sounds pretty good to me.  Although, I would argue that "Implementations MUST ignore (any) attributes they don't understand..." is a better alternative that helps ensure interoperability between servers and clients written to slightly different versions of the spec (which, as mentioned, is bound to happen given that the spec itself is still evolving).

Either way, the purpose of this pull request is to make AutobahnPython compliant with that proposed spec text, because right now, it crashes if it receives any attribute (custom or standard) it doesn't understand.  As mentioned, this is preventing interoperability with erwa (and possibly other WAMP servers as well), even though the client and erwa are perfectly capable of talking to each other (note that I have tested that my modified client can connect to erwa successfully).  As a user of the AutobahnPython client, this is a big issue, because it completely prevents me from being able to connect to the server or use the client for anything at all, and I have to turn on debugging and dig into the source code to understand why.  In my opinion, it's a much better user experience to have the client continue to work if it can (and as mentioned, it certainly can in this case).

oberstet:
@bb4242 see also: https://github.com/tavendo/WAMP/issues/150

bb4242:
@oberstet Thanks! :)

Issue #287, opened 2014-10-29
Problem with autobahn >= 0.9.0


balek:
After upgrade crossbar has very strange problems. Some of client requests (`call`s and `subscribe`s) simply not working, if they are goes very quickly one by one. I can't descibe problem more clearly, but I know, that problem is in Autobahn and appeared after commit 3ce3be20a519c173ea80b186ee94572574cac7b7 .

PR #375, opened 2015-04-02
Check for transport in onLeave

meejah:
If our connection was lost we call onClose, which always calls
onLeave, which always does disconnect() -- however if we've
already lost our connection, we don't need to disconnect.

This results in spurious errors during SIGINT shutdown; see crossbar/#278

Two user-error handling test-cases were already "expecting" this
error, so those are changed too...

### DIFF
--- a/autobahn/wamp/protocol.py
+++ b/autobahn/wamp/protocol.py
@@ -961,7 +961,9 @@ class ApplicationSession(BaseSession):
         """
         Implements :func:`autobahn.wamp.interfaces.ISession.onLeave`
         """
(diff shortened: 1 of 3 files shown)


meejah:
build is failing due to a flake8-thing that's on master

estan:
Can confirm that this fixes my problems in Crossbar issue [#278](https://github.com/crossbario/crossbar/issues/278). See my comments/results on Crossbar pull request [#287](https://github.com/crossbario/crossbar/pull/287).

PR #377, opened 2015-04-03
code cleaning

SergeVL:
I am just replace code like
if key in d:
      ...
     del d[key]

to better:
d.pop(key, None)

And I did not touch code in  asyncio/websocket.py which could be done too.


SergeVL:
Any comments?
Failure is not by my code.

oberstet:
I think the code as it is is just fine

SergeVL:
-        if 'serializers' in kwargs:
-            serializers = kwargs['serializers']
-            del kwargs['serializers']
-        else:
-            serializers = None
+        serializers = kwargs.pop('serializers', None)
Sure it's not error, but so many words when there is pop

PR #381, opened 2015-04-15
Fix a typo in websocket/protocol.py (s/taxio/txaio)

cfillion:
Most test cases were raising the following exception when running `wstest -m fuzzingclient` since 7c5c882f62408ee7e7367d196c1c0ccc8ceab582:

```
An exception was raised from application code while processing a reactor selectable
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/twisted/python/log.py", line 88, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/usr/local/lib/python2.7/site-packages/twisted/python/log.py", line 73, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/usr/local/lib/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/usr/local/lib/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/usr/local/lib/python2.7/site-packages/twisted/internet/kqreactor.py", line 277, in _doWriteOrRead
    why = selectable.doRead()
  File "/usr/local/lib/python2.7/site-packages/twisted/internet/tcp.py", line 209, in doRead
    return self._dataReceived(data)
  File "/usr/local/lib/python2.7/site-packages/twisted/internet/tcp.py", line 215, in _dataReceived
    rval = self.protocol.dataReceived(data)
  File "/usr/local/lib/python2.7/site-packages/autobahn/twisted/websocket.py", line 101, in dataReceived
    self._dataReceived(data)
  File "/usr/local/lib/python2.7/site-packages/autobahn/websocket/protocol.py", line 1299, in _dataReceived
    self.consumeData()
  File "/usr/local/lib/python2.7/site-packages/autobahn/websocket/protocol.py", line 1314, in consumeData
    while self.processData() and self.state != WebSocketProtocol.STATE_CLOSED:
  File "/usr/local/lib/python2.7/site-packages/autobahn/websocket/protocol.py", line 1466, in processData
    return self.processDataHybi()
  File "/usr/local/lib/python2.7/site-packages/autobahn/websocket/protocol.py", line 1777, in processDataHybi
    fr = self.onFrameEnd()
  File "/usr/local/lib/python2.7/site-packages/autobahn/websocket/protocol.py", line 1875, in onFrameEnd
[...]

PR #382, opened 2015-04-15
Fix typo (taxio -> txaio).

estan:


estan:
Oh, nevermind. Didn't see #381 :)

PR #394, opened 2015-05-10
Added ssl argument to asyncio.wamp.ApplicationRunner to allow passing an SSL context for client connections

terrycojones:
Added tests for this.

Also (sorry...) fixed an unrelated trivial error (in `autobahn/twisted/wamp.py`) in which `pop` was being called on a `tuple`. This was causing the following failure with `make test_asyncio`:

```
autobahn/wamp/test/test_component.py:37: in <module>
    from autobahn.twisted import wamp
autobahn/twisted/wamp.py:57: in <module>
    __all__.pop(__all__.index("Service"))
E   AttributeError: 'tuple' object has no attribute 'pop'
```

### DIFF
--- a/autobahn/twisted/wamp.py
+++ b/autobahn/twisted/wamp.py
@@ -54,7 +54,7 @@ try:
 except ImportError:
     # Not on PY3 yet
     service = None
-    __all__.pop(__all__.index("Service"))
+    __all__ = tuple(list(__all__).pop(__all__.index('Service')))
 
 
 class ApplicationSession(protocol.ApplicationSession):
--- a/autobahn/twisted/wamp.py
+++ b/autobahn/twisted/wamp.py
@@ -41,20 +41,20 @@ import txaio
 txaio.use_twisted()
 
 
-__all__ = (
+__all__ = [
     'ApplicationSession',
     'ApplicationSessionFactory',
     'ApplicationRunner',
     'Application',
     'Service'
-)
+]
 
 try:
     from twisted.application import service
 except ImportError:
     # Not on PY3 yet
     service = None
-    __all__ = tuple(list(__all__).pop(__all__.index('Service')))
+    __all__.pop(__all__.index('Service'))
 
 
 class ApplicationSession(protocol.ApplicationSession):
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -32,11 +32,11 @@ from autobahn.websocket.protocol import parseWsUrl
 from autobahn.asyncio.websocket import WampWebSocketClientFactory
 
 try:
-    from asyncio import get_event_loop
+    import asyncio
 except ImportError:
-    # Trollius >= 0.3 was renamed
+    # Trollius >= 0.3 was renamed to asyncio
     # noinspection PyUnresolvedReferences
-    from trollius import get_event_loop
+    import trollius as asyncio
 
 import txaio
 txaio.use_asyncio()
@@ -75,7 +75,8 @@ class ApplicationRunner(object):
     """
 
     def __init__(self, url, realm, extra=None, serializers=None,
-                 debug=False, debug_wamp=False, debug_app=False):
+                 debug=False, debug_wamp=False, debug_app=False,
+                 ssl=None):
         """
 
         :param url: The WebSocket URL of the WAMP router to connect to (e.g. `ws://somehost.com:8090/somepath`)
@@ -93,6 +94,9 @@ class ApplicationRunner(object):
         :type debug_wamp: bool
         :param debug_app: Turn on app-level debugging.
         :type debug_app: bool
+        :param ssl: An (optional) SSL context instance or a bool. See the documentation for the
+           `loop.create_connection` asyncio method, to which this value is passed.
+        :type ssl: :class`ssl.SSLContext`
         """
         self.url = url
         self.realm = realm
@@ -102,6 +106,7 @@ class ApplicationRunner(object):
         self.debug_app = debug_app
         self.make = None
         self.serializers = serializers
+        self.ssl = ssl
 
     def run(self, make):
         """
@@ -119,22 +124,32 @@ class ApplicationRunner(object):
             except Exception as e:
                 # the app component could not be created .. fatal
                 print(e)
-                get_event_loop().stop()
+                asyncio.get_event_loop().stop()
             else:
                 session.debug_app = self.debug_app
                 return session
 
         isSecure, host, port, resource, path, params = parseWsUrl(self.url)
 
+        if self.ssl is None:
+            ssl = isSecure
+        else:
+            if self.ssl and not isSecure:
+                raise Exception(
+                    'ssl argument value passed to %s conflicts with the "ws:" '
+                    'prefix of the url argument. Did you mean to use "wss:"?' %
+                    self.__class__.__name__)
+            ssl = self.ssl
+
         # 2) create a WAMP-over-WebSocket transport client factory
         transport_factory = WampWebSocketClientFactory(create, url=self.url, serializers=self.serializers,
                                                        debug=self.debug, debug_wamp=self.debug_wamp)
 
         # 3) start the client
-        loop = get_event_loop()
+        loop = asyncio.get_event_loop()
         txaio.use_asyncio()
         txaio.config.loop = loop
-        coro = loop.create_connection(transport_factory, host, port, ssl=isSecure)
+        coro = loop.create_connection(transport_factory, host, port, ssl=ssl)
         loop.run_until_complete(coro)
 
         # 4) now enter the asyncio event loop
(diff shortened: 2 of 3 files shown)


autobahn/twisted/wamp.py
```diff
@@ -54,7 +54,7 @@
 except ImportError:
     # Not on PY3 yet
     service = None
-    __all__.pop(__all__.index("Service"))
+    __all__ = tuple(list(__all__).pop(__all__.index('Service')))
```
meejah: A perhaps simpler option here is to just change `__all__` to be a list instead of a tuple; it can be either.

terrycojones: OK, changed to `list`.

oberstet:
Yes, being able to pass in an SSL context is useful. As the API between the Twisted and asyncio variant of `ApplicationSession` should be the same, we would at least include a mock implementation.

meejah:
I'm taking a look at the Twisted side. Twisted does provide a nice and much higher-level API for these things, but it also looks like its possible to wedge in a straight SSL context if you really need to -- so I *think* what will work is accepting ``ssl=obj`` where ``obj`` can be an SSL context (for either asyncio or Twisted), *or* one of the appropriate Twisted instances (``twisted.internet.ssl.CertificateOptions``, only for Twisted obviously).

I'll make a proof-of-concept of this to confirm it all works.

I believe @terrycojones actual use-case is being able to use a self-signed certificate successfuly (and at least one other person in ``#autobahn`` has asked how to do this).

terrycojones:
Yes, I just wanted to pass in an SSLContext for the reason @meejah says. I tried to do it in a way that would be as generally useful as possible. I took a quick look at the Twisted code and it wasn't clear how I could go about passing an SSLContext - if at all. See the final part of `_parseClientSSL` on line 1590 of `twisted/internet/endpoints.py`.

meejah:
Hmm. Maybe we *do* want to provide some kind of wrapper in an attempt at sanity. Although "import ssl" uses openssl under the hood, this is a different wrapper from the pyOpenSSL that Twisted uses (so, e.g., even though you can get a Context object passed through to a client endpoint, it *can't* be the one returned from ``ssl.create_default_context()`` but instead must be a ``OpenSSL.SSL.Context`` instance). Similarly, ``asyncio`` wants an instance from ``ssl``...

As @terrycojones points out there's no way to pass raw Context objects (or anything like that) via Twisted's string-parsing endpoint API, but you can get it to work by wrapping the Context instance in a class that provides ``getContext`` and passing that as ``trustRoot=`` to the ``SSL4ClientEndpoint()`` directly.

So, I guess we'd need an API for this wrapper and some use cases to cover?

The main use-case has to be "I want to use a self-signed certificate but still connect somehow" as that's what this PR is about. This can be accomplished a number of ways, the simplest being to turn off all verification (that's I believe what @terrycojones was doing?).

Perhaps we should just add a ``ssl_verify=True`` option for now, and do "whatever it takes" under the hood? Another option might be to pass a list of certificates to trust. Either of these should cover the "self-signed cert" or "development; don't care about certs" types of use-case.

It seems like a bad idea to expose the full horror of the OpenSSL/Context API to users of ``ApplicationRunner``...

terrycojones:
Hey @meejah.

One reaction is to say that if Twisted wants an `OpenSSL.SSL.Context` instance, and asyncio wants an `SSLContext`, then maybe it would work to have `ApplicationRunner` accept an `ssl` argument whose type differs depending on whether you're writing for Twisted or asyncio. Twisted and asyncio are different beasts, and they always will be, so I don't think it's a bad thing to have the same API for `ApplicationRunner` for both of them but a different type for an argument.

I'd caution against adding an `ssl_verify` argument, even though I suspect it would solve my use case (I can test it if you like). The reason is that you'd be trying to dumb down a complex issue and you'd be stuck supporting that dumbing down. If you one day want to support passing an `SSLContext` instance, or some other modifier on SSL behavior, what would you do, add another `ssl*` argument? You can put things into APIs, but you can't take them out without breaking things and pissing people off. Besides which, the `SSLContext` class and the simplified `create_default_context` method have been, I suspect, thought about and designed by people who have a lot of experience in trying to provide for programmers to flexibly and conveniently specify desired SSL behavior. Adding a quick-and-dirty `ssl_verify` argument and calling their API a "horror" :-) is a bit like saying "The people who built that API, they really didn't know what they were doing, that SSLContext that they exposed to regular programmers shouldn't be exposed, I know better, etc."  I know you're *not* saying that, I'm just suggesting that the people who designed `SSLContext` and `create_default_context` probably knew what they were doing, and that I think it would be better to trust them and allow folks to pass such a thing directly through to the API (`create_connection`) that expects it and can deal with it. If we can!

Sorry for such a long reply, and thanks again. I'm happy to keep helping and to modify this pr if wanted. I wont complain if you go with an `ssl_verify` argument, the above are just some thoughts.

meejah:
I do hear you, but even the Python "ssl" module docs basically say "don't make your own SSLContext". Ditto with the Twisted docs, essentially. The "highly recommended" APIs are pretty high-level in both cases.

I also agree its okay (although, IMO, perhaps not ideal) to have an arg that takes different types for asyncio/Twisted -- after all, you're importing a different ``ApplicationRunner`` class as it is...

Another API I tried, which works actually not too bad, is to add a `ca_certificates=` kwarg, which if present is a list of unicode blobs of PEM-encoded certificates to use as CA certificates. This translates into a call to ``ssl.create_default_context(cadata=''.join(ca_certificates))`` (basically) on asyncio or creation of an `OpenSSLCertificateOptions` instance with an appropriate `trustRoot=` arg under Twisted.

One "pro" here is that you don't have to import anything special for either platform. Of course, the main "con" is that it's less flexible of course.

That said, though, if we go with a super-flexible API and let users pass in an SSLContext/ssl.Context right now, that's a **ton** of fleibility to keep supporting...and to me it seems the trend is *away* from exposing the "low-level" OpenSSL API directly (but that could easily be a mistaken impression).

In any case, with a `ca_certificates=[...]` API you'd just pass your self-signed cert(s) (or your own CA cert if you made your own CA and then signed a bunch of certs with it for example, or if your company has a CA cert they use locally, etc). In some ways, this is sort-of a lowest-common-denominator between the public APIs that ``ssl.create_default_context(...)`` and ``twisted.internet.ssl`` expose. It seems powerful enough to cover a few use-cases while not being "too flexible/powerful" for future compatibility...

If a lower-level API is desired, I think the right thing to do would be to accept ``ssl=`` as per @terrycojones patch for asyncio (providing an ssl.Context) and on the Twisted side this arg would be an instance suitable to pass to SSL4ClientEndpoint as ``sslContextFactory=`` (which means ideally an ``OpenSSLCertificationOptions`` instance, None, or a ``ClientContextFactory``). Then, of course, client code *could* achieve the ``ca_certificates=`` functionality themselves (i.e. use OpenSSL to load the certs, and construct the right instances) or do whatever else they felt like (e.g. turn off verification entirely, etc).

/cc @oberstet

terrycojones:
From IRC (#autobahn), for the record:

```
terrycojones: meejah: just read your latest comment. i don’t really know enough about SSL to have an opinion that i think should be weighted much. a good person to ask would be JP (exarkun) of twisted fame, and who maintained pyOpenSSL (I think) for ages. he might have a suggestion about a good approach. he’s a good guy & super knowledgeable & helpful.

meejah: terrycojones: lets chat tomorrow. I'll ask #twisted's opinion too, which I suspect will lean towards "please use the APIs we just introduced in 14.0.0 or so"  but good idea
```

@meejah, I woke up this morning with what I think could be a better approach, which is to change the `ApplicationRunner` class so that it has a `create_connection` method. Then `run` would call `self.create_connection` and anyone who wanted to do something more elaborate could just subclass `ApplicationRunner` and implement their own `create_connection`.

That could be done with the current API, but if that's open to change then I think it could be done quite cleanly. E.g., I'd have the `make` argument be an arg to the class itself, not to `run`.

If you think this approach is worth considering, I'll send a pr (no tests, for now) showing what it might look like. I'd need to know whether the current API needs to be maintained, though.

meejah:
That might be pretty good as well. My inclination in this case would be to pass in a callable, personally, but I guess you might want access to some of the instance-vars.

However, I can see the wisdom of essentially just re-exporting the underlying APIs; if asyncio or Twisted change their APIs, then crossbar users also have to make changes -- but they would anyway, if using those APIs. So I think the ``ssl=`` way is easier in the end? Subclassing ``ApplicationRunner`` just to set an SSL option seems like overkill.

meejah:
@terrycojones I'm not done with this, but here's how a corresponding ``ssl=`` argument to the Twisted side might look: https://github.com/tavendo/AutobahnPython/compare/master...meejah:pass-ssl-context-393 (that re-bases your branch on master, and I fixed up the comment a little).

terrycojones:
This work is carried forward in #398 by @meejah.

PR #396, opened 2015-05-11
Port parts of the Twisted sections to Python 3

hawkowl:
This is enough to get Crossbar working, so far, plus a fair amount of tests passing that don't depend on `twisted.application`.

### DIFF
--- a/autobahn/twisted/resource.py
+++ b/autobahn/twisted/resource.py
@@ -34,6 +34,7 @@ except ImportError:
     # starting from Twisted 12.2, NoResource has moved
     from twisted.web.resource import NoResource
 from twisted.web.resource import IResource, Resource
+from twisted.python.compat import _PY3
 
 # The following imports reactor at module level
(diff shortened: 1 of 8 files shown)


hawkowl:
cc @oberstet

autobahn/twisted/resource.py
```diff
@@ -34,6 +34,7 @@
     # starting from Twisted 12.2, NoResource has moved
     from twisted.web.resource import NoResource
 from twisted.web.resource import IResource, Resource
+from twisted.python.compat import _PY3
```
tardyp: This looks like a private API from twisted. why not using six?

This breaks with twisted < 12. see https://github.com/buildbot/buildbot/pull/1808

hawkowl: I'll replace it with the six version, the t.p.compat one is muscle memory for me.

Also, please don't use Twisted < 12. Buildbot should be testing against newer versions -- 12 hasn't been supported for at least three years.

PR #398, opened 2015-05-13
Pass TLS context

meejah:
This builds on @terrycojones PR #394 for adding an ``ssl=`` keyword argument to ``ApplicationRunner`` by adding a similar corresponding Twisted API, and examples using a self-signed certificate for both asyncio and Twisted.

Essentially we just export the underlying asyncio or Twisted APIs so autobahn's API becomes no more unstable that either of those.

Python's documentation recommends using ``ssl.create_default_context`` and only exposes an API that allows you to set ``ssl=`` kwarg (which is precisely what the asyncio version of this is doing).

Twisted's documentation recommends using a ``CertificateOptions``instance  as the ``sslContextFactory=`` kwarg to ``SSL4ClientEndpoint`` which is just what this code does.

Note if we merge/take this PR, #394 should be closed and remain unmerged (all terry's changes are in here, with correct author information). Or vice versa.

### DIFF
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -32,11 +32,11 @@ from autobahn.websocket.protocol import parseWsUrl
 from autobahn.asyncio.websocket import WampWebSocketClientFactory
 
 try:
-    from asyncio import get_event_loop
+    import asyncio
 except ImportError:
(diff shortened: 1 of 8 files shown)


terrycojones:
LGTM. I'm happy to close my pr, just let me know.

PR #401, opened 2015-05-19
Examples improvements

meejah:
This makes all the "basic" examples for both Twisted and asyncio:

 - more similar
 - use `ApplicationRunner` to run things
 - provide a AUTOBAHN_DEMO_ROUTER env-var to easily change where your router is
 - better instructions
 - a working, local crossbar.io configuration
 - an `index.html` that the above can serve to access the javascript demos
 - a few minor fixes (e.g. some argument names were wrong)
 - removes references to router etc


oberstet:
awesome!

PR #407, opened 2015-06-12
ensure 'extra' is a dict()

meejah:


oberstet:
Extra does not need to be a dict, but can be anything .. any reason why should we limit?

meejah:
No, not at all; that's just what the documentation said. I'll close this and change the docs instead.

Issue #409, opened 2015-06-15
ApplicationRunner should support unix-domain sockets


meejah:
Currently it's only possible to open TCP (or TLS) connections on IPv4. This should include other socket types that Autobahn supports.

We can't just export Twisted endpoints as "the" API as that can't easily be supported for the asyncio ApplicationRunner...

oberstet:
Yes, that is a major issue.

FWIW, CB lets you connect your component over Unix domain sockets as well. It can do that, as it uses the raw classes of AutobahnPython, not the ApplicationRunner convenience thing https://github.com/crossbario/crossbar/blob/master/crossbar/worker/container.py#L281

Maybe we need to define a goal first ..

hynek:
If someone could point me to code that connects to CB using unix domain sockets I’d be most grateful. :)

(and if there’s a way to use raw sockets I’d be even more delighted by any example :))

oberstet:
@hynek currently, `ApplicationRunner` lacks the necessary bits. @meejah and I are working on this. At the moment, only CB is able to host components which connect over UDS (and RawSocket and MsgPack) back to a router.

Here is an example: https://github.com/crossbario/crossbarexamples/blob/master/rest/caller_performance/.crossbar/config_multi_worker.json

meejah:
Work is continuing on https://github.com/tavendo/AutobahnPython/tree/refactor-transport and feedback is appreciated on the API; there are examples showing its use.

meejah:
Since this refactoring also involves TLS (among other) configuration, @oberstet and I had discussed "why would I want to pass a Twisted-native or asyncio-native (i.e. ssl.SSLContext) object"? Some use-cases would be:

 - self-signed certificates
 - the (default) system-wide certificate store is unsuitable (for some reason)
 - you have a client-side certificate to load + supply
 - you wish to do certificate-pinning (and/or other custom cert-chain verification)
 - you have specific requirements for cipher-selector (or other TLS options?)

om26er:
Does the component API support connecting to a unix socket ? I am working on something that requires to connect to a unix socket created by crossbar.

oberstet:
maybe we should nail this;)

fwiw, AB has a UDS aware WebSocket URL parser:

* https://github.com/crossbario/autobahn-python/blob/941600534ea28120342b39ad53c71aa920abb54c/autobahn/websocket/util.py#L125

and CB has a client connecting twisted endpoint:

[...]

PR #410, opened 2015-06-15
configurable transports

meejah:
Here is *one* idea for configuring the transports for ApplicationRunner to get some discussion going about how to support #409. For context, this is a bit bigger than *just* unix-domain sockets -- there are two transports involved:

- the "stream transport" (e.g. TCP4 vs. Unix socket)
- the "WAMP transport" (e.g. web socket vs raw socket)

Both of these can have additional options (e.g. timeout options, TLS options, etc). For those unfamiliar with Twisted, they have a concept of "endpoints" that can be configured with arbitrary strings, which is exposed here -- however simply adding an "endpoint=" kwarg to ApplicationRunner leaves asyncio in the cold.

I will post some example code using this branch with both asyncio and Twisted in the PR comments.

### DIFF
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -66,6 +66,68 @@ class ApplicationSessionFactory(protocol.ApplicationSessionFactory):
    """
 
 
+def _create_tcp4_stream_transport(loop, apprunner, wamp_transport_factory, **kw):
+    """
+    The default callable used for _create_stream_transport in
+    ApplicationRunner, this will parse the provided websocket URL and
+    connect via TCP4 (with TLS if it's a `wss://` URL).
+    """
+
+    isSecure, host, port, resource, path, params = parseWsUrl(apprunner.url)
+
+    if apprunner.ssl is None:
+        ssl = isSecure
+    else:
+        if apprunner.ssl and not isSecure:
+            raise RuntimeError(
+                'ssl argument value passed to %s conflicts with the "ws:" '
+                'prefix of the url argument. Did you mean to use "wss:"?' %
+                apprunner.__class__.__name__)
+        ssl = apprunner.ssl
+
+    coro = loop.create_connection(wamp_transport_factory, host, port, ssl=ssl)
+    return asyncio.async(coro)
+
+
+def _create_unix_stream_transport(loop, apprunner, wamp_transport_factory, **kw):
+    """
+    Connects to a Unix socket as the transport. You must pass the
+    socket path as `socket_path=` kwarg to `ApplicationRunner.run()`
+    """
+
+    try:
+        socket_path = kw['socket_path']
+    except KeyError:
+        raise RuntimeError("Pass socket_path= kwarg to ApplicationRunner.run()")
+
+    url = None  # see autobahn.websocket.protocol.WebSocketClientFactory
+
+    if apprunner.ssl is not None:
+        raise RuntimeError('ssl= argument inconsistent with Unix-domain sockets')
+
+    coro = loop.create_unix_connection(wamp_transport_factory, socket_path)
+    return asyncio.async(coro)
+
+
+def _create_websocket_wamp_factory(reactor, apprunner, session_factory, **kw):
+    # factory for using ApplicationSession
+    def create():
+        cfg = ComponentConfig(apprunner.realm, apprunner.extra)
+        try:
+            session = session_factory(cfg)
+        except Exception as e:
+            print(e)
+            asyncio.get_event_loop().stop()
+        else:
+            session.debug_app = apprunner.debug_app
+            return session
+
+    return WampWebSocketClientFactory(
+        create, url=apprunner.url, serializers=apprunner.serializers,
+        debug=apprunner.debug, debug_wamp=apprunner.debug_wamp
+    )
+
+
 class ApplicationRunner(object):
     """
     This class is a convenience tool mainly for development and quick hosting
@@ -77,7 +139,8 @@ class ApplicationRunner(object):
 
     def __init__(self, url, realm, extra=None, serializers=None,
                  debug=False, debug_wamp=False, debug_app=False,
-                 ssl=None):
+                 ssl=None,
+                 _create_stream_transport=None, _create_wamp_factory=None):
         """
         :param url: The WebSocket URL of the WAMP router to connect to (e.g. `ws://somehost.com:8090/somepath`)
         :type url: unicode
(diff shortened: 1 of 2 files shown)


meejah:
To try this branch for unix-sockets, add a suitable transport to a Crossbar configuration, such as:

```
            "transports": [
                {
                    "type": "websocket",
                    "endpoint": {
                        "type": "unix",
                        "path": "/tmp/cb-socket"
                    }
                }
            ]
```

meejah:
Here is a Twisted worker that can use this branch and connect to the above Crossbar unix-socket:

```python
from __future__ import print_function

from twisted.internet.defer import inlineCallbacks, DeferredList

from autobahn.twisted.wamp import ApplicationSession, ApplicationRunner
from autobahn.twisted.wamp import _create_unix_stream_transport, _create_endpoint_stream_transport


class ClientSession(ApplicationSession):
    @inlineCallbacks
    def onJoin(self, details):
        print("Joined", details)
        sub = yield self.subscribe(self.subscription, "test.sub")
        print("subscribed", sub)
        reactor.callLater(2, reactor.stop)

    def subscription(self, *args, **kw):
        print("sub:", args, kw)


if __name__ == '__main__':
    if True:
        runner = ApplicationRunner(None, u"realm1")
        runner.run(ClientSession, start_reactor=False, socket_path='/tmp/cb-socket')
        reactor.run()

    else:
        # here's with a Twisted endpoint parser, so we can also pass                                                                               
        # more options; I'm just doing another Unix transport here,                                                                                
        # but of course this could be anything at all that's a valid                                                                               
        # Twisted endpoint...
        runner = ApplicationRunner(None, u"realm1", _create_stream_transport=_create_endpoint_stream_transport)
        runner.run(ClientSession, start_reactor=False, endpoint="unix:/tmp/cb-socket:timeout=10")
        reactor.run()

    print("exiting.")
```

meejah:
Here is a more-or-less equivalent asyncio worker:

```python
from __future__ import print_function

import asyncio

from autobahn.asyncio.wamp import ApplicationSession, ApplicationRunner
from autobahn.asyncio.wamp import _create_unix_stream_transport


class ClientSession(ApplicationSession):
    @asyncio.coroutine
    def onJoin(self, details):
        print("Joined", details)
        sub = yield from self.subscribe(self.subscription, "test.sub")
        print("subscribed", sub)

    def subscription(self, *args, **kw):
        print("sub:", args, kw)


if __name__ == '__main__':
    if True:
        # could add kwarg "_create_stream_transport=_create_unix_stream_transport" for clarity ...
        runner = ApplicationRunner(None, u"realm1")
        runner.run(ClientSession, socket_path='/tmp/cb-socket')

    print("exiting.")

```

meejah:
For *just* unix-domain support, of course we can simply *not* expose the `_create_stream_transport` etc. kwargs at all in `__init__()`.

I also don't particularly like the "arbitrary kwargs to `run()`" that get passed to the underlying transport creators, but I also don't have any more-compelling suggestions at the moment.

Feedback very much encouraged!

`ApplicationRunner` needs to be easy to use for the usual case, but of course also needs to support more than just TCP4/TLS and a single WAMP transport option going forwards...

estan:
I also think the passing through of feels a little like leaky abstractions, but have no good idea either. Unless you build some big thing to really abstract the asyncio/Twisted ways. That would give the most convenient API I think, but would be more work.

Regarding the examples, in the Twisted one you make use of `_create_endpoint_stream_transport`, but I cannot find this in the changed `autobahn/twisted/wamp.py`, only the `_create_unix_stream_transport` and `_create_tcp4_stream_transport` (BTW, is tcp6 not available?).

Also, with these changes, should these factory methods not be made public API? (underscore removed).

meejah:
Hi @estan thanks for the feedback!

There is a new branch I'm working on with different ideas. The underscores were to indicate "private API" -- the API is still supposed to be ApplicationRunner and .run() primarily. (Also, this wasn't really intended to be merged; was mockup of an idea).

See https://github.com/meejah/AutobahnPython/tree/issue409-transports-from-dicts-rebase for latest ideas (this also includes some stuff around re-connecting as well). It still needs a bunch of cleanup among other things if will be merged...

estan:
Aha, now I see. You were only using those private methods to illustrate how to use it.

Thanks for the link, though it takes me to the front page of your repo fork (are issues not publicly available at that fork page?).

meejah:
@estan nothing should be private in that repo; I was trying to link to a branch (it doesn't have a PR yet). There are a couple branches starting with "issue409-" in my fork...

estan:
Aha, my bad, I misread the URL and expected to end up at an issue or pull request or something. I'll take a look at the branch. Thanks.

oberstet:
I think, the hard part is the API design. What do we want to achieve?

Do we want:

a) keep `ApplicationRunner` agnostic/neutral (API wise) w.r.t. network framework (Twisted/asyncio)
b) have 2 flavors of `ApplicationRunner` for Twisted and asyncio, with different API (e.g. expose Twisted endpoints, SSL contexts, ..)

Do we want:

c) let the user specify exactly one WAMP transport
d) let the user specify a list of WAMP transports, and the underlying implementation automatically tries/chooses from these, e.g. until a working one is found (this is what AutobahnJS provides)

-----

a) would follow the approach we take for `ApplicationSession`: network framework neutral (a user can in principle write code that is source compatible with both the network frameworks).

 A user of `ApplicationRunner` could describe the desired transport (underlying stream transport, and WAMP framing/serializer above) similar to Crossbar.io configs (`{"type": "websocket", "endpoint": {"type": "tcp", ...`).

Based on the user config, the `ApplicationRunner` **implementation** would then create a suitable transport using the facilities of the respective networking framework.

When a Twisted user wants to use Twisted specific things like Twisted endpoints or SSL context _objects_ directly with Autobahn classes, then the user needs to dig below the `ApplicationRunner` level to more basic classes.

Actually, this was the original motivation for `ApplicationRunner`: I got tired of the roughly 30-50 lines boilerplate to boot up everything from atomic classes, and instead wanted the "ease of use" from AutobahnJS also here.

-----

b) would bite the bullet, and just expose network framework specific features in the **API** of `ApplicationRunner`.

That is, have `ssl` (Twisted SSL context) and `endpoint` (Twisted stream endpoint) be exposed in either `a.t.wamp.ApplicationRunner.__init__` or `a.t.wamp.ApplicationRunner.run`.

Do we expose Twisted endpoints as strings or objects?

Variant 1 (endpoint strings):

```python
endpoint=b"tcp:host=www.example.com:port=80"
runner = ApplicationRunner(endpoint=endpoint, realm=u"realm1")
```

Variant 2 (endpoint objects)

```python
endpoint = clientFromString(reactor, b"tcp:host=www.example.com:port=80")
runner = ApplicationRunner(endpoint=endpoint, realm=u"realm1")
```

The latter has the advantage of solving the SSL problem also:

```python
endpoint = SSL4ClientEndpoint(reactor, 'www.example.com', 80, ssl_context)
runner = ApplicationRunner(endpoint=endpoint, realm=u"realm1")
```

meejah:
In code, I think the right API for Twisted endpoints is to expose the endpoints as objects. This is most flexible. Then users of Autobahn can easily accept endpoint-strings from users -- or provide some other API (e.g. GUI) -- their choice. I think error-handling is more clear this way, too.

oberstet:
So if we decided to expose native Twisted endpoint objects, then that would allow you to eg run wamp over Tor (what you recently mentioned)?  And it would also do away with exposing ssl context?

So: could we get away with one Twisted specific parameter (endpoint object)? What else would Twisted users expect?

meejah:
Ah, was updating my comment.

As far as other Twisted things, there's also ``twisted.internet.task.react`` which is a nice way to run clients (or, things that don't expect to run forever). You can use this currently with ApplicationRunning by passing ``start_reactor=False``. Like:

```python
@inlineCallbacks
def main(reactor):
    yield ApplicationRunner(...).run(FirstComponent, start_reactor=False)
    yield ApplicationRunner(...).run(SecondComponent, start_reactor=False)
    # whatever else ...
task.react(main)
```

meejah:
As for Tor, yes if you either expose objects *or* some way to pass endpoint-strings -- and have txtorcon installed -- you could pass "tor:timaq4ygg2iegci7.onion:9000" to connect via Tor to the hidden-service that hosts txtorcon's documentation (if there were a crossbar.io running there ;) ). Or, you could constuct ``txtorcon.TorClientEndpoint``.

Issue #419, opened 2015-06-23
Fix WAMP ID generation


oberstet:
* https://github.com/tavendo/WAMP/blob/master/spec/basic.md#ids
* https://en.wikipedia.org/wiki/Birthday_problem#Approximations

```python
import random

MAX = 2**31
#MAX = 2**53
COUNT = 2**20
#COUNT = 10000

def gen_id():
   return random.randint(0, MAX)

so_far = {}
collisions = []
c_max = 0
c_min = MAX
c_sum = 0
for x in xrange(COUNT):
    i = gen_id()
    if i in so_far:
        distance = x - so_far[i]
        if distance > c_max:
            c_max = distance
        if distance < c_min:
            c_min = distance
        c_sum += distance
        collisions.append(distance)
        #print("collision: {} in set, id={}, distance={}".format(len(so_far), i, distance))
    else:
        so_far[i] = x

if len(collisions):
    print("found {} collisions in {}".format(len(collisions), COUNT))
    print("min distance = {}, max distance = {}, avg distance = {}".format(c_min, c_max, float(c_sum)/float(len(collisions))))
else:
    print("no collisions found!")
```

oberstet:
```console
(pypy260_1)oberstet@thinkpad-t430s:~/scm/scratchbox$ python test_randint.py 
found 263 collisions in 1048576
min distance = 4994, max distance = 1046004, avg distance = 310519.163498
(pypy260_1)oberstet@thinkpad-t430s:~/scm/scratchbox$ python test_randint.py 
found 256 collisions in 1048576
min distance = 3325, max distance = 953221, avg distance = 352879.664062
(pypy260_1)oberstet@thinkpad-t430s:~/scm/scratchbox$ python test_randint.py 
found 257 collisions in 1048576
min distance = 128, max distance = 988844, avg distance = 340396.346304
(pypy260_1)oberstet@thinkpad-t430s:~/scm/scratchbox$ python test_randint.py 
found 248 collisions in 1048576
min distance = 1731, max distance = 987363, avg distance = 325821.637097
(pypy260_1)oberstet@thinkpad-t430s:~/scm/scratchbox$ python test_randint.py 
found 265 collisions in 1048576
min distance = 832, max distance = 992627, avg distance = 384565.124528
(pypy260_1)oberstet@thinkpad-t430s:~/scm/scratchbox$ python test_randint.py 
found 267 collisions in 1048576
[...]

PR #420, opened 2015-06-24
sequential id generation

meejah:
Adds a sequential ID generator in BaseSession (that wraps at 2**53) to provide IDs for all the Request (session-scoped IDs) replacing use of util.id() for these IDs. Issue #419

### DIFF
--- a/autobahn/twisted/rawsocket.py
+++ b/autobahn/twisted/rawsocket.py
@@ -120,6 +120,7 @@ class WampRawSocketProtocol(Int32StringReceiver):
                 self._session.onMessage(msg)
 
         except ProtocolError as e:
+            log.msg(str(e))
             if self.factory.debug:
                 log.msg("WampRawSocketProtocol: WAMP Protocol Error ({0}) - aborting connection".format(e))
             self.abort()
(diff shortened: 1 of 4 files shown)


oberstet:
Looks good. Just one thing: could you replace the  generator by something simple? It doesn't add value, and since this is a hot function, I am wary of the performance impacts.

Something like in `BaseSession`:

```python
def __init__(self, ..):
    self._last_request_id = -1
```
and

```python
def _next_request_id(self):
     self._last_request_id += 1
     if self._last_request_id > 2**53: # replace 2**53 with constant? not sure how smart CPy/PyPy are
         self._last_request_id = 0
     return self._last_request_id
```

meejah:
Okay, changed it like that; make sense. Also, I was starting the IDs at 1, not 0 like the spec says.

The generator *should* be the same, but yeah it's not as obvious and I don't actually know if there's overhead. generator might use slightly more memory too...

meejah:
(Note that I just force-pushed the new, changed comits over this same branch...)

oberstet:
@meejah perfect. thanks! yeah, generators ... minimize "unknowns" in this case is good I guess

meejah:
Yeah, good call on eliminating needless complexity ;) (it's now way easier to read the code).

PR #424, opened 2015-07-08
#409: transport config via dicts + Connection abstraction

meejah:
A mostly-complete idea for a cross-asyncio/Twisted API
to do transport configuration, including examples. To see
the different ways of using the API, look at the Twisted
example examples/twisted/wamp/unix-rawsocket There's a basic
asyncio example as well.

Features:

 - Connection abstraction, similar to JS library
 - subscribe to transitions (transport, session up/down)
 - interface with Twisted endpoints via strings and/or IStreamClientEndpoint
 - asyncio implementation
 - access to "native" protocol/transport objects
 - play nicely with twisted.internet.task.react
 - will work with retry/reconnect logic

Still needs:

 - tests
 - cleanups, to XXX comments
 - change ApplicationRunner.run() API even more? (see comments)
 - are TLS options sufficient, or go to recent "ssl=" kwarg?
 - "listener" API on Connection: good, bad?
 - is keeping wamp/runner.py things generic a stretch?
 - port "endpoint" helpers from crossbar.io?


meejah:
Note that the "diff view" makes this kind of hard to look at. Probably best to first look at the ``examples/twisted/wamp/unix-rawsocket`` example to get an idea of the API and then look at ``autobahn/twisted/wamp.py`` (or the asyncio equivalent) and go from there.

I attempted to make at least some of the helper code generic, in ``autobahn/wamp/transport.py`` and ``autobahn/wamp/runner.py``

autobahn/wamp/runner.py
```diff
@@ -0,0 +1,416 @@
...
+        # our event listeners
+        self._event_listeners = {
+            self.ERROR: [],
+            self.CREATE_SESSION: [],
+            self.SESSION_LEAVE: [],
+            self.CONNECTED: [],
+            self.CLOSED: [],
+        }
+
+        # generate a new transport to try
```
isra17: Isn't just an [itertools.cycle](https://docs.python.org/2/library/itertools.html#itertools.cycle)?

meejah: Yes, that's probably more clear in this case, thanks :)

oberstet:
Looks like a good start! I've merged this and current trunk on https://github.com/tavendo/AutobahnPython/tree/refactor-transport and now am working from there. @meejah could you join working / tracking above branch? Please drop this PR then, and .. what was the status of https://github.com/tavendo/AutobahnPython/pull/410 again? I had a quick look into the changes again .. isn't that already adressed now in this PR?

meejah:
Yes, #410 is basically dead (it was a different way to do what's in this PR/branch).

oberstet:
this has been merged to https://github.com/tavendo/AutobahnPython/tree/refactor-transport - work continues there before we merge to master ..

PR #425, opened 2015-07-13
rename _log to _ab_log so that Twisted 15.3 works

hawkowl:
Twisted 15.3 adds a _log component on Protocols, and Autobahn overwrites this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ egg\-info
 egg
 .egg*
 .tox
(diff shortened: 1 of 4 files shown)


oberstet:
what's the problem with overwriting? in general, we can't change private attributes each time twisted decides to introduce some private attr.

hawkowl:
Got an alternate solution

PR #428, opened 2015-07-19
Replace skip skipTest in tests

Jenselme:
According the documentation of [unittest](), there is no skip method in
class TestCase. The intended behaviour must be achieved with
[skipTest](https://docs.python.org/2/library/unittest.html#unittest.TestCase.skipTest)


meejah:
Thanks, this looks good.

/cc @oberstet

PR #430, opened 2015-07-23
Logger shim to work around Twisted 15.3

hawkowl:


autobahn/logger.py
```diff
@@ -0,0 +1,37 @@
...
+        # If we've been asked for the Twisted logger, try and get the new one
+        try:
+            from twisted.logger import Logger
+            return PassthroughLogger(Logger())
+        except ImportError:
+            pass
+
+    from logging import getLogger
+    return PassthroughLogger(getLogger())
+
```
meejah: Why do we need the pass through logger instead of just returning the logger instance?

autobahn/websocket/protocol.py
```diff
@@ -2146,11 +2146,11 @@ def sendCloseFrame(self, code=None, reasonUtf8=None, isReply=False):
         """
         if self.state == WebSocketProtocol.STATE_CLOSING:
             if self.debugCodePaths:
-                self.factory._log("ignoring sendCloseFrame since connection is closing")
```
meejah: I  wonder if (hope!) we can just nuke debugCodePaths (and just rely on log.debug)...? @oberstet?

hawkowl: This might end up trashing other people's logs so I'm not sure.

meejah: Ahh, hmm yeah, like if they had .debug()s as well you mean?

Certainly this whole PR is a step forward for Autobahn's logging, so bogging it down trying to *also* fix the .debugCodePaths, .debug_wamp etc things is silly; sorry for the noise ... but I'm still confused about what ``PassthroughLogger`` is really accomplishign.

hawkowl: More that people might not have debug turned off, and we'll end up inadvertantly filling up logs.

The PassthroughLogger will gain a new later function (`.trace()`) which will then call debug. This patch just gets the groundwork in.

hawkowl:
Hey @oberstet the 15.3 release is soon, so we need to visit this so that it doesn't break Autobahn...

PR #451, opened 2015-08-14
More twisted tox versions + codecov

hawkowl:


hawkowl:
cc @oberstet, this is ready for review

tox.ini
```diff
@@ -1,180 +1,47 @@
 [tox]
-envlist = flake8,
-          py26twisted,
-          py26asyncio,
```
oberstet: isn't this missing after the change?

hawkowl: gah yes
I wanted to remove py26-twtrunk, as Twisted trunk no longer supports 2.6. Will fix.

tox.ini
```diff
@@ -1,180 +1,47 @@
...
-   twisted
-   unittest2
-commands =
-   sh -c "which python"
-   sh -c "which trial"
-   python -V
-   trial --version
-   trial autobahn
-basepython = python2.6
-install_command = pip install {packages} autobahn[twisted,serialization]
```
oberstet: applies to all envs: `serialization` is an extra that eg installs msgpack. without that, msgpack WAMP serialization won't be tested (only JSON). is that extra install dep lost?

hawkowl: Yep, it installs it explicitly at line 42.

tox.ini
hawkowl: OK, `py26-trollius` is restored.

tox.ini
hawkowl: Or rather, no, it's not lost, it's installed explicitly at line 42.

oberstet: I'm too dumb .. where is it? https://github.com/hawkowl/AutobahnPython/blob/more-twisted-tox-versions/tox.ini#L42 doesn't have it ..

hawkowl: Erm, seems like I'm the dumb one here actually -- this should be the URL: https://github.com/hawkowl/AutobahnPython/blob/more-twisted-tox-versions/tox.ini#L14

hawkowl:
Okay, travis (with `py26-trollius`) seems to have passed fine @oberstet

tox.ini
oberstet: Ahh, right. That's fine As long as we keep that in sync with https://github.com/tavendo/AutobahnPython/blob/master/setup.py#L123

PR #452, opened 2015-08-14
Handle exceptions better on Py2

### LINKED ISSUE #435: No handlers could be found for logger "trollius"
Most likely it's trollius bug, but may be you advise how resolve this issues.

This occur when I try launch example (examples/asyncio/websocket/slowsquare, server_py2.py and client.py)
Issue arise when I launch the client while server_py2.py already launch. Message(No handlers could be found for logger "trollius") appears in console(server_py2.py) and interaction between client and server stops. This behavior doesn't repeat always, but often.

hawkowl:
Fixes #435. Also includes #451.

### DIFF
--- a/autobahn/logger.py
+++ b/autobahn/logger.py
@@ -28,6 +28,8 @@ from __future__ import absolute_import
 
 import logging
 
+
+# A logging handler for Trollius that prints everything out
 class PrintHandler(logging.Handler):
     def emit(self, record):
         print(record)
--- a/autobahn/logger.py
+++ b/autobahn/logger.py
@@ -26,6 +26,16 @@
 
 from __future__ import absolute_import
 
+import logging
+
+class PrintHandler(logging.Handler):
+    def emit(self, record):
+        print(record)
+
+
+h = PrintHandler()
+logging.getLogger("trollius").addHandler(h)
+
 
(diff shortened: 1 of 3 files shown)


hawkowl:
@oberstet this should fix up the issue in the linked ticket.

hawkowl:
```
Request to square 5.161147993505438 sent.
Exception raised: number too large
```
...with this Autobahn installed and running `python server_py2.py`.

oberstet:
@hawkowl that exception is excepted .. it's a user-defined WAMP exception

hawkowl:
@oberstet I don't understand -- that's what it should say (as it's a user defined one), right?
Py34 has that error printed, but 2.7 drops the connection.

oberstet:
I just merged https://github.com/tavendo/AutobahnPython/pull/451 .. this PR seems to have some intersection with the former. Which makes it hard to see what is new .. could you probably update this PR so it only contains new stuff?

hawkowl:
@oberstet I've done so, but it might need a refresh -- I just rebased and repushed.

oberstet:
I think the new behavior of `sendClose` (coercing/truncating instead of bailing out) is better from a user point of view. Merging ..

PR #468, opened 2015-08-28
Remove unused self.make variable in asyncio.wamp.ApplicationRunner

terrycojones:
Another extremely trivial clean up.


oberstet:
ah, yes. should be removed from here as well: https://github.com/tavendo/AutobahnPython/blob/master/autobahn/twisted/wamp.py#L132

terrycojones:
@oberstet Removed in the twisted code too, thanks.

PR #489, opened 2015-09-08
Change way how log messages are created.

inirudebwoy:
Hi, 

I have been seeing this error message when calling connectionLost.

```python
  File "/home/majki/projects/iep-id-mapper/python/idmapper/imagination/idmapper/output/websocket/management/commands/broadcast_server.py", line 51, in connectionLost
    WebSocketServerProtocol.connectionLost(self, reason)
  File "/home/majki/projects/AutobahnPython/autobahn/twisted/websocket.py", line 104, in connectionLost
    peer=self.peer)
  File "/usr/lib64/python2.7/logging/__init__.py", line 1148, in debug
    self._log(DEBUG, msg, args, **kwargs)
exceptions.TypeError: _log() got an unexpected keyword argument 'peer'
```

Code that fails is 

```python
 self.factory.log.debug("Connection to/from {peer} was closed cleanly", peer=self.peer)
```

This looked like a mistake in code, however it happened in few places, and looked like intended action. Please have a look at this PR and if for some reason this is not an issue explain how this would work.

### DIFF
--- a/autobahn/twisted/websocket.py
+++ b/autobahn/twisted/websocket.py
@@ -100,12 +100,12 @@ class WebSocketAdapterProtocol(twisted.internet.protocol.Protocol):
 
     def connectionLost(self, reason):
         if isinstance(reason.value, ConnectionDone):
-            self.factory.log.debug("Connection to/from {peer} was closed cleanly",
-                                   peer=self.peer)
+            self.factory.log.debug(
+                "Connection to/from {peer} was closed cleanly".format(peer=self.peer))
 
         elif isinstance(reason.value, ConnectionAborted):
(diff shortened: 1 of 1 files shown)


oberstet:
No, it's correct. This is brand new structured logging:

https://github.com/tavendo/AutobahnPython/blob/master/autobahn/twisted/websocket.py#L227
https://github.com/tavendo/AutobahnPython/blob/master/autobahn/logger.py

meejah:
It looks like a bug when we're using the not-Twisted logger...?

inirudebwoy:
OK, fair enough. I can't reproduce the problem.

PR #490, opened 2015-09-08
use public Twisted API, not private

### LINKED ISSUE #487: Import from private Twisted module breaks Autobahn on Windows
When running on Windows with Twisted 15.4.0, the changes made in c517e9f generate an exception, since importing ```PipeAddress``` from ```twisted.internet._posixstdio``` calls ```detectLinuxBrokenPipeBehavior```, which only works on Posix systems.

Removing the changes from this commit fixes the problem, and seems to only make a cosmetic change.  Arguably, importing from a private module is fraught with peril, and probably shouldn't be done.

Here is an example stack trace showing the problem:
```
  File "C:\Python27\Scripts\crossbar-script.py", line 9, in <module>
    load_entry_point('crossbar==0.10.4', 'console_scripts', 'crossbar')()
  File "C:\Python27\lib\site-packages\crossbar\controller\cli.py", line 665, in run
    options.func(options)
  File "C:\Python27\lib\site-packages\crossbar\controller\cli.py", line 402, in run_command_start
    from crossbar.controller.node import Node
  File "C:\Python27\lib\site-packages\crossbar\controller\node.py", line 51, in <module>
    from crossbar.controller.process import NodeControllerSession
  File "C:\Python27\lib\site-packages\crossbar\controller\process.py", line 52, in <module>
    from autobahn.twisted.util import sleep
  File "C:\Python27\lib\site-packages\autobahn\twisted\util.py", line 31, in <module>
    from twisted.internet._posixstdio import PipeAddress
  File "C:\Python27\lib\site-packages\twisted\internet\_posixstdio.py", line 16, in <module>
    from twisted.internet import process, error, interfaces
  File "C:\Python27\lib\site-packages\twisted\internet\process.py", line 119, in <module>
    brokenLinuxPipeBehavior = detectLinuxBrokenPipeBehavior()
  File "C:\Python27\lib\site-packages\twisted\internet\process.py", line 107, in detectLinuxBrokenPipeBehavior
    reads, writes, exes = select.select([w], [], [], 0)
select.error: (10038, 'An operation was attempted on something that is not a socket')
```

meejah:
This should fix #487

### DIFF
--- a/autobahn/twisted/util.py
+++ b/autobahn/twisted/util.py
@@ -28,7 +28,7 @@ from __future__ import absolute_import
 
 from twisted.internet.defer import Deferred
 from twisted.internet.address import IPv4Address, UNIXAddress
-from twisted.internet._posixstdio import PipeAddress
+from twisted.internet.stdio import PipeAddress
 
 try:
     from twisted.internet.address import IPv6Address


haroldfredshort:
That change fixes it for me on Windows 10, thanks.

PR #491, opened 2015-09-09
default implementation for onConnect/onChallenge

meejah:
While looking at https://github.com/crossbario/crossbar/issues/412 there's a lot of ``onChallenge`` code to get right in order for authenticated login to work.

This adds a default ``onChallenge`` (and extends the default ``onConnect``) method to make authentication much easier: you simply add a 'wamp_cra' section to your ``ApplicationSession``'s ``config.extra``, which itself is a dict containing 'secret' and 'user'. Like:

```
extra = {
    'wamp_cra': {
        'user': 'meejah',
        'secret': 'cr0ssb4r',
    }
    # whatever else your app uses
}
ApplicationSession(..., extra=extra)
```

If ``extra`` is not a dict, or doesn't have ``wamp_cra`` in it, ``onConnect`` works as before (i.e. you have to write your own, and your own ``onChallenge``).

### DIFF
--- a/autobahn/wamp/protocol.py
+++ b/autobahn/wamp/protocol.py
@@ -45,6 +45,7 @@ from autobahn.wamp import uri
 from autobahn.wamp import message
 from autobahn.wamp import types
 from autobahn.wamp import role
+from autobahn.wamp import auth
 from autobahn.wamp import exception
 from autobahn.wamp.exception import ApplicationError, ProtocolError, SessionNotReady, SerializationError
 from autobahn.wamp.types import SessionDetails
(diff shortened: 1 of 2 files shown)


oberstet:
this is introducing new public API. we need to carefully consider this. won't merge for now ..

meejah:
Yes, fair enough. Mostly I was motivated by "the example code didn't quite work, so I should put this code I *did* get working somewhere". Perhaps I should instead merge this to the ``refactor_transport`` branch...? (But, that's already big :/ )

p.s. PR coming for the documentation repo...

PR #498, opened 2015-09-11
Do some Flake8 cleanups

hawkowl:


oberstet:
looks good. might break some stuff down the line, but that's expected. need get through this ..

oberstet:
ah, the make_logger will come from txaio as soon as we have it there ...

PR #502, opened 2015-09-15
Remove Hixie-76 support from Autobahn

hawkowl:
It's disabled by default, it's outdated, it's insecure, and it needlessly complicates Autobahn's protocol implementation. Removing it is a good first step for cleaning it all up.


hawkowl:
https://github.com/tavendo/AutobahnPython/pull/502/files?w=1 will give a lot better diff...

oberstet:
;) Yes, it's time to move on. All that cruft is a drag.

I have announced the removal: https://groups.google.com/forum/#!topic/autobahnws/dFXEjsr9ILA

There are some more bits in examples and `autobahn.twisted`: `find . -name "*.py" -exec grep -Hi "hixie" {} \;`

Maybe fix these in the same PR?

hawkowl:
@oberstet Sure, I'll find them and clean them out. I'm also writing a test to verify that clients handshaking with Hixie will get a handshake failure.

PR #509, opened 2015-09-16
fix reference to datetime.datetime

hawkowl:
This breaks Crossbar: https://travis-ci.org/crossbario/crossbar/jobs/80563907


meejah:
:+1:

PR #518, opened 2015-09-21
Transport configuration checking code

meejah:
Ported from refactor-transports branch; includes unit-tests and
split out "native" (asyncio/twisted) checking code.


oberstet:
There are some thing in here I would merge, and others not.

PR #520, opened 2015-09-23
Rely on __str__() to extract exception information

### LINKED ISSUE #484: INVOCATION endpoint error handler fails if exception has an empty args attribute
[protocol.py l.805](https://github.com/tavendo/AutobahnPython/blob/4aec0c539d0bd50b526c441d5e652fd2447bde33/autobahn/wamp/protocol.py#L805) throws an UNHANDLED IndexError in that case.

jvdm:
For the purpose of extracting exception messages from possible user-defined exceptions (which is the case of #484), the best approach IMHO is relying on `__str__()`.

This patch does this when handling exceptions from INVOCATION endpoints.

Fix #484.

### DIFF
--- a/autobahn/wamp/protocol.py
+++ b/autobahn/wamp/protocol.py
@@ -604,7 +604,7 @@ class ApplicationSession(BaseSession):
 
                         def error(err):
                             # errmsg = 'Failure while invoking procedure {0} registered under "{1}: {2}".'.format(endpoint.fn, registration.procedure, err)
-                            errmsg = u"{0}".format(err.value.args[0])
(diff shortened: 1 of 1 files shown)


meejah:
Thanks.

Yes, str() is a good solution however the plan with autobahn logging is to use the newly-landed txaio logging API, so it will look something like: https://github.com/meejah/AutobahnPython/blob/txaio-logging-rebase/autobahn/wamp/protocol.py#L606

jvdm:
Got it. I'll close the pull request then.

Thank you for your thoughtful comment.

PR #521, opened 2015-09-24
Fix Python 2.6 support and some txaio logging switch stuff

hawkowl:


autobahn/twisted/choosereactor.py
```diff
@@ -67,9 +67,9 @@ def install_optimal_reactor(verbose=False):
             try:
                 from twisted.internet import kqreactor
                 kqreactor.install()
-            except Exception:
+            except:
                 log.failure(("Running on *BSD or MacOSX, but cannot install kqueue Twisted reactor\n"
-                             "{log_failure.getTraceback()}"))
+                             "{log_failure.value}"))
             else:
```
meejah: Since we're in a "twisted" package, this probably doesn't matter, but ``txaio.ILogger`` doesn't actually have a .failure() method. Formatting a traceback from a txaio.IFailuedFuture is ``txaio.failure_format_traceback(fail)``. But above we're just in an except, so maybe:

```python
log.critical("Running on ... Twisted reactor")
log.debug(traceback.format_exc())
```

Also, why did you switch to a bare "except:" instead of "except Exception"?

PR #524, opened 2015-09-24
Newapi joined left  service

meejah:


oberstet:
This again changes the public API that we wanted to try: Connection does not take a session instance.

PR #529, opened 2015-10-19
Fix client auto-ping

krrr:
There is a auto-ping option in client factory, but client never start auto-ping. Is it a bug?

### DIFF
--- a/autobahn/websocket/protocol.py
+++ b/autobahn/websocket/protocol.py
@@ -3688,6 +3688,11 @@ class WebSocketClientProtocol(WebSocketProtocol):
             self.current_frame = None
             self.websocket_version = self.version
 
+            # automatic ping/pong
+            #
+            if self.autoPingInterval:
+                self.autoPingPendingCall = txaio.call_later(self.autoPingInterval, self._sendAutoPing)
+
             # we handle this symmetrical to server-side .. that is, give the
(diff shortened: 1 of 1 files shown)


jegger:
Can confirm that this fixes the issue. (and the issue exists)

meejah:
Looks good, thanks for the fix!
...but I haven't tried running this yet. I will do so shortly.

meejah:
@jegger or @krrr can you describe what the issue is (i.e. what do I need to do to repeat it)? I do see existing code to start the auto-ping when the handshake has completed (line 2890 in ``succeedHandshake``).

krrr:
@meejah The code you said is in `WebSocketServerProtocol.succeedHandshake`, and server is ok. But the problem is that client do not auto-ping. I saw corresponding options in factory of client, so I thought client has this feature

jegger:
The high-level functionality wich this enables for me, is the that the client recognizes a connection loss.

meejah:
@krrr ah, okay, client-side. Sorry about that. I'll write a test for that side too :)

meejah:
Okay, I've included @krrr's fix in #544 along with unit-tests for both the client and server cases.
Thanks for the patch!

(If the other looks fine, I'll close this too).

oberstet:
merged via other PR

PR #530, opened 2015-10-19
Support WSS + proxy

shuckc:
Hi. This patch allows the connectWS() method to supports proxy + WSS:// combination. It adds a new method startTLS() to the network layer, which is called after plain-text proxy interaction concludes. The twisted implementation delegates to transport.startTLS().

Additionally the SSL context passed into connectWS needs to be preserved until startTLS is called, so poke it into the factory.

When testing with Squid I found it (perhaps erroneously) returns a HTTP/1.0 response to the connect method, so this patch also relaxes that check.

### DIFF
--- a/autobahn/twisted/websocket.py
+++ b/autobahn/twisted/websocket.py
@@ -218,7 +218,10 @@ class WebSocketClientProtocol(WebSocketAdapterProtocol, protocol.WebSocketClient
 
     def _onConnect(self, response):
         self.onConnect(response)
-
+    def startTLS(self):
+        self.log.debug("Starting TLS upgrade")
+        self.transport.startTLS(self.factory.contextFactory)
+        self.startHandshake()
 
 class WebSocketAdapterFactory(object):
     """
@@ -515,17 +518,17 @@ def connectWS(factory, contextFactory=None, timeout=30, bindAddress=None):
     else:
         from twisted.internet import reactor
 
+    if factory.isSecure:
+        if contextFactory is None:
+            # create default client SSL context factory when none given
+            from twisted.internet import ssl
+            contextFactory = ssl.ClientContextFactory()
+
     if factory.proxy is not None:
-        if factory.isSecure:
-            raise Exception("WSS over explicit proxies not implemented")
-        else:
-            conn = reactor.connectTCP(factory.proxy['host'], factory.proxy['port'], factory, timeout, bindAddress)
+        factory.contextFactory = contextFactory
+        conn = reactor.connectTCP(factory.proxy['host'], factory.proxy['port'], factory, timeout, bindAddress)
     else:
         if factory.isSecure:
-            if contextFactory is None:
-                # create default client SSL context factory when none given
-                from twisted.internet import ssl
-                contextFactory = ssl.ClientContextFactory()
             conn = reactor.connectSSL(factory.host, factory.port, factory, contextFactory, timeout, bindAddress)
         else:
             conn = reactor.connectTCP(factory.host, factory.port, factory, timeout, bindAddress)
--- a/autobahn/twisted/websocket.py
+++ b/autobahn/twisted/websocket.py
@@ -218,10 +218,12 @@ class WebSocketClientProtocol(WebSocketAdapterProtocol, protocol.WebSocketClient
 
     def _onConnect(self, response):
         self.onConnect(response)
+
     def startTLS(self):
         self.log.debug("Starting TLS upgrade")
         self.transport.startTLS(self.factory.contextFactory)
         self.startHandshake()
+
 
 class WebSocketAdapterFactory(object):
     """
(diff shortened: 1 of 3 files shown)


shuckc:
Clean sweep from travis https://travis-ci.org/crossbario/autobahn-python/builds/86455793

autobahn/twisted/websocket.py
```diff
@@ -515,17 +520,17 @@ def connectWS(factory, contextFactory=None, timeout=30, bindAddress=None):
     else:
         from twisted.internet import reactor
 
+    if factory.isSecure:
+        if contextFactory is None:
+            # create default client SSL context factory when none given
+            from twisted.internet import ssl
+            contextFactory = ssl.ClientContextFactory()
+
```
meejah: Shouldn't this be ``optionsForClientTLS(factory.host)`` to get browser-like verification?

shuckc:
I would encourage Autobahn to use `optionsForClientTLS(...)` as a sensible default when a contextFactory is not otherwise provided. The problem is that it would be a breaking change for users with a deficient local ca-store, so it is entirely _inappropriate_ to bring that in by the back door with this change. 

While the diff shows I added `ssl.ClientContextFactory()` it is simply existing behavior being relocated up a few lines.

meejah:
Yes, okay, fair enough.

@oberstet I think we should change the default for Twisted 14.x+ to use ``optionsForClientTLS`` though, so I filed #537

meejah:
Also: works for me with Polipo.

meejah:
@hawkowl does adding ``startTLS`` have wider implications?

shuckc:
Can I assist with this discussion? Perhaps if we had a twisted unit test that sets up a basic CONNECT-style proxy and an ssl-enabled wss server that might convince you... ?

meejah:
@shuckc I don't see a problem, but I was interested in @hawkowl's feedback regarding the addition of ``startTLS``. @oberstet?

If possible, I'd rather not have "real" networking done in the unit-tests but any additional test-coverage is certainly welcome :)

hawkowl:
You can do startTLS tests using fake transports -- it's just data after all.

I don't think this is a bad change, overall -- I'd prefer the startTLS method not call the handshake itself, though.

meejah:
@hawkowl like this? https://github.com/crossbario/autobahn-python/compare/master...meejah:issue530-proxy-wss?expand=1

autobahn/websocket/protocol.py
```diff
@@ -3394,7 +3394,7 @@ def processProxyConnect(self):
             # HTTP version
             #
             http_version = sl[0].strip()
-            if http_version != "HTTP/1.1":
+            if http_version not in ("HTTP/1.1", "HTTP/1.0"):
                 return self.failProxyConnect("Unsupported HTTP version ('%s')" % http_version)
```
meejah: @oberstet any concerns with adding 1.0 here?

hawkowl: If we're going to support proxies, we'll need to support HTTP/1.0. A whole bunch of proxies (Squid, as mentioned, for example) only support HTTP/1.0, until very recently.

shuckc: Once the CONNECT exchange concludes, the protocol is effectively restarted with the destination server, so this doesn't affect the HTTP version of the onward connection.

meejah:
Okay, convinced. Shall I do a new PR with the ``startHandshake()`` change?

(Alternatively, @shuckc could cherry-pick 46ec9e0 over to this PR, or just make a new commit himself  :)

oberstet:
@meejah does http 1.0 even allow the "upgrade connection" header that websocket is based on? If so, I am fine.

The "poking into factory" is sth that is questionable. I admit, we are doing that also in various places. And its wrong. Eg doesnt blend well with slots and the like. 

But in general, yes, supporting proxies over tls seems useful.

shuckc:
Good spot, startHandshake() can stay in the protocol.py, let me test that locally and I'll push another change if it works.

shuckc:
Works with `startHandshake()` put back in websocket/protocol.py.

Once question: what do you want to happen to the alternative "asyncio" back end when proxy+wss combination is chosen? At the moment I guess it has changed from an error message as connect time to a failure to find the `startTLS()` method later in the connection. Is this a concern? I have not used asyncio so would probably not be able to implement it fully, but I could at last add the `startTLS()` method and have it produce the same error as before this PR.

meejah:
Regarding asyncio, yes we should provide a ``startTLS`` in the asyncio version that raises some error (maybe just ``NotImplementedError``?)

shuckc:
I've added the failure for asyncio with wss+proxies, using the exact same exception type and text that was raised previously.

meejah:
Great, thanks for making these changes :)

shuckc:
You guys are very welcome :-) autobahn-python is a handy project and I am glad to help out.

PR #540, opened 2015-11-01
Added caller_role and caller_id to message.Invocation

Kostronor:
Important addition to support a pull request in crossbar.io


meejah:
**Note:** please keep discussion about this feature on the related crossbar PR (https://github.com/crossbario/crossbar/pull/501) so it's all in one place.

oberstet:
session ID is already forwarded, and authrole isn't in the spec right now. so until the spec is finalized, we are not going to change the implementation

Issue #541, opened 2015-11-02
Sequential IDs should begin at 1


meejah:
According to https://github.com/wamp-proto/wamp-proto/blob/c3c72c66a9fe169bff1a0a1c06b02897490c2b5c/rfc/draft-oberstet-hybi-tavendo-wamp.txt#L798 sequential, session-scoped IDs should start at 1; they currently start at 0 in Autobahn.

meejah:
Also, from a report in #autobahn, this breaks the Bonefish router.

oberstet:
fixed now

PR #542, opened 2015-11-03
WAMP Session IDs start at 1 per the RFC/spec text

meejah:
Change documentation and implementation to use 1 as the
starting point, and a unit-test for wraparound. See also #541


meejah:
@hawkowl or @oberstet?

Issue #535, opened 2015-10-22
secure websockets broken with latest twisted


sherpya:
I'm using twisted 15.4.0, in autobhan/twisted/wamp.py (setupService)

you need this code, but I don't known if it breaks old twisted versions, also it would be nice to pass the context e.g. to add certificates

```python
            if isSecure:
                from twisted.application.internet import SSLClient
                from twisted.internet import ssl
                contextFactory = ssl.ClientContextFactory()
                client = SSLClient(host, port, transport_factory, contextFactory)
            else:
                from twisted.application.internet import TCPClient
                client = TCPClient(host, port, transport_factory)

            client.setServiceParent(self)
```

sherpya:
I propose something like
```
diff --git a/autobahn/twisted/wamp.py b/autobahn/twisted/wamp.py
index 6d4b8bb..f3684ed 100644
--- a/autobahn/twisted/wamp.py
+++ b/autobahn/twisted/wamp.py
@@ -548,7 +548,7 @@ if service:
         """
         factory = WampWebSocketClientFactory
 
-        def __init__(self, url, realm, make, extra=None,
+        def __init__(self, url, realm, make, contextFactory=None, extra=None,
                      debug=False, debug_wamp=False, debug_app=False):
             """
 
@@ -573,6 +573,7 @@ if service:
             """
             self.url = url
             self.realm = realm
+            self.contextFactory = contextFactory
             self.extra = extra or dict()
             self.debug = debug
             self.debug_wamp = debug_wamp
@@ -602,12 +603,14 @@ if service:
 
             if isSecure:
                 from twisted.application.internet import SSLClient
-                clientClass = SSLClient
+                if self.contextFactory is None:
+                    from twisted.internet import ssl
+                    self.contextFactory = ssl.ClientContextFactory()
+                client = SSLClient(host, port, transport_factory, self.contextFactory)
             else:
                 from twisted.application.internet import TCPClient
-                clientClass = TCPClient
+                client = TCPClient(host, port, transport_factory)
 
-            client = clientClass(host, port, transport_factory)
             client.setServiceParent(self)
```

meejah:
We are working on a better API for ``ApplicationRunner``, including support for retrying, reconnecting and SSL options. Passing a context factory (should be ``context_factory``) is certainly a good first step though!

[...]

PR #543, opened 2015-11-04
pass contextFactory for SSLClient and reveal to Service API

meejah:
Fix for #535 with a couple pep8 variable re-names as well.


meejah:
@hawkowl opinions?

PR #549, opened 2015-11-17
implement WAMP-CBOR serialization

oberstet:


oberstet:
see: https://github.com/wamp-proto/wamp-proto/issues/151

PR #551, opened 2015-11-18
Fix onDisconnect invocation to match interface

meejah:
It seems that one of the new-API commits on master changed the invocation of ``onDisconnect``, but this doesn't match the interface and so anything that overrides ``ApplicationSession`` may hit this (e.g. many examples break).

Also, include a unit-test I wrote to ensure we're logging errors properly if an onDisconnect has an unexpected signature


oberstet:
fails on travis

meejah:
@oberstet re-opened w/ additional commit; now passes

PR #553, opened 2015-11-19
Port Rawsocket on Twisted to Python 3

hawkowl:


hawkowl:
This seems to be the underlying issue in https://github.com/crossbario/crossbar/issues/475 .

autobahn/twisted/rawsocket.py
```diff
@@ -198,20 +198,20 @@ def dataReceived(self, data):
                 if self.factory.debug:
                     self.log.info("WampRawSocketProtocol: opening handshake received - {0}".format(binascii.b2a_hex(self._handshake_bytes)))
 
-                if ord(self._handshake_bytes[0]) != 0x7f:
+                if ord(self._handshake_bytes[0:1]) != 0x7f:
                     if self.factory.debug:
```
oberstet: why is that needed for py3?

hawkowl: Slicing on bytes works differently in 2/3.

```
red ~> python
Python 2.7.10 (default, Aug 22 2015, 20:33:39)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.59.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> b"foo"[0]
'f'
>>> ^D
red ~> python3
Python 3.5.0 (default, Oct 16 2015, 21:03:27)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> b"foo"[0]
102
>>> b"foo"[0:1]
b'f'
```
Single element lookup returns an int, a x-y slice returns another bytestring.

oberstet: huh, ok. I wasn't aware of that. that's evil.

hawkowl: It makes some form of sense, if your interpretation of a bytestring is a "series of octets" rather than strings. But still, confusing :(

Issue #547, opened 2015-11-13
Wrong txaio backend is used


hlamer:
I got a very interesting bug.

The project works on Python3 and asyncio. But Twisted is also installed. txaio chose twisted as default backend.
I use `autobahn.asyncio.websocket`. It works at most fine. Only autoping functionality got broken because autoping is scheduled with twisted version of call_later. The bug was discovered a few month after it has appeared.

I fixed this situation with manual `txaio.use_asyncio()` call, however would be better if `autobahn.asyncio.websocket` checks if txaio `using_asyncio()` and either configures txaio or fails loudly.

meejah:
Are you using ``ApplicationRunner`` to start things?

meejah:
(Either way, though, there should probably be some kind of check in ``autobahn.asyncio.*`` somewhere)

hlamer:
No, I don't use ApplicationRunner

meejah:
Are you running your sessions as either a router or container worker in crossbar.io?

hlamer:
No, I don't

meejah:
Hmm, okay.

There are already explicit calls to ``.use_asyncio()`` in the ``autobahn.asyncio.wamp`` package -- but if anything else imports something out of ``autobahn.twisted.wamp`` then that will call ``.use_twisted()`` and "win". That is: you can't mix twisted + asyncio in one process.

So, I think the best thing to do is make either of the ``.use_*()`` calls "one-time" in txaio and have it throw an exception if you try to use something different (e.g. calling ``.use_asyncio()`` twice is okay, but calling ``.use_twisted()`` after gives you an exception).

meejah:
I'm still curious how you're hitting this bug, though; there may be some flow of imports in Autobahn that cases a ``.twisted.*`` thing to get imported ...?

Perhaps after I fix the txaio bug it will become apparent what's happening...

hlamer:
Twisted wasn't imported by me. It just was installed for Python 3.
Look at [this](https://github.com/crossbario/txaio/blob/master/txaio/__init__.py#L120) code. If I understood correctly, it just uses twisted by default.

meejah:
Yes, it does use Twisted by default if it's available, but: https://github.com/crossbario/autobahn-python/blob/master/autobahn/asyncio/wamp.py#L45

hlamer:
Hmm. I have a hypothesis.

[...]

PR #557, opened 2015-11-27
Make all twisted-using import paths call txaio.use_twisted

meejah:
Could potentially be the actual problem from #547

### DIFF
--- a/autobahn/twisted/choosereactor.py
+++ b/autobahn/twisted/choosereactor.py
@@ -28,7 +28,8 @@ from __future__ import absolute_import
 
 import traceback
 
-from txaio import make_logger
+import txaio
+txaio.use_twisted()
 
 __all__ = (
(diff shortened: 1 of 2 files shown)


oberstet:
shouldn't the asyncio import paths then also explicitly select txaio-asyncio?

meejah:
They should, yes. No unit-tests revealed any missing cases, but I didn't really audit the asyncio code -- trying to run-all-examples.sh just now, there's at least one asyncio failure; I'll open a new PR.

oberstet:
Ah, great.

I haven't looked closely .. but I had expected some lines in "autobahn/asyncio/websocket.py" for explicit selection. Unrelated: merged your latest PR on txaio, but the build is red: "fixture 'framework' not found" ..

meejah:
See #560 (and 44 in txaio)

PR #558, opened 2015-12-01
WAMP - Twisted: Support proxy configuration in ApplicationRunner

william-p:
Tested with `Squid 3.3.8` and `wss` endpoint.
Warning: by default Squid deny `CONNECT` on !SSL_ports.


oberstet:
Thanks for contributing! (Note: we might need to change the interface in the upcoming future - likely, we'll have a better and more flexible way than ApplicationRunner)

PR #560, opened 2015-12-02
explicit .use_asyncio() in right spot

meejah:


meejah:
Weird, works locally...

meejah:
The "base-class" unit-tests needed to select a txaio subsystem (but, this was a bit order-dependent on which tests ran first). Works locally and on travis. I also re-ran all the examples in python3.

PR #568, opened 2015-12-09
fix build

meejah:
remove weird character from comment


meejah:
Still the python3 + 'twisted.cred.strcred' failures on Twisted trunk, but this fixes all the py2 builders.

Issue #573, opened 2015-12-28
Crash on Python3 because of unicode type check


hlamer:
In 0.11.0
```
  File "/home/a/kpenv/lib/python3.4/site-packages/autobahn/websocket/types.py", line 245, in __init__
    assert(reason is None or type(reason) == unicode)
NameError: name 'unicode' is not defined
```

PR #576, opened 2016-01-14
ConnectionDeny: improved Python 3 support.

mrwhhc:
Using Autobahn's Twisted server protocols on Python 3 can result in unreadable error messages when an incoming connection is denied.  That's because `ConnectionDeny` asserts that it's reason is of type `unicode`, which doesn't exist on Python 3.

A solution is to a) change the assertion to check against `six.text_type` and b) ensure that all relevant call points provide a unicode string.

This pull request does just that, and also fixes a similar issue in `ConnectionAccept.`

Fuzz testing passed and unittest2 passed for Python 2 but I'm unclear how best to write a test for Python 3, as it appears the unit tests for Python 3 require asyncio.

### DIFF
--- a/autobahn/twisted/test/test_protocol.py
+++ b/autobahn/twisted/test/test_protocol.py
@@ -1,4 +1,4 @@
-###############################################################################
+##############################################################################
 #
 # The MIT License (MIT)
 #
@@ -29,11 +29,16 @@ from __future__ import absolute_import, print_function
 import unittest2 as unittest
 from mock import Mock
 
+from autobahn.websocket.types import ConnectionRequest, ConnectionDeny
+
 from autobahn.twisted.websocket import WebSocketServerFactory
 from autobahn.twisted.websocket import WebSocketServerProtocol
+from autobahn.twisted.websocket import WrappingWebSocketServerFactory
 from twisted.python.failure import Failure
 from twisted.internet.error import ConnectionDone, ConnectionAborted, \
     ConnectionLost
+from twisted.internet.protocol import Protocol, Factory
+from twisted.internet.address import IPv4Address
 from autobahn.test import FakeTransport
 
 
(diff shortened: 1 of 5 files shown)


meejah:
This looks great, thanks! :)

As to testing, we use "Tox" to run multiple test environments; see ``tox.ini`` at the top level. Running ``tox -l`` will list the available environments, of which one is ``py34-tw155`` against Twisted.

For this PR, you could add tests under ``autobahn/twisted/test/*`` .. perhaps a new file, or maybe a new class in ``test_protocol.py``?

mrwhhc:
Thanks for the quick reply!  I'll add a test to `autobahn.twisted.test.test_protocol` and verify with tox.  New commit forthcoming!

mrwhhc:
This seems to close #573

oberstet:
this is fixed

mrwhhc:
This is not fixed. 

Without fixing those call points, the problem has been changed from a `NameError` to an `AssertionError`. That's because , a unicode string must be passed into `ConnectionDeny` everywhere it's instantiated.  Here's what happens in Python 2 without addressing that:

````
File "/home/mrw/src/python/autobahn-python/autobahn/twisted/websocket.py", line 299, in onConnect
    raise ConnectionDeny(ConnectionDeny.NOT_ACCEPTABLE, "this server only speaks %s WebSocket subprotocols" % self.factory._subprotocols)
  File "/home/mrw/src/python/autobahn-python/autobahn/websocket/types.py", line 245, in __init__
    assert(reason is None or type(reason) == six.text_type)
exceptions.AssertionError: 
````

This is why the pull request changed all the places where `ConnectionDeny` was instantiated:

https://github.com/mrwhhc/autobahn-python/blob/7e75e4571fa04996c1a4e67e58ebf389db2cc89a/autobahn/wamp/websocket.py#L186
https://github.com/mrwhhc/autobahn-python/blob/7e75e4571fa04996c1a4e67e58ebf389db2cc89a/autobahn/twisted/websocket.py#L299

And included tests that covered those lines.

oberstet:
Ah, right: https://github.com/crossbario/autobahn-python/commit/7f1935d0b22c465f6684ce5b42388010324d5c4e

PR #582, opened 2016-01-26
added mode for balanced shared registrations

markope:
added the balance mode to the message register. This pull request goes along with the pull request https://github.com/crossbario/crossbar/pull/613 on crossbar.


oberstet:
The mode is misnamed: `roundrobin` and `random` are also balancing modes. The mode should be named `leastbusy` https://github.com/crossbario/crossbar/issues/606

PR #592, opened 2016-02-04
Strengthed uri Pattern tests and fixed parsing edge case

Paranaix:
Why can named components without type not start with a number? The old regex implicated that this was done with intention? However IMO this is totally unintuitive (I wouldn't be doing this PR if it were). Some feedback what the original thougth process was behind this would be highly appreciated.


oberstet:
@Paranaix yes, I agree. URI components with digits only are useful. E.g. like in `com.product.<product_id>. ...` with `product_id` an integer. In fact, that's what the WAMP spec says (even for strict URIs). So it was in fact a bug! Thanks!

https://github.com/wamp-proto/wamp-proto/blob/master/rfc/text/basic/bp_identifiers.md#strict-uris

PR #593, opened 2016-02-05
Observable API and listeners fixups

meejah:
This makes the ``ObservableMixin`` less fragile for use as a mix-in by eliminating its parameter-taking ``__init__`` method, and some cleanups and docstrings. It also adds an (optional!) list of "valid events" to make users of ``.on()`` (or even internal uses of ``.fire()``) complain if you mis-type an event name.

The other commit tackles the LTS listener API by fixing how they're fired/invoked. Doing the invokes in e.g. ``onJoin`` won't work with most of the code out there (including all the examples) because the API contract never specified to call ``super()`` (and basically no code does this). This means that the 'join' event is basically never fired properly.

So, as per LTS API discussions there are 5 events. Note that I made one subtle change that makes a big difference (to when 'join' fires):

 - join: right before we call onJoin
 - leave: right *after* we call onLeave
 - ready: after all 'join' listeners have completed, and after onJoin has completed
 - connect: right before onConnect
 - disconnect: right after onDisconnect

Above "completed" means the method has run, and any future/deferred returned has also completed. The reason I made 'join' fire *before* onJoin has run and completed is, consider code like:

```
class Foo(ApplicationSession):
    def onJoin(self, details):
        yield self.call(...)
        # etc
        yield self.leave()
```

That is, things that use ``onJoin`` as essentially a "main" function -- do all their work, and then disconnect the session. In the above, if we wait for ``onJoin`` to complete before we fire the 'join' listeners, then we very counter-intuitively would fire 'leave' *before* firing 'join'.

So, now 'join' fires as soon as we get the Hello message, and then the previously-discussed 'ready' fires *after* onJoin has completed (and all 'join' listeners have completed). Also, this is implemented such that it doesn't matter which of the over-ridable methods in ApplicationSession you choose to override, and doesn't care if you call ``super()`` or not. Also, any errors from listeners are logged (but ignored).

### DIFF
--- a/autobahn/util.py
+++ b/autobahn/util.py
@@ -598,20 +598,83 @@ def wildcards2patterns(wildcards):
 
 
 class ObservableMixin(object):
+    """
+    Internal utility for enabling event-listeners on particular objects
+    """
 
-    def __init__(self, parent=None):
-        self._parent = parent
-        self._listeners = {}
+    # A "helper" style composable class (as opposed to a mix-in) might
+    # be a lot easier to deal with here.  Having an __init__ method
+    # with a "mix in" style class can be fragile and error-prone,
+    # especially if it takes arguments. Since we don't use the
+    # "parent" beavior anywhere, I didn't add a .set_parent() (yet?)
+
+    # these are class-level globals; individual instances are
+    # initialized as-needed (e.g. the first .on() call adds a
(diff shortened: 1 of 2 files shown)


oberstet:
I like that! It brings a lot of non-invasive flexibility ..

PR #597, opened 2016-02-08
Fix log method in transport serialization error handling.

hozn:
The error handler on dispatch is swallowing exceptions and never raising SerializationError due to "AttributeError: '_TxaioLogWrapper' object has no attribute 'failure'" being fired in the exception handling routine.


meejah:
thanks!

PR #600, opened 2016-02-13
fix error message

AndrewSamokhvalov:


oberstet:
No, we use structured logging, which doesn't preformat log messages

PR #615, opened 2016-03-18
Marshal fixes

meejah:
I discovered this from some "exclude_me=False" test-cases; not sure if they're all strictly necessary but it matches the way the `messages.py` stuff works (and at least when you pass `exclude_me=False` it just doesn't work right now because that never gets serialized across).


oberstet:
Yeah, that's the idea: if not None, then marshal, even when the value is default. So yes, this patch making that systematic behavior.

PR #619, opened 2016-03-30
Move parseWsUrl and createWsUrl to a new package

meejah:
The 'protocol' package imports txaio, and so unneccesarily requires anyone who uses "parseWsUrl" (e.g. a test) to initialize txaio first (and thereby choose twisted vs asyncio).

There may be other candidates in 'protocol' to move. Note that of course this breaks any existing imports of these functions...but not part of "public API" :/

### DIFF
--- a/autobahn/websocket/util.py
+++ b/autobahn/websocket/util.py
@@ -51,11 +51,11 @@ urlparse.uses_query.extend(wsschemes)
 urlparse.uses_fragment.extend(wsschemes)
 
 __all__ = (
-    "create_ws_url",
-    "parse_ws_url",
+    "create_url",
+    "parse_url",
 )
 
-def create_ws_url(hostname, port=None, isSecure=False, path=None, params=None):
+def create_url(hostname, port=None, isSecure=False, path=None, params=None):
     """
     Create a WebSocket URL from components.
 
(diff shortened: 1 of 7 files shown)


oberstet:
I think `autobahn.websocket.util` is a good place for these functions! They are networking library independent, and about WebSocket.

Also: as these 2 functions are supposed to be part of the informal semi-public API, and moving them might break user code, we might take the chance and do the PEP8 compliant renaming for these functions as well.

meejah:
@oberstet good suggestion, which also revealed several places I didn't fix the imports :+1:

autobahn/websocket/util.py
```diff
@@ -0,0 +1,150 @@
...
+urlparse.uses_params.extend(wsschemes)
+urlparse.uses_query.extend(wsschemes)
+urlparse.uses_fragment.extend(wsschemes)
+
+__all__ = (
+    "create_url",
+    "parse_url",
+)
+
+def create_url(hostname, port=None, isSecure=False, path=None, params=None):
```
oberstet: PEP8: `is_secure` ?

examples/twisted/websocket/echo_multicore/server.py
```diff
@@ -63,7 +63,7 @@
 from twisted.web.server import Site
 from twisted.web.static import File
 
-from autobahn.websocket.protocol import parseWsUrl
+from autobahn.websocket.util import parse_ws_url
```
oberstet: `parse_url`?

oberstet:
there are a few places with `parse_ws_url` left in unit tests as far as I see ..

meejah:
Hmm, yes. "git grep" vs. "grep -R" perhaps; should be fixed now...also rebased over the ``examples/`` now on master.

oberstet:
only pep8 left: "autobahn/websocket/util.py:58:1: E302 expected 2 blank lines, found 1"

PR #630, opened 2016-04-09
skip autodoc of any member NOT having the magic @public 'decorator' i…

oberstet:
…n the docstr


oberstet:
stashing away this idea for a sec .. this is unfinished

doc/conf.py
```diff
@@ -185,10 +185,24 @@ def use_aio():
...
+# http://stackoverflow.com/a/21449475/884770
+# http://www.sphinx-doc.org/en/stable/ext/autodoc.html#event-autodoc-skip-member
+
+# app, what, name, obj, skip, options:
+# <sphinx.application.Sphinx object at 0x2b0192ab2f90> class __module__ autobahn.asyncio.websocket True {'show-inheritance': True, 'members': <object object at 0x2b018c791710>, 'undoc-members': True}
+#
+def autodoc_skip_member(app, what, name, obj, skip, options):
+    if hasattr(obj, '__doc__') and obj.__doc__ is not None and obj.__doc__ != '':
+        if '@public' not in obj.__doc__:
+            return True
```
meejah: Interesting!
Also, since we get the actual ``obj`` in the hook, it *could* be a real decorator (e.g. mark the function/class) or a global list somewhere like ``public_api = [ApplicationRunner, ...]`` perhaps. Neat idea, though (listing public API explicitly in code + docs).

oberstet: Ah, yes. A real decorator! Seems nicer. I will try that. (I'd prefer that vs a global list, because it's "in place", less likely to be overlooked, etc ..).

> What's also possible with this approach is: we could generate docs for ourselves in a special "AB devs doc generating mode" too in addition to the "AB user docs" which only dump the public API.

meejah:
With a "real" decorator, it could be the decorator itself which is responsible for populating a global list *or* via "marking" the function w/ an attribute. e.g.:

```
_public_api = []
def public(obj):
    _public_api.append(obj)
    return obj
```
Versus:
```
def public(obj):
    obj._public_api = True
    return obj
```

Perhaps there's a reason to prefer one over the other; in either case it's easier for the hook-method to figure out what to do (i.e. ``getattr(obj, '_public_api', False)`` versus ``obj in _global_api`` basically).

oberstet:
superseded by https://github.com/crossbario/autobahn-python/pull/794

PR #647, opened 2016-04-21
Correction of little typo, add of a video on asyncio

potens1:
The video describe very well the asyncio lib and is very easy to understand. Jesse Jiryu Davis has done more than one video on the subject (https://www.google.be/search?q=python+asyncio+%2B"jesse+jiryu+davis"&tbm=vid), but they are almost interchangeable, I think this one is good enough


oberstet:
thanks!

Issue #642, opened 2016-04-21
Use batching/chunking timers


oberstet:
There are multiple places within Autobahn (at the WebSocket and at the WAMP level) where we use timers sticked to protocol instances.

When running massive numbers of connections, this can lead to significant CPU load.

Since these timers don't need to be precise, and hence we can conserve CPU by batching timeouts into buckets (and process timeout callbacks within a bucket in chunks).

The underlying functionality will be in `txaio`: https://github.com/crossbario/txaio/issues/66

Once that is done, we should use them here.

oberstet:
See:

* https://github.com/crossbario/crossbar/issues/722
* https://github.com/crossbario/txaio/issues/66

meejah:
With the changes from https://github.com/crossbario/txaio/pull/67 I hacked together some quick numbers after replacing all the ``call_later``s in websocket stuff to batched timeouts.

This is just a local test on an 8-core machine, ramping up to 20k do-nothing websocket connections (in one client process) with 10s auto-pings and 60s ping-timeouts enabled and then idling for a while. I'm showing the CPU usage of the server process as grok'd by ``collectl`` every second (there's almost no green impulses in the latter part of the graph because the CPU usage is very low).

Conclusion: well worth the batching optimization. I will do a few more tests tomorrow playing with some of the batching parameters (this was 5s bucket size, chunked at 1000 per yield). The high CPU usage on the left is during the ramp-up phase...

![batching](https://cloud.githubusercontent.com/assets/145599/14733893/a0431c42-0823-11e6-86dd-34a8faa5b526.png)

oberstet:
@meejah 

> Conclusion: well worth the batching optimization.

Coool!!

We should really do that in AB.

We might also use this to implement write coalescing: https://github.com/crossbario/autobahn-python/compare/master...oberstet:write_coalescing

oberstet:
rgd the value of write coalescing, this is two-fold:

* it reduces the number of syscalls and CPU load (tradeoffs is increased latency vs throughput)
* it reduces the wire level traffic when run over TLS - we should make it such that writes to the underlying Twisted stream protocol only does 1 write

Rgd the latter: http://tavendo.com/blog/post/dissecting-websocket-overhead/

DZabavchik:
Night and day.  Looks more like 20x+ improvement, way better than what I got.

meejah:
[...]

PR #654, opened 2016-04-28
batched timers

meejah:
See #642 

Note the txaio version-bump; I put in a PR there too which bumps the version as I don't believe the millisecond stuff nor the "spread out notifications" things are in 2.4.x
https://github.com/crossbario/txaio/pull/73

### DIFF
--- a/autobahn/websocket/protocol.py
+++ b/autobahn/websocket/protocol.py
@@ -987,7 +987,6 @@ class WebSocketProtocol(object):
                 self.onOpenHandshakeTimeout,
             )
 
-
     def _connectionLost(self, reason):
         """
         This is called by network framework when a transport connection was
--- a/autobahn/websocket/protocol.py
+++ b/autobahn/websocket/protocol.py
@@ -469,6 +469,10 @@ class WebSocketProtocol(object):
     """
 
     log = txaio.make_logger()
+    ping_timer = txaio.make_batched_timer(
+        bucket_seconds=0.200,
+        chunk_size=100,
+    )
 
     def __init__(self):
         #: a Future/Deferred that fires when we hit STATE_CLOSED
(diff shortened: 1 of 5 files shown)


oberstet:
@meejah I've pushed the `txaio` 2.5.0 release https://pypi.python.org/pypi/txaio/2.5.0 but there a failing tests still. Interestingly, the asyncio platform tests run green, while the Twisted ones do not ..

meejah:
Darn, this was partially a txaio bug ... https://github.com/crossbario/txaio/pull/74

oberstet:
@meejah cool, everything well. np. I published a new 2.5.1 `txaio` and now everything passes;)

autobahn/websocket/protocol.py
```diff
@@ -657,11 +665,10 @@ def onCloseFrame(self, code, reasonRaw):
                 # When we are a client, the server should drop the TCP
                 # If that doesn't happen, we do. And that will set wasClean = False.
                 if self.serverConnectionDropTimeout > 0:
-                    call = txaio.call_later(
+                    self.serverConnectionDropTimeoutCall = txaio.call_later(
```
oberstet: Drop connection timer is regular, not BatchTimer? Any specific reason?

meejah: Hmm, I think i was just simplifying/white-spacing and it snuck it ;)

oberstet: Ah;) Maybe we should make it a BatchTimer too. Also, as this will this be public API, I think it's worth to think a little about naming and such. So I'd like to propose:

* `ping_send_timer`: This timer is used for automatic, periodic sending of WebSocket pings.
* `pong_timeout_timer`: This timer is used for timeout receiving pongs to automatic pings.
* `opening_handshake_timeout_timer`: This timer is used for timout on finishing the initial WebSocket opening handshake.
* `drop_connection_timeout_timer`: This timer is used for timeout the peer automatically dropping the connection as it should happen after the WebSocket closing handshake has completed.

meejah:
Thinking about separate timers: I wonder if it partially defeats the purpose (especially if they're all using the same batching parameters)? That is, what not have just *one* timer-creator thing for everything in ``websocket/protocol.py`` to use?

I guess too I didn't think about this being public or not. But, should it be? Obviously the timeouts themselves should be, but the only reason anything else might want access to the batching objects is to change the batching parameters, right? (Or, is that what you're getting at: if we have different timer objects then you can override the batching parameters for the different timers if need be...?)

oberstet:
@meejah sorry, was interrupted again. Rethinking about this, I guess you are right. We only need 1 batch timer. Only the `call_later` calls themselfes need different parameters for the different thing (like the list of 5 different things above) - but these parameters (like ping period etc) are actually unrelated to the bucket size / chunk size.

So screw what I said above, and let's make it a single, private class batch timer: `WebSocketProtocol._batch_timer`

PR #667, opened 2016-05-01
Removed `debug` kwarg from examples

shabanski:
Some examples contain a nonexistent `debug` kwarg in `__init__`.


oberstet:
@shabanski cool! thanks for cleaning up;)

PR #675, opened 2016-05-05
Add showServerVersion controlling full/partial version in headers

meejah:
This defaults to False, and so will only show that we're using "Autobahn" and not the precise version.

(See https://github.com/crossbario/crossbar/issues/770 for discussion)

### DIFF
--- a/autobahn/websocket/protocol.py
+++ b/autobahn/websocket/protocol.py
@@ -445,6 +445,7 @@ class WebSocketProtocol(object):
     """
 
     CONFIG_ATTRS_SERVER = ['versions',
+                           'showServerVersion',
                            'webStatus',
                            'requireMaskedClientFrames',
                            'maskServerFrames',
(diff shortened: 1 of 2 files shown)


autobahn/websocket/protocol.py
```diff
@@ -2758,7 +2759,10 @@ def succeedHandshake(self, res):
         if self.factory.server is not None and self.factory.server != "":
             response += "Server: %s\x0d\x0a" % self.factory.server
 
-        response += "X-Powered-By: AutobahnPython/{0}\x0d\x0a".format(__version__)
+        if self.showServerVersion:
+            response += "X-Powered-By: AutobahnPython/{0}\x0d\x0a".format(__version__)
+        else:
+            response += "X-Powered-By: AutobahnPython\x0d\x0a".format(__version__)
```
oberstet: Should probably not have a __version__ formatting arg - but on the other hand, maybe do away with "X-Powered-By" header altogether. One server header is enough.

meejah:
See instead: https://github.com/crossbario/autobahn-python/pull/677

PR #685, opened 2016-05-30
Fixing inconsitency between PublishOptions and Publish message 

izderadicka:
- while PublishOptions allow single authid, role in eligible/excluded Publish message requires lists

### DIFF
--- a/autobahn/wamp/types.py
+++ b/autobahn/wamp/types.py
@@ -511,19 +511,19 @@ class PublishOptions(object):
             options[u'exclude'] = self.exclude if type(self.exclude) == list else [self.exclude]
 
         if self.exclude_authid is not None:
-            options[u'exclude_authid'] = self.exclude_authid if type(self.exclude_authid) == list else self.exclude_authid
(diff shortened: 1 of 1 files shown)


oberstet:
No, exclude_authid / exclude_authrole etc does not require a list

izderadicka:
This is in source for autobahn.wamp.message  (Publish message):

line 1176 :  assert(exclude_authid is None or type(exclude_authid) == list)

and this is in autobahn.wamp.types,  (Publish Options):

line 483:  assert(exclude_authid is None or type(exclude_authid) == 
six.text_type or (type(exclude_authid) == list and all(type(x) == 
six.text_type for x in exclude_authid)))

line 513:        if self.exclude_authid is not None:
             options[u'exclude_authid'] = self.exclude_authid if 
type(self.exclude_authid) == list else self.exclude_authid


If I use PublishOptions(exclude_authid='something'), options object is 
create but later got AssertionError
when publish is called ( and publish message is being created).

So there is problem there.

I.

meejah:
Yeah, it looks like some of the code is insisting on a list, and some will accept list or string...

PR #688, opened 2016-06-06
Pass a "headers" parameter through to the WampWebSocketClientFactory

pramukta:
I think this allows extra request headers to be sent along with the HTTP UPGRADE request (as supported by the WampWebSocketClientFactory) while using the ApplicationRunner

Would a change like this be reasonable to add?  It's possible that discussion on #674 would imply that this kind of thing could go into a more comprehensive refactor instead, but figured it was worth suggesting.

The use case here is that I want to add an HTTP authorization header with an JWT Bearer token to pass through a separate authenticating proxy. I'm sure there are other uses for being able to specify HTTP headers as well.  Thanks.

### DIFF
--- a/autobahn/twisted/wamp.py
+++ b/autobahn/twisted/wamp.py
@@ -91,7 +91,7 @@ class ApplicationRunner(object):
 
     log = txaio.make_logger()
 
-    def __init__(self, url, realm, extra=None, serializers=None, ssl=None, proxy=None):
+    def __init__(self, url, realm, extra=None, serializers=None, ssl=None, proxy=None, headers=None):
         """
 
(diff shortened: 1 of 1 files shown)


oberstet:
> It's possible that discussion on #674 would imply that this kind of thing could go into a more comprehensive refactor instead

Yes, exactly. `ApplicationRunner` currently doesn't allow to use RawSocket instead of WebSocket transport etc. The API isn't general enough.

But since we already have `proxy` in it (which also is WebSocket specific), I am merging this - in the longer run, we will have something better than `ApplicationRunner`, but for now, haveing `headers` is "ok".

Issue #691, opened 2016-06-14
Insecure allowedOrigins validation


vetsin:
Autobahn|Python incorrectly checks the Origin header when the 'allowedOrigins' value is set.

The following will set
```
class OriginCheckServerFactory(WebSocketServerFactory):
    protocol = ...arbitrary entry here...

    def __init__(self, url):
        WebSocketServerFactory.__init__(self, url)
        self.setProtocolOptions(allowedOrigins=[u"127.0.0.1",u"*.example.com"])
```

Then the following connection request will result in a valid 101 Protocol Switch Response:

```
GET /ws HTTP/1.1
Host: www.example.com
Sec-WebSocket-Version: 13
Origin: http://www.example.com.malicious.com
Sec-WebSocket-Extensions: permessage-deflate
Sec-WebSocket-Key: tXAxWFUqnhi86Ajj7dRY5g==
Connection: keep-alive, Upgrade
Upgrade: websocket
```

This is due to the [wildcard2patterns](https://github.com/crossbario/autobahn-python/blob/master/autobahn/util.py#L632) functions, which turns `u"*.example.com"` into `r".*\.example\.com"`. This regex pattern is then [matched against the complete incoming origin value](https://github.com/crossbario/autobahn-python/blob/1c86fefa733901247d6439af346d77cbb58a9bf4/autobahn/websocket/protocol.py#L2588). The websocket_origin value should be first parsed as a URI object and the authority section extracted, and the regex matched against authority ensuring that ^ and $ are used. E.g. `r"^.*\.example\.com$"`

eastein:
The scheme must be validated as well. Merely validating the host would be insufficient as MITM + failure to validate scheme = vuln, if the end user intends to use HTTPS only.

meejah:
I have written a unit-test simulating vestin's example above, and can confirm that the matching does indeed fail (i.e. the above will *not* call `failHandshake()` as it should).

meejah:
Thanks to ``Toba``, ``vetsin`` and ``jchampion`` for discussion in ``#autobahn``

meejah:
Okay, I think I incorporated the feedback in the PR.

[...]

PR #693, opened 2016-06-15
Issue691 allowed origins

meejah:
Fixes for #691, allowed-origin checks.

(In-progress).

### DIFF
--- a/autobahn/twisted/test/test_protocol.py
+++ b/autobahn/twisted/test/test_protocol.py
@@ -29,11 +29,13 @@ from __future__ import absolute_import, print_function
 import unittest2 as unittest
 from mock import Mock
 
+from autobahn.util import wildcards2patterns
 from autobahn.twisted.websocket import WebSocketServerFactory
 from autobahn.twisted.websocket import WebSocketServerProtocol
 from twisted.python.failure import Failure
 from twisted.internet.error import ConnectionDone, ConnectionAborted, \
     ConnectionLost
+from twisted.test.proto_helpers import StringTransport
 from autobahn.test import FakeTransport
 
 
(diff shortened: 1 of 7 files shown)


autobahn/websocket/protocol.py
```diff
@@ -2577,19 +2609,33 @@ def processHandshake(self):
...
+                # as "origin" to onConnect, so conceivably
+                # implementations of that could mistake "" as the same
+                # origin as another WebSocket that also just lacks on
+                # Origin header...?
+
+            # XXX to pednatically follow RFC6454, we should probably
+            # assign websocket_origin as the triple (scheme, host,
+            # port) but that would change what's in onConnect's
+            # information (for .origin). Worth breaking the API for?
+            if have_origin:
```
meejah: @oberstet should we do either of the above XXX things? Or should I just change the comments?

eastein:
IMO, treating the Origin as a triple is pretty much required by the RFC and to make things safe.

meejah:
For the "built in" checks, I believe I am treating it as a triple that must match -- my question is more regarding the API presented via ``ConnectionRequest`` to ``onConnect``. For example, we could maintain compatibility by leaving ``.origin`` as a string with the host, and add ``.scheme`` and ``.port`` or similar -- *or* force everyone to (possibly) change existing code and make ``.origin`` a ``(scheme, host, port)`` triple.

Certainly the latter seems "more correct", but might not be worth breaking code. I guess a "third way" would be to add a new ``.websocket_origin`` as a triple, deprecate ``.origin`` (removing ``.origin`` entirely at some point).

autobahn/twisted/test/test_protocol.py
```diff
@@ -120,3 +122,111 @@ def test_handshake_fails(self):
...
+        # we're just testing the low-level function here...
+        from autobahn.websocket.protocol import _is_same_origin
+        policy = wildcards2patterns(['example.com'])
+
+        self.assertFalse(_is_same_origin('http://example.com:8080/', 80, policy))
+        self.assertFalse(_is_same_origin('http://www.example.com/', 80, policy))
+        self.assertFalse(_is_same_origin('https://example.com:80/', 443, policy))
+        self.assertFalse(_is_same_origin('https://example.com/', 80, policy))
+        self.assertFalse(_is_same_origin('http://example.org/', 80, policy))
+        self.assertFalse(_is_same_origin('http://ietf.org/', 80, policy))
```
vetsin: Additional valid URIs, though i'm unsure if these would be possible to spoof into the Origin given any browser:

* `http://example.com./`
* `//example.com/`
* `http://@example.com`

autobahn/twisted/test/test_protocol.py
```diff
@@ -120,3 +122,111 @@ def test_handshake_fails(self):
...
+        """
+        All the example origins from RFC6454 (3.2.1)
+        """
+        # we're just testing the low-level function here...
+        from autobahn.websocket.protocol import _is_same_origin
+        policy = wildcards2patterns(['example.com'])
+
+        self.assertTrue(_is_same_origin('http://example.com/', 80, policy))
+        self.assertTrue(_is_same_origin('http://example.com:80/', 80, policy))
+        self.assertTrue(_is_same_origin('http://example.com/path/file', 80, policy))
```
vetsin: * `http://example.com/;semi=true`

autobahn/websocket/protocol.py
```diff
@@ -74,6 +75,77 @@
...
+        return 'null'
+
+    (scheme, netloc, path, _, _, _) = urllib.parse.urlparse(url)
+    host = netloc
+    if ':' in netloc:
+        host, port = netloc.split(':')
+        port = int(port)
+    else:
+        try:
+            port = {'https': 443, 'http': 80, 'file': 80}[scheme]
```
eastein: does it really make sense to treat file:// as port 80? Why do that?

autobahn/websocket/protocol.py
```diff
@@ -74,6 +75,77 @@
...
+    if websocket_origin == 'null':
+        # nothing is the same as the null origin
+        return False
+
+    if not isinstance(websocket_origin, tuple) or not len(websocket_origin) == 3:
+        raise ValueError("'websocket_origin' must be a 3-tuple")
+
+    (origin_scheme, origin_host, origin_port) = websocket_origin
+
+    # XXX maybe better to explicitly pass in desired scheme, instead
```
eastein: yeah that would be better..

meejah:
@oberstet what do you think of 671266f (comes with example code)? This allows the existing API to function while also allowing one to be explicit about ports. Perhaps I should also remove the explicit "scheme" check (in `_is_same_origin()`), too? (And rely on the reg-ex?). Currently, this will *not* allow you to accept e.g. an http origin at an https websocket endpoint (or vice-versa) -- but that probably shouldn't be allowed anyway.

oberstet:
@meejah The pragmatic approach looks good to me. Filtering on origin in a way is snake oil anyway: a non-browser client can fake it. Could you squash the PR and merge?

PR #694, opened 2016-06-25
Rewrite of asyncio wamp to let user manage the loop.

potens1:
This modification does not change anything for users using the
ApplicationRunner.run method but let users have more control on
the loop by using ApplicationRunner.create_endpoint an managing
loop as usual with asyncio.


potens1:
I'll do a cleaner pull request by creating a branch

PR #698, opened 2016-07-02
Rewrite of asyncio wamp to let user manage loop.

potens1:
This modification does not change anything for users using the
ApplicationRunner.run method but let users have more control on
the loop by using ApplicationRunner.create_endpoint an managing
loop as usual with asyncio.

### DIFF
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -108,27 +108,19 @@ class ApplicationRunner(object):
         self.serializers = serializers
         self.ssl = ssl
 
-    def run(self, make):
+    def ssl_check(self, isSecure=None):
         """
-        Run the application component.
+        Check the ssl parameters to avoid conflicting values and return it.
 
-        :param make: A factory that produces instances of :class:`autobahn.asyncio.wamp.ApplicationSession`
-           when called with an instance of :class:`autobahn.wamp.types.ComponentConfig`.
-        :type make: callable
+        :param isSecure: value extracted from url.
+
+        :returns: :class:`ssl.SSLContext` or bool -- An (optional) SSL context 
+           instance or a bool. See the documentation for the 
+           `loop.create_connection` asyncio method, to which this value is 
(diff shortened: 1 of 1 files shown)


FirefighterBlu3:
@potens1 in case the WAMP code is running secondary to the main process and is executed in a separate thread, it might be wise to use asyncio.new_event_loop() instead of getting the default loop. perhaps we ought to consider being passed a loop as well, it already being established for us such as in the case of loop.run_in_executor(...)

potens1:
@FirefighterBlu3 Maybe I miss the point but I'm not sure at all about that. I think a library's behavior must be the most predictable for a new user and, right now, I'm not aware of an asyncio library that start a new loop even if one is already started. I'm almost convinced this will break user expectation since, i.e. if the user use asyncio.queue, it will just block when he will try to read it since the loop used to read it and the one use to generate the "something to read" event are not the same.
I agree with you about the fact the loop should be an argument passed to the ApplicationSession but I've done this PR with changing the least possible things in mind.
What is the rational behind having a second event loop ?

FirefighterBlu3:
choosing a new loop won't break anything nor be visible to the default user. attempting to use the default loop will break for anyone who runs this in a different thread. it's easily tested. use loop = get_event_loop() and loop.run_in_executor() then attempt to use get_event_loop() inside that thread and ascyncio will complain and halt your task.

potens1:
Ok, I see you point here. the thing is I made this to mimic the exact same behavior of the original one. Your proposition will change that (and I guess for something better).
What you propose is still possible using ApplicationRunner.create_endpoint, and in fact, it is the whole point of the PR, let you manage the loop.
Now the thing is, people using ApplicationRunner.run don't mind about the loop used since, I guess, they are using "self contained" Application, so, it makes a lot of sense. Now, I would like to have the opinion of one of the owners before change commit.

oberstet:
The probem I see is that it introduces new API, and as above discussion shows, this isn't easy;)

PR #703, opened 2016-07-07
[proof-of-concept] Fix #701.

### LINKED ISSUE #701: Is autobahn compatible with twisted 16.3.0?
Since twisted 16.3.0, buildbot's websocket ceased to function.

after bisect, it looks like related to commit 2d39ccad9d8a9b86229dd7fb1ce47ca841b730be from twisted
which is related to this comment in release note:

Deprecations and Removals
-------------------------
 - twisted.web would previously dispatch pipelined requests
   simultaneously and queue the responses. This behaviour did not
   enforce any of the guarantees required by RFC 7230 or make it
   possible for users to enforce those requirements. For this reason,
   the parallel dispatch of requests has been removed. Pipelined
   requests are now processed serially. (#8320)

Are you testing crossbar.io with latest twisted? do you have this issue, or is this related to the way buildbot integrates autobahn?

tomprince:


meejah:
This does fix the issue, and is pretty simple. I'll merge it so at least master works out-of-the-box on 16.3.0

Issue #641, opened 2016-04-21
Reverse protocol handling for WebSocketResource


warner:
I have a set of unit tests (in https://github.com/warner/magic-wormhole) that recently started failing when I combine autobahn-0.13.1 and Twisted trunk. It looks like https://github.com/twisted/twisted/commit/4f9e36d8ad21f9d73369f4d506bab30f61c2ad71 (landed 29-Mar-2016) is where the problem started (that's the first revision after Twisted-16.1.1 was branched off that causes my project's tests to fail).

The commit says:

    Merge hide-request-transport-8191-5: Hide the transport backing the HTTPChannel object from twisted.web Resource objects.

And my tests are using `WebSocketResource` to put a websocket on a specific URL. I haven't tracked down the specific failure yet, but it seems like the server is having an internal error when the client attempts to connect to it, or the server isn't accepting the connection at all.

The relevant Twisted bug is at https://twistedmatrix.com/trac/ticket/8191 .

When I get some time (maybe next week), I'll figure out how to run autobahn's test suite against various versions of Twisted and see what happens.

meejah:
Hi @warner, the test-suite can be run with ``tox`` so you could tweak one of the existing environments for the "right" Twisted version, probably.

oberstet:
this: https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/resource.py#L127

warner:
Ah, thanks. `tox -e py27-twtrunk` looks like it ought to do the right thing, but all the tests pass on my box, so I think magic-wormhole must be doing something unusual (maybe correct and uncovered by autobahn's tests, or maybe wrong).

oberstet:
> Connect the transport to our protocol. Once #3204 is fixed, there may be a cleaner way of doing this.

http://twistedmatrix.com/trac/ticket/3204

8 years old bug.

The background is: to add a Twisted Web resource that runs WebSocket, we (Autobahn) have to get at the full initial HTTP request bytes - which have been eaten away by Twisted Web before. 

I consider the code in Autobahn a "gentle hack" .. due to lack of appropriate interfaces in Twisted (and me, and that time, not having the patience to go through months of discussions to land something on Twisted;)

meejah:
Yeah, could certainly be an Autobahn bug -- not 100% line-coverage, even ...

oberstet:
Twisted has made an attribute private that was formerly public?

"""
[...]

PR #704, opened 2016-07-08
tests, guards for #701

meejah:
Add some simple tests for resource, and check we have an IPushProducer (for #701)


meejah:
After some discussion with @tomprince I will just guard the resumeProducing call (i.e. instead of throwing)

oberstet:
We should work on https://github.com/crossbario/autobahn-python/issues/641 instead of sinking more time in dealing with HTTP2 fallouts

PR #705, opened 2016-07-08
Fix WebSocketResource on 16.3.0

hawkowl:
Similar to @tomprince 's patch, but should last a bit longer (we're going to remove `request.transport` sooner or later).


hawkowl:
cc @meejah

oberstet:
So ghis only works on Twisted 16.3? Because that channel crap wasn't there before?

oberstet:
Instead of putting more hacks on top, IMO we should pin at Twisted <= 16.2 until https://github.com/crossbario/autobahn-python/issues/641 is implemented.

hawkowl:
No, `self.channel.transport` has been around for ever (the implementation is in `HTTPChannel`); but it was copied in to `self.transport`. This is soft-deprecated for now, and will go away (since it's an abstraction violation on HTTP/1 with pipelining and HTTP/2 fundamentally). https://github.com/twisted/twisted/blob/twisted-8.0.0/twisted/web/http.py#L504

oberstet:
Whatever. Really. https://github.com/crossbario/autobahn-python/issues/641

PR #706, opened 2016-07-08
Issue701 just pin

meejah:
I was premature in merging the "fix" (#703) for #701. So, this reverts that merge and then pins the Twisted dependency <= 16.2.0 until we have a better solution.

### DIFF
--- a/autobahn/twisted/resource.py
+++ b/autobahn/twisted/resource.py
@@ -158,6 +158,5 @@ class WebSocketResource(object):
                 data += "%s: %s\x0d\x0a" % (h[0], ",".join(h[1]))
             data += "\x0d\x0a"
         protocol.dataReceived(data)
-        transport.resumeProducing()
 
         return NOT_DONE_YET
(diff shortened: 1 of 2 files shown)


hawkowl:
Twisted 16.2 has a bug in how it handles HTTP timeouts that was fixed in a 16.3 prerelease, so if pinning is required, sticking on 16.1 is better until Autobahn can move to 16.3.

meejah:
Ah, okay thanks I'll change the pin

oberstet:
I am implementing the multi transport now: https://github.com/crossbario/autobahn-python/compare/master...oberstet:multisocket

When this is done, we'll deprecate WebSocketResource.

oberstet:
@hawkowl how do I stick a transport onto this channel thing? I got the WebSocket / RawSocket switch already working, but not yet Twisted Web .. see https://github.com/crossbario/autobahn-python/compare/master...oberstet:multisocket

```
  File "/home/oberstet/cpy2711_10/lib/python2.7/site-packages/twisted/web/http.py", line 736, in requestReceived
    self.client = self.channel.transport.getPeer()
exceptions.AttributeError: 'NoneType' object has no attribute 'getPeer'
```

```
    self._proto.channel.transport = self.transport
exceptions.AttributeError: '_GenericHTTPChannelProtocol' object has no attribute 'channel'
```

PR #707, opened 2016-07-10
Doc rework to use the async/await syntax.

potens1:
I was not able to build the doc (sphinx complains about TypeError: _warn_node() got an unexpected keyword argument 'type', will have a look later). Could be good is someone just give an eye on it before merging (but I think all is good)


oberstet:
Ah, nice! Very much appreciated. Cheers, Tobias

PR #719, opened 2016-07-25
Correcting minor typo Twisted ApplicationRunner

a-musing-moose:
Removing erroneous space typo in autobahn.twisted.wamp:ApplicationRunner

It still worked, but looked like that it isn't how you normally style things


oberstet:
strange that it didn't fail .. thanks!

a-musing-moose:
No problem @oberstet - I had to double check it myself. Apparently it is totally valid

PR #732, opened 2016-08-19
Issue653 async await examples

meejah:
This changes all the asyncio examples to use `await` and `async def`. I also took out the conditional imports for trollius (as a separate commit, if you don't want that). There was also one file with `CRLF` line-endings still...


oberstet:
awesome! thanks for that. at some point, we'd also need to update the docs to reflect the new style .. but I merged this now as is.

PR #744, opened 2016-10-27
Fixes derive_key()'s keylen argument documentation

trevora-edge:
This patch corrects the documentation in the comments for the keylen argument of the derive_key() function to specify that the key length is expressed in bytes rather than bits.

I realize that this is a pretty minor inconsistency, but it caused some issues for me recently when I was trying to implement equivalent functionality in Java for an Android app. I couldn't find documentation on what units are used for the keylen value in the WAMP-CRA authentication challenge message and ended up consulting this code as a reference implementation. The comments suggested that the value is expressed in bits, but after some debugging we came to the conclusion that the value is expressed in bytes rather than bits.


meejah:
Yup! Thanks for the patch :)

trevora-edge:
Great! Thanks.

PR #745, opened 2016-10-27
fix support coroutine/future for asyncio WebSocketServerProtocol.onConnect

laz2:
Inspired by twisted support of onConnect method. I think with this code _onConnect correct handles futures and coroutines from onConnect.

### DIFF
--- a/autobahn/asyncio/websocket.py
+++ b/autobahn/asyncio/websocket.py
@@ -194,18 +194,21 @@ class WebSocketServerProtocol(WebSocketAdapterProtocol, protocol.WebSocketServer
     def _onConnect(self, request):
         # onConnect() will return the selected subprotocol or None
         # or a pair (protocol, headers) or raise an HttpException
-        ##
-        # noinspection PyBroadException
-        try:
-            res = self.onConnect(request)
-            if yields(res):
-                asyncio.async(res)
-        except ConnectionDeny as e:
-            self.failHandshake(e.reason, e.code)
-        except Exception as e:
-            self.failHandshake("Internal server error: {}".format(e), ConnectionDeny.INTERNAL_SERVER_ERROR)
+        fut = txaio.as_future(self.onConnect, request)
+        fut.add_done_callback(self._on_connect_callback)
+
+    def _on_connect_callback(self, fut):
+        if fut.cancelled():
+            self.failHandshake("Cancel connection", ConnectionDeny.INTERNAL_SERVER_ERROR)
         else:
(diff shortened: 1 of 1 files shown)


meejah:
Thanks for the patch!

I think this also relates (tangentially) to https://github.com/crossbario/autobahn-python/pull/748 (e.g. because this introduces more code that calls `Future()` without passing `loop=`) and I have similar questions (i.e. why don't we just use e.g. `txaio.as_future()`). Similarly, I think `txaio.is_future()` provides the same functionality as `yields()`

laz2:
Wow. Thank you. I don't know about txaio. I will try to reimplement with `txaio.as_future`.

laz2:
Reimplemented with `txaio.as_future`.

autobahn/asyncio/websocket.py
```diff
@@ -194,18 +194,21 @@ class WebSocketServerProtocol(WebSocketAdapterProtocol, protocol.WebSocketServer
...
-            res = self.onConnect(request)
-            if yields(res):
-                asyncio.async(res)
-        except ConnectionDeny as e:
-            self.failHandshake(e.reason, e.code)
-        except Exception as e:
-            self.failHandshake("Internal server error: {}".format(e), ConnectionDeny.INTERNAL_SERVER_ERROR)
+        fut = txaio.as_future(self.onConnect, request)
+        fut.add_done_callback(self._on_connect_callback)
+
```
meejah: Basically all operations on Futures/Deferreds should use txaio. So this should probably be`txaio.add_callbacks(fut, self._on_connect_callback, self._on_connect_error)` or similar. However, see discussion in the bugs (and, it's also "okay-ish" to use asyncio-only APIs in this particular module because it's `autobahn.asyncio.*` and therefore shouldn't even be imported on non-asyncio platforms -- i.e. when using Twisted).

laz2: Even, with `txaio` we can move `_onConnect` to generic `protocol.WebSocketServerProtocol`. Do we need this?

meejah: yeah, if it's possible to make it all generic I think that's a win.

laz2: Mmmmm. No (:. With `txaio` we need modify global `txaio.config.loop` before each call its api for working with multiple event loops. So now this pull request is not good.

oberstet:
superseded by https://github.com/crossbario/autobahn-python/pull/792 which also does the code refactoring

PR #748, opened 2016-11-03
Add missing loop keyword to asyncio protocol

### LINKED ISSUE #747: a Future constructor is missing the `loop` kwarg
There's a `Future` [created](https://github.com/crossbario/autobahn-python/blob/master/autobahn/asyncio/websocket.py#L103) by the `_consume` method of `autobahn.asyncio.websocket.WebSocketAdapterProtocol`.

It is not passing the `loop` kwarg, which leads to `NotImplementedError`when the implicit default 'global' loop is not being used. 

To reproduce, for example run an autobahn websocket connection test using `py.test` (with help from the `pytest-asyncio` package that provides some useful asyncio testing fixtures). Use the `pytest.mark.asyncio` fixture, and make sure to set the `forbid_global_loop`parameter to `False`. Upon connect, the NotImplementedError will be raised. 

I am sure there are easier ways to reproduce this, though, but I am unfamiliar with asyncio loop policies (or whatever lower-level stuff) so it'd take some more effort for me to understand the details.

As for fixing the issue, the loop appears to be available in the protocol as `self.transport._loop`, and I can confirm that passing it as `loop`kwarg fixes the issue.

soulmerge:
The WebSocketAdapterProtocol was missing the loop keyword, causing a
RuntimeError if the server was not using the global asnycio loop.

This commit resolves #747.

### DIFF
--- a/autobahn/asyncio/websocket.py
+++ b/autobahn/asyncio/websocket.py
@@ -100,7 +100,7 @@ class WebSocketAdapterProtocol(asyncio.Protocol):
         self.transport = None
 
     def _consume(self):
-        self.waiter = Future()
+        self.waiter = Future(loop=self.factory.loop)
 
         def process(_):
             while len(self.receive_queue):


oberstet:
pls lets first see where the discussion on #747 goes to

PR #750, opened 2016-11-04
Issue747 future loops

meejah:
Okay, here's a straw-man solution for #747 -- it at least honors the "loop=" arg to WebSockets (unless that's None, in which case it uses `txaio`'s loop, if configured). Includes a simple test.

Something puzzling: I had to change the `protocol=` attribute for the factory to actually get an asyncio-specific protocol instance from the factory...

I also changed a couple RawSocket bare `Future()`s


oberstet:
I think this is good as a stop gap, without anticipating what results from the #747 discussion

PR #754, opened 2016-11-12
Implement client-side event retention

hawkowl:
Implementing the client-side requirements for https://github.com/wamp-proto/wamp-proto/commit/17c70b5acc8bab04f3573053a931def43c8679e6

### DIFF
--- a/autobahn/wamp/message.py
+++ b/autobahn/wamp/message.py
@@ -1114,10 +1114,10 @@ class Publish(Message):
                  eligible=None,
                  eligible_authid=None,
                  eligible_authrole=None,
+                 retain=None,
                  enc_algo=None,
                  enc_key=None,
-                 enc_serializer=None,
-                 retain=None):
+                 enc_serializer=None):
         """
 
         :param request: The WAMP request ID of this request.
@@ -1151,14 +1151,14 @@ class Publish(Message):
         :type eligible_authid: list of unicode or None
         :param eligible_authrole: List of WAMP authroles eligible to receive this event.
         :type eligible_authrole: list of unicode or None
+        :param retain: If ``True``, request the broker retain this event.
+        :type retain: bool or None
         :param enc_algo: If using payload encryption, the algorithm used (currently, only "cryptobox" is valid).
         :type enc_algo: unicode
         :param enc_key: If using payload encryption, the message encryption key.
         :type enc_key: unicode or binary
         :param enc_serializer: If using payload encryption, the encrypted payload object serializer.
         :type enc_serializer: unicode
-        :param retain: If ``True``, request the broker retain this event.
-        :type retain: bool or None
         """
         assert(type(request) in six.integer_types)
         assert(type(topic) == six.text_type)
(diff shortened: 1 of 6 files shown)


hawkowl:
@oberstet what were the cosmetics you wanted on this PR? I'm guessing the impl is all right, so I'll start work on the CB stuff assuming this is the API.

oberstet:
@hawkowl "cosmetics": the parameter ordering I mentioned in the review comment above ..

hawkowl:
@oberstet I don't see a review comment anywhere... I think you have to complete the review for it to show up?

oberstet:
@hawkowl this is very confusing .. do you see it now?

hawkowl:
@oberstet sure do, will fix

autobahn/wamp/message.py
```diff
@@ -2023,7 +2065,8 @@ def parse(wmsg):
                     topic=topic,
                     enc_algo=enc_algo,
                     enc_key=enc_key,
-                    enc_serializer=enc_serializer)
+                    enc_serializer=enc_serializer,
+                    retained=retained)
```
hawkowl: Okay, non-keyword based calling was my only concern. If we're not worried about that, I can relocate them to where it makes sense.

oberstet:
@hawkowl great! rgd reviews only visible when review is submitted/finished: hopefully I remember that trap next time.

autobahn/wamp/message.py
```diff
@@ -2067,7 +2113,7 @@ def __str__(self):
         """
         Returns string representation of this message.
         """
-        return u"Event(subscription={0}, publication={1}, args={2}, kwargs={3}, publisher={4}, publisher_authid={5}, publisher_authrole={6}, topic={7}, enc_algo={8}, enc_key={9}, enc_serializer={9}, payload={10})".format(self.subscription, self.publication, self.args, self.kwargs, self.publisher, self.publisher_authid, self.publisher_authrole, self.topic, self.enc_algo, self.enc_key, self.enc_serializer, b2a(self.payload))
+        return u"Event(subscription={0}, publication={1}, args={2}, kwargs={3}, publisher={4}, publisher_authid={5}, publisher_authrole={6}, topic={7}, enc_algo={8}, enc_key={9}, enc_serializer={9}, payload={10}, retained={11})".format(self.subscription, self.publication, self.args, self.kwargs, self.publisher, self.publisher_authid, self.publisher_authrole, self.topic, self.enc_algo, self.enc_key, self.enc_serializer, b2a(self.payload), self.retained)
```
hawkowl: @oberstet cool, no problem. Do we want this on any of the other objects?

hawkowl:
@oberstet yeah, I think it's so that it's one event & email, whilst the old way of doing it, the person whose PR it is would get like twenty emails trickling in... most of which don't make sense out of context.

autobahn/wamp/message.py
oberstet: I don't understand .. "get_retained" = SUBSCRIBE, "retained" => EVENT. what other WAMP messages do you mean?

oberstet:
@hawkowl ah, ok. that is valid reason (not wanting to receive dozens of notifications) .. I see.

autobahn/wamp/message.py
oberstet: Ah, right. Yes, like you say: EVENT, SUBSCRIBE and PUBLISH have the flag(s), and in each case, the respective parameter order should be like I indicated for EVENT.

hawkowl: Cool, can do.

hawkowl:
@oberstet I think I got them all now? And in the ones where I had to re-arrange, I removed the Python 2.6 specific explicit `{}` numbering junk... don't have to support that anymore! :D

oberstet:
perfect. yeah, the numbering stuff was annoying and error prone ..

PR #759, opened 2016-11-27
Properly handle onConnect if it is an async method

### LINKED ISSUE #752: Unable to allow "file://" as allowed_origin for WebSocket
It should be possible to allow `file://` as an allowed origin - both by explicitly adding it, and also by pattern (eg `*://*` or something).

meejah:
Includes a simple unit-test, and should close #752


meejah:
Hmm, it seems my test isn't getting run at all by the CI stuff ... :/

PR #768, opened 2016-12-12
Pre-parse configuration to _Transport class

meejah:
This gives a single place to deduce defaults (so, for example, you can specify a transport by URI alone if you want defaults). `_Transport` is also used to store state about the transport (number of connects etc), simplifying the main re-connect loop. This means: all internal users of `_Transport` don't have to know what defaults are allowed, etc *and* the incoming configuration dict isn't mangled/updated by any of the normalization / validation code.

Parsing / validation of configuration is also improved to cover serializers. And a few unit-tests, even.

So, for example, a valid configuration is `{"uri": "ws://example.com/ws"}` because "websocket" is the default, and if no "endpoint" is provided, we deduce one from the URI (which must be provided). The above config is the same as:

```
{
    "type": "websocket",
    "uri": "ws://example.com/ws",
    "endpoint": {
        "type": "tcp",
        "host": "example.com",
        "port": 80
    }
}
```

### DIFF
--- a/autobahn/twisted/component.py
+++ b/autobahn/twisted/component.py
@@ -35,7 +35,6 @@ from twisted.internet.interfaces import IStreamClientEndpoint
 from twisted.internet.endpoints import UNIXClientEndpoint
 from twisted.internet.endpoints import TCP4ClientEndpoint
 from twisted.internet.error import ReactorNotRunning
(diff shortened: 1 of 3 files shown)


oberstet:
Looks good! Seems to fit into the picture ..

PR #792, opened 2017-03-18
allow WebSocketServerProtocol.onConnect to return a future/deferred; …

oberstet:
…refactor code

### DIFF
--- a/autobahn/websocket/protocol.py
+++ b/autobahn/websocket/protocol.py
@@ -2757,9 +2757,12 @@ class WebSocketServerProtocol(WebSocketProtocol):
 
                 def forward_error(err):
                     if isinstance(err.value, ConnectionDeny):
+                        # the user handler explicitly denies the connection
                         self.failHandshake(err.value.reason, err.value.code)
                     else:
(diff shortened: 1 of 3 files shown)


autobahn/websocket/protocol.py
```diff
@@ -2747,10 +2747,22 @@ def processHandshake(self):
...
+                #   - return a pair (subprotocol, headers)
+                #   - raise a ConnectionDeny to dismiss the client
+                f = txaio.as_future(self.onConnect, request)
+
+                def forward_error(err):
+                    if isinstance(err.value, ConnectionDeny):
+                        self.failHandshake(err.value.reason, err.value.code)
+                    else:
+                        self.log.debug("Unexpected exception in onConnect ['{err.value}']", err=err)
+                        return self.failHandshake("Internal server error: {}".format(err.value), ConnectionDeny.INTERNAL_SERVER_ERROR)
```
meejah: I think it's slightly better to log this at "info" level, and log a full stacktrace at debug -- e.g. if there's a syntax error or something in your onConnect this makes it easier to discover. e.g. `self.log.debug("{tb}", tb=txaio.failure_format_traceback(err))`.

PR #798, opened 2017-03-27
A decorator-based API for subscribe and register

meejah:
This is a decorator-based API for components to subscribe and register functions. This lets you do things like this:

```python
component = Component(...)

@component.subscribe(
    u"com.example.",
    options=SubscribeOptions(match=u"prefix", details_arg='details'),
)
def an_event(details=None):
    print("topic '{}'".format(details.topic))
```

**However**, this doesn't have a good story for "how do I get crossbar to run it".

One option could be config-parsing so that you can describe the topic + options.

Another option would be to define an `object` component in crossbar that is given a module to load and it looks for a "component" object which it uses to connect to crossbar. There would be a question here about what to do about the transports/connection information -- override whatever the component has with transports defined in the crossbar config, or accept what it has from code? Or it could just be an error if there's redundant transports.

### DIFF
--- a/autobahn/wamp/component.py
+++ b/autobahn/wamp/component.py
@@ -300,6 +300,50 @@ class Component(ObservableMixin):
     The factory of the session we will instantiate.
     """
 
+    def subscribe(self, topic, options=None):
+        """
+        A decorator as a shortcut for subscribing during on-join
+
+        For example::
+
+            @component.subscribe(
+                u"some.topic",
+                options=SubscribeOptions(match=u'prefix'),
+            )
+            def topic(*args, **kw):
+                print("some.topic({}, {}): event received".format(args, kw))
+        """
+        assert options is None or isinstance(options, types.SubscribeOptions)
+
+        def decorator(fn):
+
+            def do_subscription(session, details):
+                return session.subscribe(fn, topic=topic, options=options)
+            self.on('join', do_subscription)
+        return decorator
+
+    def register(self, uri, options=None):
+        """
+        A decorator as a shortcut for registering during on-join
+
+        For example::
+
+            @component.register(
+                u"com.example.add",
+                options=RegisterOptions(invoke='round_robin'),
+            )
+            def add(*args, **kw):
+                print("add({}, {}): event received".format(args, kw))
+        """
+        assert options is None or isinstance(options, types.RegisterOptions)
+
+        def decorator(fn):
+
+            def do_registration(session, details):
+                return session.register(fn, procedure=uri, options=options)
+            self.on('join', do_registration)
+        return decorator
+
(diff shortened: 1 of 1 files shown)


meejah:
The "config parsing" thing might look like some configuration in Crossbar like this (e.g. inside a "component"):

```
subscriptions: [
    {
        "topic": "com.example.",
        "handler": "module.sub.my_function",
        "options": {
            "match": "prefix"
        }
    }
]
```

The thing I *don't* like about that, though, is there's some redundancy with what the code says (i.e. you'd be spelling out "options", "topic" etc in the code as well).

A similar option might be to leverage the decorators better and have them 'mark' the functions with the options etc directly, so that the above config would become *just* the essential information:

```
subscriptions: [
    {
        "handler": "module.sub.my_function"
    }
]
```
(or maybe it's just a list of strings that are function-names). Then the crossbar code would introspect the functions (that were marked with the decorators) to pull out topic, options, etc.

meejah:
The "component" suggestion in the original comment would look kind-of like `.tac` files from Twisted. It might work like this:

```
"components": [
    {
        "type": "object",
        "module": "module.sub"
]
```
Then, crossbar would load `module.sub` and look for a variable in it called `component` that must be an Autobahn `Component`. The only question here would be: can you override connection information from crossbar's configuration? (i.e. when the component was created "in code" it probably had `transports=` set to something useful). This is also a bit on the "magic" side, although I guess the "name of the variable" could be an option in the config above too (i.e. default "component").

oberstet:
I'm not sure about the component config in CB providing actual app business stuff (registrations, subscriptions, ..) - but this independent of having decorators for regs/subs on components at the code level, right? Or do you think the former should already be given some thoughts at this stage?

Another thing that boggles me with decorators for regs/subs: what about errors? (not authorized, shared reg and previous callee gave different invocation policy, etc)

And what about progressive invocations? Does that fit into the picture?

meejah:
As for "what about crossbar?" This comes down to "what if I want to move this code to a crossbar-launched thing". So yes, mostly orthogonal, because worst-case you could just run the component as a guest and it'll "just work".

What about errors? You will see the errors as logged by Autobahn/Crossbar as if they happened in any other `session.on("join", ..)` listener (because they do). So this would be *exactly* equivalent to making a bunch of `.register(..)` calls in an on-join listener (in fact, that's literally what the decorators do).

Progressive invocations: yes those should work fine too. On the register side, you just have to pass the "details_arg" thing, which is fine via the decorators; it's the "call" side that passes `on_progress` and there's no "call" decorator (and I don't think making one makes sense). In any case, I will cook up an example to make sure.

meejah:
Okay, yeah progressive calls work fine; see example in PR.

PR #799, opened 2017-03-27
experiment: authentication API

meejah:
This is an experiment with an authentication API for "new Session".

This means normal users wouldn't have to use onConnect or onChallenge at all -- they'd just pass configuration to Component (or use `.add_authenticator` if they got a `Session` instance somehow).

Any client-type connections that need authentication could be configured (in e.g. crossbar) with the same dicts that are given to "authentication=" list in `Component`.

This would need fleshing out to more authentication types and moving / improving the validation code to checkconfig etc.

### DIFF
--- a/autobahn/twisted/wamp.py
+++ b/autobahn/twisted/wamp.py
@@ -34,7 +34,6 @@ import txaio
 txaio.use_twisted()  # noqa
 
 from twisted.internet.defer import inlineCallbacks, succeed
-from zope.interface import Interface, implementer
 
 from autobahn.util import public
 
@@ -47,9 +46,9 @@ from autobahn.twisted.rawsocket import WampRawSocketClientFactory
 from autobahn.websocket.compress import PerMessageDeflateOffer, \
     PerMessageDeflateResponse, PerMessageDeflateResponseAccept
 
-from autobahn.wamp import protocol
+from autobahn.wamp import protocol, auth
+from autobahn.wamp.interfaces import IAuthenticator
 from autobahn.wamp.types import ComponentConfig
-from autobahn.wamp import auth
 
 
 __all__ = [
@@ -860,14 +859,6 @@ class Session(ApplicationSession):
 
 
 # experimental authentication API
-class IAuthenticator(Interface):
-
-    def on_challenge(session, challenge):
-        """
-        """
-
-
-@implementer(IAuthenticator)
 class AuthCryptoSign(object):
 
     def __init__(self, **kw):
@@ -906,9 +897,9 @@ class AuthCryptoSign(object):
 
     def on_challenge(self, session, challenge):
         return self._privkey.sign_challenge(session, challenge)
+IAuthenticator.register(AuthCryptoSign)
 
 
-@implementer(IAuthenticator)
 class AuthWampCra(object):
 
     def __init__(self, **kw):
@@ -935,4 +926,4 @@ class AuthWampCra(object):
             challenge.extra['challenge'].encode('utf8')
         )
         return signature.decode('ascii')
-
+IAuthenticator.register(AuthWampCra)
--- a/autobahn/twisted/wamp.py
+++ b/autobahn/twisted/wamp.py
@@ -28,11 +28,13 @@ from __future__ import absolute_import
 
 import six
 import inspect
+import binascii
 
 import txaio
 txaio.use_twisted()  # noqa
 
 from twisted.internet.defer import inlineCallbacks, succeed
+from zope.interface import Interface, implementer
 
 from autobahn.util import public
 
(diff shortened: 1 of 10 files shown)


oberstet:
So, overall I think this runs in the right direction. Let me understand some of the bigger things:

> ApplicationSession a subclass of Session -- and it can then be separate deprecated and removed, ultimately, if desired.

So the idea is, use the new `Session` class to get the new shiny, PEP8 API in place of `ApplicationSession`. But if you have existing code, that continues to run, since the latter is a subclass of the former, and has an implementation that forwards to the new impl. in the base class?

meejah:
Yeah, so `Session` is "the new API" and can thus evolve differently from `ApplicationSession`.

That comment isn't really to do with the authentication API per se, but was looking forward: if we want to ultimately deprecate ApplicationSession we could: switch the inheritance around; change all Autobahn-internal calls to PEP8-style (`on_join`, etc); but keep ApplicationSession as a translation-layer. That is, kind of the inverse of the relationship in there currently (i.e. an `on_join` implementation that just calls-through to the now-deprecated `onJoin`). In either case, "old" code still works. Either way is fine, that note-to-self was just that we *could* have it the other way around if desired. I think (I didn't actually-try-that ;)

meejah:
The main point of the idea in this PR is: "normal" user code would never override `on_connect` or `on_challenge` and instead would use `add_authenticator` or (more likely) pass some appropriate configuration to the `Component` kwargs -- or, e.g., static config that looks the same in crossbar if their component was configured inside crossbar.

oberstet:
Could I use `Session` like this independent of component with `ApplicationRunner` as today?

```python
session = Session()
session.add_authenticator(...)
runner.run(session)
```
---
Two smallish things:
* could we use ABCs instead of Zope interfaces? best place would be add it to autobahn/wamp/interfaces.py right away, even though only the tx implementation might be there
* examples/router/.crossbar/config-with-tls.json is gone on my fab2 branch;) it complained about missing server keys, and is it really necessary _here_ for what it is testing primarily?

oberstet:
"but keep ApplicationSession as a translation-layer": I think this approach of not hard deprecating sth, but implementation a translation layer on top of the new thing is really nice.

So the comment really hints at using inheritance to implement the translation. What about composition? I mean only attribute left on existing ApplicationSession would be sth like `._private_new_session_obj`, and the translation code in the public API impl. of ApplicationSession - app code that ever used non public obscure attributes of ApplicationSession would immediately pop up (eg CB;) - and could be fixed before migrating to Session or better Component thereafter.

Anyway: you are right, this is now a different thing, not authentication. If the code can be used like above with existing code using ApplicationRunner also, not only in conjunction with the new Component, I'm sold=)

meejah:
I haven't tried the `ApplicationRunner` thing, but in principle that should work fine. All the `Component` code does is remember the auth config and call `.add_authenticator` when it creates the Session instance. I will try that, though.

Also, yes ABCs would be way more consistent than `zope.Interface` for AB, indeed ;) Also I haven't thought too hard about what should actually *be* in that API and just had it there for easy reference for now. I think it would just be an `on_challenge` method and some way to get the name to give out in the `join` call in `on_connect` though.

As to the two configs in the example router: sure, that doesn't have really anything to do with this PR per se.

oberstet:
Sounds good!

Btw: I think the `examples/router/.crossbar/config.json` (only file left on the  Rel 0 18 1 branch) might be the only conflict with my  Rel 0 18 1 branch .. I have extended the "examples" running to test both WS and RS transports. The examples runner is really useful too;) At least one bug surfaced (well, omission) in WAMP/asyncio .. fixed

meejah:
Okay :) I actually don't remember why I thought one "non-TLS" and "TLS" config was needed. Maybe I didn't want the extra clutter? Is "having TLS support" optional?

In any case, can I just put any TLS and authentication config needed for those examples to work in "the one config"? Might be more "clutter" but then at least all the examples would work without having to copy-pasta some config snippets to the right spot...

oberstet:
@meejah I'm fine with putting everything in 1 config file with TLS, yeah. pls merge by current rel 0.18.1 branch, add as you wish, plus add any server keys needed into the repo examples folder. so such it runs without fiddling ..

meejah:
Okay, I re-based this to master, used ABC instead of zope, added WAMP-CRA and put in examples of using this via `ApplicationRunner`. This part isn't *quite* a drop-in because you don't pass an `ApplicationSession` instance to `.run()` but pass in a callable. Looks like:

```python
    runner = ApplicationRunner(...)

    def make(config):
        session = AnySessionSubclass(config)
        session.add_authenticator(
            u'cryptosign',
            authid=u'alice',
            privkey=u'6e3a302aa67d55ffc2059efeb5cf679470b37a26ae9ac18693b56ea3d0cd331c',
        )
        return session
    runner.run(make)
```

This branch isn't yet ready to merge, I think it would still need at least factoring out the config-checking stuff into checkconfig (and use that in Component). Maybe support for other auth types?

oberstet:
LGTM!

meejah:
One thing I was thinking about API-wise: would it be better to pass actual "IAuthenticator" objects to `.add_authenticator` and have the config-style stuff "higher up" (in Component)?

An issue with the way it is now is that the actual object instantiation is hidden from users so you can only use this with Autobahn-supported things. If we added a "name" attribute to IAuthenticator, this would look like:

```python
   session = Session()
   session.add_authenticator(AuthWampCra(...))
```
whereas Component would still be like:
```python
    comp = Component(..., authentication={"wampcra": {"authid": "username", "secret": "seekrit"}})
```
```

PR #804, opened 2017-03-31
make pubsub TLS example work properly out-of-the-box

meejah:
This lets the pubsub + TLS example work. The certificate + key included are self-signed ones for "localhost".


oberstet:
perfect .. no hassles, tests both TLS and non-TLS

PR #805, opened 2017-03-31
work with salted WAMP-CRA too

meejah:


oberstet:
nice! so the component alternative is really coming

PR #809, opened 2017-04-05
add initial cut of interface for payload codec

oberstet:


autobahn/wamp/interfaces.py
```diff
@@ -647,3 +667,70 @@ class IAuthenticator(object):
...
+
+    Payload transparency can be used eg for these use cases:
+
+    * end-to-end encryption of application payloads (WAMP-cryptobox)
+    * using serializers with custom user types, where the serializer and
+      the serializer implementation has native support for serializing
+      custom types (such as CBOR)
+    * transmitting MQTT payloads within WAMP, when the WAMP router is
+      providing a MQTT-WAMP bridge
+    """
```
meejah: Could also note here that "completely custom serializers can be used if you control the whole ecosystem of clients".

Might also be worth noting the obvious: "When this is enabled, the router of course cannot provide cross-serializer transparency as it normally does."

autobahn/wamp/interfaces.py
```diff
@@ -647,3 +667,70 @@ class IAuthenticator(object):
...
+    @public
+    @abc.abstractmethod
+    def encode(is_originating, uri, args=None, kwargs=None):
+        """
+        Encodes application payload.
+
+        :param is_originating: Flag indicating whether the encoding
+            is to be done from an originator (a caller or publisher).
+        :type is_originating: bool
+
```
meejah: I recently learned a fun trick: the above can be `:param bool is_originating` and therefore omit the "type" thing entirely.

autobahn/wamp/message.py
```diff
@@ -92,14 +95,69 @@
...
+    For WAMP payload transparency mode, check if the provided ``enc_algo``
+    identifier in the WAMP message is a valid one.
+
+    Currently, the only standard defined identifier are
+
+    * ``u"cryptobox"``
+    * ``u"mqtt"``
+
+    Users can select arbitrary identifiers too, but these MUST start with ``u"x_"``.
+
```
meejah: +1 :)

autobahn/wamp/message.py
```diff
@@ -92,14 +95,69 @@
...
+    :param enc_algo: The payload transparency algorithm identifier to check.
+    :type enc_algo: str
+
+    :returns: Returns ``True`` if and only if the payload transparency
+        algorithm identifier is valid.
+    :rtype: bool
+    """
+    return type(enc_algo) == six.text_type and (enc_algo in PAYLOAD_ENC_STANDARD_IDENTIFIERS or _CUSTOM_ATTRIBUTE.match(enc_algo))
+
```
meejah: I think it's slightly "more correct" to do `isinstance(enc_algo, six.text_type)` -- but that might be a better as a "change the whole codebase" PR so we're at least consistent ;)

PR #810, opened 2017-04-05
fix mqtt example

meejah:
make the MQTT example work properly with existing "payload_type = opaque" code.


oberstet:
merged, but this will change anyways .. I am going over the MQTT code after reworking the "payload transparency" stuff - see recent thread on IRC (markos_ / argronholm)

meejah:
yeah i know, but figured we might as well fix it as I was looking at it via slack convo anyway...

PR #816, opened 2017-04-13
Fix start_loop=False using asyncio

shirk3y:
Currently, there is no way to start app when `asyncio`'s loop is running. `loop.run_until_complete(coro)` is just trying to start it, resulting in exception that loop is already running. 
This fix makes `run()` to return coroutine, so it can be easily `await`ed somewhere in higher-level application code.

Also, we are using following helper to create and use `ApplicationSession` in more flexible way:

```python

async def start_wamp_session(url, realm):
    connected = asyncio.Future()

    class Component(ApplicationSession):
        async def onJoin(self, details):
            connected.set_result(self)

    runner = ApplicationRunner(url, realm)
    await runner.run(Component, start_loop=False)

    return await connected
```

I know it looks hackish, but it's currently the only way to obtain `ApplicationSession` object. The usage is simple:

```python
async def run_app(wamp_url, wamp_realm, http_bind):
    # We can start multiple services here
    http_server = await start_http_server(http_bind)

    wamp_session = await start_wamp_session(wamp_url, wamp_realm)
    await wamp_session.register(handler_defined_somewhere, 'procedure')
    await wamp_session.publish('some_topic', 'foo')

    # We can pass WAMP Session object to another service that for example
    # registers new procedure
    yet_another_service = await start_yet_another_service(wamp_session)


if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(run_app())
    loop.run_forever()
    loop.close()
```

### DIFF
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -241,12 +241,13 @@ class ApplicationRunner(object):
         txaio.use_asyncio()
         txaio.config.loop = loop
         coro = loop.create_connection(transport_factory, host, port, ssl=ssl)
-        (transport, protocol) = loop.run_until_complete(coro)
 
         # start a asyncio loop
         if not start_loop:
-            return protocol
+            return coro
         else:
+            (transport, protocol) = loop.run_until_complete(coro)
+
             # start logging
             txaio.start_logging(level='info')
 


meejah:
Hmm. I wonder how I get my documentation commit to appear here? I'll just submit another PR...

Issue #336, opened 2015-02-12
Make @subscribe and @register accept options


deceze:
The `@subscribe` and `@register` decorators should accept the same options as the corresponding methods in `ApplicationSession`, e.g.:

``` python
@register('foo', RegisterOptions(details_arg='bar', discloseCaller=True))
def baz(bar):
  ..
```

Perhaps even:

``` python
@register('foo', details_arg='bar', discloseCaller=True)
```

This would make code which relies on `RegisterOptions`/`SubscribeOptions` much more readable, since it keeps the annotations inline instead of requiring a fallback to `onJoin` callbacks.

meejah:
Although not nearly as general as the above,  it's possible to set the options for *all* @register methods in your class by passing an `options=` arg like `self.register(self, options=...)` when you do the registration.

Something like the above would be good, though.

oberstet:
This will land via https://github.com/crossbario/autobahn-python/pull/824

oberstet:
ok, landed

PR #824, opened 2017-04-17
Register/Subscribe Decorator Improvements

OOPMan:
A small improvement which allows kwargs to be passed in to the register/subscribe decorators which allows for construction of RegisterOptions  and SubscriveOptions in ApplicationSession.register and ApplicationSession.subscribe

### DIFF
--- a/autobahn/wamp/uri.py
+++ b/autobahn/wamp/uri.py
@@ -142,7 +142,7 @@ class Pattern(object):
         :param target: The target for this pattern: a procedure endpoint (a callable),
            an event handler (a callable) or an exception (a class).
         :type target: callable or obj
-        
+
         :param options: An optional options object
         :type options: None or RegisterOptions or SubscribeOptions
         """
@@ -224,7 +224,7 @@ class Pattern(object):
     def options(self):
         """
         Returns the Options instance (if present) for this pattern.
-        
+
         :return: None or the Options instance
         :rtype: None or RegisterOptions or SubscribeOptions
         """
@@ -235,8 +235,8 @@ class Pattern(object):
     def uri_type(self):
         """
         Returns the URI type of this pattern
-        
-        :return: 
+
+        :return:
         :rtype: Pattern.URI_TYPE_EXACT, Pattern.URI_TYPE_PREFIX or Pattern.URI_TYPE_WILDCARD
         """
         return self._type
@@ -312,10 +312,10 @@ class Pattern(object):
 def register(uri, options=None):
     """
     Decorator for WAMP procedure endpoints.
-    
+
     :param uri:
     :type uri: str
-    
+
     :param options:
     :type options: None or RegisterOptions
     """
@@ -332,10 +332,10 @@ def register(uri, options=None):
 def subscribe(uri, options=None):
     """
     Decorator for WAMP event handlers.
-    
+
     :param uri:
     :type uri: str
-    
+
     :param options:
     :type options: None or SubscribeOptions
     """
--- a/autobahn/wamp/uri.py
+++ b/autobahn/wamp/uri.py
@@ -231,6 +231,17 @@ class Pattern(object):
         return self._options
 
     @public
+    @property
+    def uri_type(self):
+        """
+        Returns the URI type of this pattern
+        
+        :return: 
+        :rtype: Pattern.URI_TYPE_EXACT, Pattern.URI_TYPE_PREFIX or Pattern.URI_TYPE_WILDCARD
+        """
+        return self._type
+
+    @public
     def uri(self):
         """
         Returns the original URI (pattern) for this pattern.
@@ -239,12 +250,6 @@ class Pattern(object):
         :rtype: str
         """
         return self._uri
-
-    def subscribe_options(self):
-        if self._type == Pattern.URI_TYPE_WILDCARD:
-            return SubscribeOptions(match=u"wildcard")
-        else:
-            return SubscribeOptions(match=u"exact")
 
     def match(self, uri):
         """
--- a/autobahn/wamp/uri.py
+++ b/autobahn/wamp/uri.py
@@ -28,6 +28,7 @@
 from __future__ import absolute_import
 
 import re
+
 import six
 
 from autobahn.util import public
@@ -282,7 +283,7 @@ class Pattern(object):
 
 
 @public
-def register(uri):
+def register(uri, **options):
     """
     Decorator for WAMP procedure endpoints.
     """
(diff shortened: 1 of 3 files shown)


oberstet:
So this is part 1, that actually only records the subscribe/register options, right?

From looking at the change, this looks good. We should make a conscious decision about these alternatives though:

a) in the decorator, not bind `**options`, but add a 2nd positional arg that needs to be `SubscribeOptions` or `RegisterOptions` respectively
b) inside the decorator recording, not record a pair with 2nd item `**options`, but record instances of `SubscribeOptions` or `RegisterOptions`.

Both a) and b) will blow up a wrong option ("foobar=23") early on, not later when actually triggering the decorators, because the constructors of `SubscribeOptions` and `RegisterOptions` are supposed to do tight type/value checking.

oberstet:
@meejah opinions?

OOPMan:
@oberstet This is should be part 2 as well actually. I made changes in `protocol.py` which *should* handle the recorded register/subscribe options as well, unless I missed how things work completely :-)

With regards to options a) and b) I think that options b) would be cleaner but I will defer to your decision on this regardless.

OOPMan:
Also, I suppose I should write some additional unit tests for this new code?

meejah:
We could use an `options=` kwarg instead of positional (so as to not break existing users). But yes, I agree passing the actual options instance is better.

autobahn/wamp/uri.py
```diff
@@ -282,29 +283,35 @@ def is_exception(self):
 
 
 @public
-def register(uri):
+def register(uri, **options):
     """
```
meejah: Just to clarify my comment, this would become: `register(uri, options=None)` and "options" should be checked for being a `RegisterOptions` instance (or None) in the decorator itself.

oberstet: This is actually the nicest API (better than what I proposed in a) above) - no breakage of existing code, explicit, safe to use. So let's go with this.

oberstet:
@OOPMan you are right of course, there is no part 2 needed;)

There are 2 details left I'd like to raise / discuss (as we are on this anyway now):

a) What is the semantics when I provide a decorator with `options` _and_ I provide `options` to the `session.subscribe` later? Which one takes precedence? Right now, it is always the latter (`subopts = options or pat.subscribe_options()`), and I'm not sure this makes sense. I'd probably expect being able to provide default options in `session.subscribe` and then override these specifically on some decorators.

b) Recording the decorator infos as a list of tuples (within `_wampuris`): right now, we'd record pairs `(Pattern, RegisterOptions | SubscribeOptions)`. From an extensibility/maintainability perspective, maybe it would be better to put the Options things into the Patterns as an attribute:

```python
class Pattern(object):
    def __init__(self, uri, target, options=None):
        pass
```

The whole purpose of `Pattern` is to store and transmit the decorator infos plus some aux code.

Ok, the naming "Pattern" is slightly misleading .. I haven't got a better one though ..

OOPMan:
@oberstet With regards to the two details

a) I actually wanted to get your input on this. Right now the way I wrote it preserves the precedence of the `options` passed in to `session.register` and `session.subscribe`. So if you pass in `options` to one of those methods it overrides what was defined using the decorator. However, I'm not sure this is actually a good thing. If one decorates multiple functions in order to `register`/`subscribe` them and one provides those functions with differing `RegisterOptions` and/or `SubscribeOptions` then is it really appropriate for all those different options to be overriden with the single input to `session.register` or `session.subscribe`. My instinct is no but maybe you guys feel differently?

b) Putting that stuff on the `Pattern` sounds way nicer than using tuples. I merely used tuples so as to keep my additional implementation details separate from the base behaviour. Putting it on the pattern would also mean I could get rid of the changes I had to make to the one test case.

Naming is one of those hard problems iirc, along with caching ;-)

autobahn/wamp/uri.py
OOPMan: I feel like this would leave us with decorators that look a little bloated but I guess it would also make the usage more consistent with that of the `session.register` and `session.subscribe` methods. Shall I make this change then?

oberstet: Yes please! It is slightly more "verbose" or more "explicit" depending on how you look at it.

oberstet:
@OOPMan 

a) I agree: `Options` on a decorator should take precedence over `Options` provided to `subscribe` later. This is a change in the current behavior. But this whole decorator thing was "experimental" up to now, and if we now move it into a supported/polished level, now is the time to make that change, document it, unit test it, ..;)

b) Great! So let's put it onto `Pattern`

OOPMan:
@oberstet @meejah Great, I'll make the recommended changes and update the PR

meejah:
I agree: if you provide an options to the `session.register` it will be a fall-back only if the options from the decorator itself was `None`.

Thanks for the PR!

Also, for future PRs, I'd try to avoid changes like moving the imports around (unless there's a reason?) Generally, I try to "group" imports with standard-library imports first and autobahn/local imports in the last "group" but this isn't in our code-style (and probably not consistent in the code anyway ;)

Worth mentioning, too: if you're comfortable re-basing or otherwise changing commits "after the fact" it's fine for you to just "force-push" onto the same branch (for your "own" branches/PRs). I mention it because some projects don't like this...

OOPMan:
@meejah Sorry, the imports were probably moved around by IntelliJ. I keep forgetting to disable that feature. Will fix those changes.

Good to know I can do some evil force pushing :-)

OOPMan:
@oberstet @meejah Okay, I have made the changes requested. They ran through tox locally with no issues so let's see how Travis does with them. I still want to add some unit tests but I don't think I'll get around to that tonight.

You guys may wonder why I removed the `subscribe_options` method on `Pattern` and relocated that logic to the `subscribe` method on `ApplicationSession`. It seemed to me that everything in the `Pattern` class was generic with the exception of that one method. A usages check confirmed that `subscribe_options` was only ever used in `session.subscribe` and it thus seemed to me to be an opportunity to clean up that piece of the code a little. It doesn't seem like it makes much sense that you can call `subscribe_options` on a Pattern that isn't a handler.

OOPMan:
Not sure what this flake8 error is? Getting it locally too. Any ideas?

oberstet:
@OOPMan there are blanks on a newline .. eg line 145 has a couple of space instead of just a newline

oberstet:
@OOPMan moving the code from `Pattern.subscribe_options` to protocol.py: absolutely. makes sense.

oberstet:
So last thing is making flake8 happy (remove spaces). This is ready to merge otherwise. Thanks for contributing! You are fixing https://github.com/crossbario/autobahn-python/issues/336 btw;)

OOPMan:
@oberstet @meejah Okay, landed commits to fix flake8 issues. Also, added two test cases to check the new decorator and Pattern features are working. 

Quick question, I see it looks like the Pattern supports syntax like `my.com.app.something.<parm:str>`. Is this coming to the next release? Will it be supported by Crossbar?

oberstet:
@OOPMan Cool! +1 for more unit tests. Rgd the `<parm:str>` etc feature: this has been in there for long, but left undocumented, and probably unused by apps. It is a mere client side magic thing. There is nothing different at the WAMP wire level, and hence Crossbar.io already "supports it". There is more magic for automapping WAMP error URI into custom exceptions: https://github.com/crossbario/autobahn-python/blob/master/examples/twisted/wamp/rpc/errors/frontend.py#L38

PR #828, opened 2017-04-27
Event retention examples

meejah:


oberstet:
Yeah, docs, examples=) Thanks!

PR #830, opened 2017-04-29
Added support for the X-Forwarded-For HTTP header

agronholm:
When the websocket server is sitting in front of one or more front end web servers, this allows the real client from the X-Forwarded-For header to be used as the peer address.


meejah:
LGTM :+1:

oberstet:
So this is essentially about putting the actual client IP into `peer`? As in `self.peer = trusted_addresses[0]`? I don't have enough experience here to judge whether that will work in practical scenarios, but the code looks good in principle. Merging ..

agronholm:
I do this on the application level at the moment, and several web frameworks have similar code in place. This should drastically help decipher logging output when it shows the actual client IPs there instead of the local address of the nginx or apache server sitting in front of Crossbar.

PR #837, opened 2017-05-13
Ensuring drop reason always gets propogated

ashafer01:
This `reason` message contains important debugging information and should be made available when needed.


meejah:
:+1: LGTM

PR #842, opened 2017-06-01
Fix onConnect docstring regarding return value

maruno:
The docstring mentioned a ConnectionAccept to be returned, whilst this model exists it doesn't seem to be used and the actual expected return value is str.


maruno:
Not sure about this Travis check? I've only changed a docstring from master, your master is in a broken state?

oberstet:
@maruno the CI is broken because pytest-asyncio dropped support for Py < 3.5 - awesome. https://github.com/crossbario/autobahn-python/issues/846

Issue #838, opened 2017-05-15
Config/Settings to ApplicationRunner (openHandshakeTimeout)


TonyRehley:
Hi,
How can I override default openHandshakeTimeout param?
I am going run ApplicationRunner with greater handshake timout in python code.
I have tried, but nothing:
`subscriber = ApplicationRunner(url=u"wss://api.....", realm=u"realm1",extra={'openHandshakeTimeout':10})`

jonathanstrong:
+1 struggling to find where to change this as well.

AlleoNET:
Some here some developers to answer?

oberstet:
This isn't currently exposed at the `ApplicationRunner` level (only on the underlying `WebSocketFactory` level).

michael-wise:
To clarify, there is currently no access via ApplicationSession or ApplicationRunner? Is there a single place to hard-patch at the module level?

oberstet:
@michael-wise monkey patching will not work, as the underlying factory is a locally scoped variable: https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/wamp.py#L256

myth:
I'm kinda stuck waiting for this fix to be able to work with a server under heavy load that usually takes longer to complete handshake.

The hard-coded values in https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/wamp.py#L256 would still override the fix in #844? Would love it if you would correct that before doing the next release.

I tried using the head of master after #844, and passing the options to the `extra` kwarg in the ApplicationRunner, which I assume should get passed as these new component options, but to no avail.

oberstet:
@myth curious, what server exactly? could you describe your setup?

myth:
Hi and thanks for the fast reply!

I'm attempting to connect to `wss://api.poloniex.com`, a cryptocurrency exchange, to get the ticker and orderbook feeds. I assume they are regularly DDoSed, but they are behind cloudflare.

I'm just prototyping the application for now, hence using the ApplicationRunner convenience, but my setup is just:

```
class Poloniex(ApplicationSession):
    def __init__(self, *args, **kwargs):
        ApplicationSession.__init__(self, *args, **kwargs)
        self.logger = getLogger(__name__)

    def on_event(self, *args):
        self.logger.info('{0}: {1} A:{2} B:{3} {4}% V:{5} H:{8} L:{9}'.format(*args))

    @inlineCallbacks
    def onJoin(self, details):
[...]

PR #844, opened 2017-06-02
Newapi transport options

meejah:
Inspired by #838 this enables passing transport options through from the `Component` configuration -- so e.g. you can set the websocket open-handshake timeout in the Component's transport configuration and it'll get passed through.

I also added some code that allows us to document options as `snake_case` and it auto-translates it to `camelCase` and tries again before reporting an error (i.e. until the options themselves are turned into snake_case). If this latter isn't the plan I can take out this commit...


meejah:
Hmm, I think the test-suite is "just broken" on py 3.3 and 3.4 ...

oberstet:
@meejah yeah, rgd CI https://github.com/crossbario/autobahn-python/issues/846

PR #848, opened 2017-06-08
fix handling of some TLS options in Component configs

meejah:


autobahn/twisted/component.py
```diff
@@ -193,7 +194,7 @@ def _create_transport_endpoint(reactor, endpoint_config):
         if endpoint_config['type'] == 'tcp':
 
             version = int(endpoint_config.get('version', 4))
-            host = str(endpoint_config['host'])
+            host = six.text_type(endpoint_config['host'])
```
oberstet: actually, I think it would be much better not to coerce the type, but to raise if the user does not provide a string. no type coercions at all (we had that, and it was painful to remove again).

oberstet:
merged via https://github.com/crossbario/autobahn-python/pull/850

PR #849, opened 2017-06-09
Examples js

meejah:
point the examples at a locally-hosted autobahn.min.js and instructions to curl it from the crossbar github repo.


oberstet:
the distribution point for ABJS is https://github.com/crossbario/autobahn-js-built, not the CB repo ..

oberstet:
merged via https://github.com/crossbario/autobahn-python/pull/851

PR #860, opened 2017-06-23
Ticket1075 revoke subscription force reregister

meejah:
Add the `force_reregister=` option to Registration, and the client-side behavior when getting a kicked registration message. Also an example of using the meta-API.

(Needs corresponding crossbar branch to actually-work)

### DIFF
--- a/autobahn/wamp/message.py
+++ b/autobahn/wamp/message.py
@@ -3158,9 +3158,10 @@ class Register(Message):
         'match',
         'invoke',
         'concurrency',
+        'force_reregister',
     )
 
-    def __init__(self, request, procedure, match=None, invoke=None, concurrency=None):
+    def __init__(self, request, procedure, match=None, invoke=None, concurrency=None, force_reregister=None):
         """
 
         :param request: The WAMP request ID of this request.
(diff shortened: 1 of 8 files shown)


meejah:
Hmm, Twisted 12 and 13 have the "too old UnitTest" dependency (or something?) for `assertRaises` to work as a context-manager. Do we definitely care about those releases?

oberstet:
@meejah awesome! comprehensive PR, nice work. I'll get to the corresponding CB PR.

oberstet:
> Do we definitely care about those releases?

IMO, those Twisted versions are really old now, so +1 for moving our minimum version for Twisted in Autobahn (in Crossbar.io we are regularily pushing to the latest releases of all deps anyways).

Question is, what minimum version of Twisted should we support in AB?

oberstet:
@meejah Twisted 15.4 or should we jump even further for minimum version?

meejah:
Let's see, releases-wise Debian "stretch" just became stable so "jessie" is now old-stable so to make any packagers lives easiest, we'd support 16.6.x+ for "stretch" (latest stable). jessie/old-stable has 14.0.2 but there is jessie-backports that has 16.2.x

So I'd say anyone packaging for latest Debian should be satisfied with anything 16.x or higher. Ubuntu should be newer. So, 15.x should be fairly safe -- and we can always bump it up later if we run into problems? (That is, might as well keep 15 if it's straightforward...?)

oberstet:
> So, 15.x should be fairly safe -- and we can always bump it up later if we run into problems

Yeah, sounds good! So we can kill 12.1 and 13.2 from Travis, which makes the CI run quicker - another nice side effect.

PR #863, opened 2017-07-02
Check for auth properties in ._args

richardartoul:
The current implementation doesn't work, but after changing it to ._args it does the right thing


meejah:
LGTM, thanks! :+1:

(I guess I should have written unit-tests for that code ....)

PR #871, opened 2017-07-10
Newapi asyncio component

meejah:
This is a version of the "component" API for asyncio as well as some changes to the (experimental) authentication API to make it more general between twisted/asyncio and easier to plug in an "unknown to Autobahn" authentication scheme.

Also tweaks the TLS component setup so you can pass `trust_root` to the client-side TLS configuration (e.g. to use a self-signed certificate, as the example router does).


oberstet:
Niice! In fact, the auth interface is in sync with the new AutobahnJava https://github.com/crossbario/autobahn-java/blob/master/autobahn/src/main/java/io/crossbar/autobahn/wamp/interfaces/IAuthenticator.java

I haven't read every line, but this is fully backwards compatible (in as, it does break stuff on the "old API"), right?

So when flake is happy, this looks good to merge for me ..

meejah:
It doesn't break anything in the old API. AFAIK.

It *does* obviously break `add_authenticator` but even early adopters using the component API shouldn't be using that (that is, they should just be passing dicts to components, which still take the same args).

It *miiiight* conceivably break someone who opted into the new-api `Session` class (i.e. the pep8 named-stuff) who *also* did their own auth stuff, but that's ... very unlikely (and marked as "experimental" anyway).

oberstet:
sounds good!

flake is complaining about `yield from`:

* https://travis-ci.org/crossbario/autobahn-python/jobs/251867102#L229
* https://github.com/meejah/AutobahnPython/blob/61ece45c2751dcb787004e3be7fd08e907ad9db9/autobahn/asyncio/component.py#L278

meejah:
On my machine the only way I fixed that was to run flake8 in a python3 (instead of 2) venv. I'd also like to make a few basic unit-tests before merging. I'm not sure if there's any way *except* running flake8 under python3 -- I was also considering: is it better to use "async def" etc syntax in here if we're going to use "python3-only" syntax at all? Or is it better to go the *other* way and only use txaio? (We do use Twisted-only syntax/classes in autobahn/twisted/* packages, though)

oberstet:
> I was also considering: is it better to use "async def" etc syntax in here if we're going to use "python3-only" syntax at all? Or is it better to go the *other* way and only use txaio?

I think this hits the nail. Either one way pure (async def) or the other way (txaio). The `yield from` syntax was transitional only in Py I think.

So I think the story to pitch to _users_ (with no "legacy" code base or restrictions) should be: go Py 3.5+ and use the new syntax.

However, _in general_, as a _library_, the "we support every Py" story is a valid and strong one, and for Twisted users I would say this is important.

Now, for asyncio the situation is probably different. Boils down to: what is the asyncio user base sitting on 3.3/3.4 and on 2.7/trollius.

If we go for new syntax within AB/asyncio, we should officially deprecate AB/asyncio/Py<3.5 though. At least ask on the mailing list before. So this aspect maybe deserves some discussion / feedback ..

meejah:
okay. For now I'll convert to txaio so that we don't lose any compatibility (i.e. older 3's or asyncio on 2.7)

oberstet:
merged via https://github.com/crossbario/autobahn-python/pull/872

PR #880, opened 2017-08-14
Register prefix

meejah:
This adds two new features to the `@wamp.register` decorator (and the related method that actually does the registrations):

 - adds a `prefix=` kwarg that is prepended to all registrations for an instance
 - allows passing `None` as the URI so the name is auto-discovered

Resulting code can look like:

```python
class Something(ApplicationSession):
    def onJoin(self, details):
        return self.register(self, prefix=u'com.example.')

    @wamp.register(None)
    def public_method(self):
        return 42
```

This will result in a method being registered at `com.example.public_method`

### DIFF
--- a/autobahn/wamp/interfaces.py
+++ b/autobahn/wamp/interfaces.py
@@ -552,7 +552,7 @@ class ISession(object):
 
     @public
     @abc.abstractmethod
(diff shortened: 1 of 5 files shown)


oberstet:
Looks good, cool

PR #897, opened 2017-09-14
create new event loop if already closed

om26er:
Currently if the event loop is closed ApplicationRunner.run() won't start and the message we get is: Event loop is closed.

### DIFF
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -246,8 +246,8 @@ class ApplicationRunner(object):
         # start the client connection
         loop = asyncio.get_event_loop()
         if loop.is_closed() and start_loop:
-            loop = asyncio.new_event_loop()
-            asyncio.set_event_loop(loop)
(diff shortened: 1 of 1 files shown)


meejah:
:+1: LGTM

oberstet:
https://travis-ci.org/crossbario/autobahn-python/jobs/275624065

om26er:
I have pushed a fix and verified the related tests to pass on my machine. CI still fails and I am not sure if my branch broke something.

meejah:
I think it's probably the mocks in those three tests? I.e. nothing called `is_closed` before so you'll maybe need to upgrade the mock to deal with that :( i.e. set a `loop.is_closed = Mock(return_value=False)` since I think all Mocks probably eval to "truthy" ..

om26er:
No, the current failures are unrelated (I fixed the failure that @oberstet pointed). This branch is now ready.

So apparently something broke our CI, example: https://travis-ci.org/crossbario/autobahn-python/jobs/276100410

PR #898, opened 2017-09-15
Backpolicy for #886

### LINKED ISSUE #886: Backoff policy options are not exposed
There should be a way to [override the default backoff policy](https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/wamp.py#L317-L330) in case people want a retry policy that is not exponentially increasing (constant interval, for example).

I don't know if it would be better to pass in a `backoffPolicy` object, or simply expose [some of the options available](https://twistedmatrix.com/documents/16.1.0/api/twisted.application.internet.backoffPolicy.html) in backoffPolicy.

I can patch this issue myself once I get some feedback.

noahrinehart:
This should fix #886. It just allows twisted's backOffPolicy arguments to be specified when constructing a ApplicationRunner object. 

I wasn't sure how to implement the arguments for the backOffPolicy object, so I used  a dictionary of arguments with a None checker to only use those that are defined. I hope this is the best (pythonic) approach.


oberstet:
cool thx! please note that the new component API also has that functionality (and a lot more improvements) - and the component API will be what is recommended for all new code (we only need some more docs - the code is already in use).

Issue #900, opened 2017-09-16
CI is broken


om26er:
Two issues:

ERROR:   pypy-asyncio: InterpreterNotFound: pypy
ERROR:   py33-tw154: InterpreterNotFound: python3.3

I am not sure if its related to the OS update in Travis or something.
https://travis-ci.org/crossbario/autobahn-python/builds/276187555

om26er:
this was fixed by Tobias a few days ago.

PR #901, opened 2017-09-17
fix some python3'isms

jmgurney:
In attempting to get CryptoSign working, I ran into a few python3'isms that make the cryptosign file not compatible w/ python2.

I have not run this through tests on python3 as I'm not at all familiar w/ 3.  From what I read, there may need to be a .encode('latin-1') or something similar on the _makepad call.

### DIFF
--- a/autobahn/wamp/cryptosign.py
+++ b/autobahn/wamp/cryptosign.py
@@ -137,10 +137,14 @@ class _SSHPacketReader:
         return value
 
     def get_uint32(self):
-        return int.from_bytes(self.get_bytes(4), 'big')
+        return struct.unpack('>I', self.get_bytes(4))[0]
 
     def get_string(self):
         return self.get_bytes(self.get_uint32())
+
+
+def _makepad(size):
+    return ''.join(chr(x) for x in range(1, size + 1))
 
 
 def _read_ssh_ed25519_privkey(keydata):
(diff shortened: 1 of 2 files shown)


jmgurney:
Looks like currently failing checks are issues w/ the CI environment and not this PR.

om26er:
> Looks like currently failing checks are issues w/ the CI environment and not this PR.

Yeah, I created an issue for that https://github.com/crossbario/autobahn-python/issues/900

autobahn/wamp/cryptosign.py
```diff
@@ -137,12 +137,16 @@ def get_bytes(self, size):
...
     def get_uint32(self):
-        return int.from_bytes(self.get_bytes(4), 'big')
+        return struct.unpack('>I', self.get_bytes(4))[0]
 
     def get_string(self):
         return self.get_bytes(self.get_uint32())
 
 
+def _makepad(l):
+    return ''.join(chr(x) for x in range(1, l + 1))
```
meejah: I would choose a different name for this arg (`size` maybe?) as it's not that easy to distinguish `1 and l`

meejah:
Thanks for the PR *with* tests :) :+1:
Looks good to me

jmgurney:
Ok, I've done both...  squashed them down.

After this gets merged, I have some other work that I'll make a PR for.

oberstet:
awesome, thx!

Issue #903, opened 2017-09-18
TypeError when session is left


om26er:
Below code throws TypeError when the program exits, apparently the fix is to close the event loop as discussed here: https://bugs.python.org/issue23548
```
import asyncio
import sys

from autobahn.asyncio.component import Component, run

component = Component(
    transports=[
        {
            "type": "rawsocket",
            "url": "ws://localhost/ws",
            "endpoint": {
                "type": "unix",
                "path": "/home/om26er/Documents/cb/.crossbar/crossbar.sock",
            },
            "serializer": "json",
        }
    ],
    realm=u"realm1"
)

@component.on_join
async def on_join(session, details):
    data = await session.call('com.example.rawsocket-cbor.run', command)
    print(data)
    session.leave()


if __name__ == "__main__":
    command = sys.argv[1]
    run([component])
```

The said error logs:
```
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
Traceback (most recent call last):
  File "/usr/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 139, in remove_signal_handler
  File "/usr/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

om26er:
fixed by https://github.com/crossbario/autobahn-python/pull/904

PR #904, opened 2017-09-18
Close aio event loop on exit

om26er:
fix for https://github.com/crossbario/autobahn-python/issues/903


meejah:
:+1: looks good to me

PR #906, opened 2017-09-23
Cleanups

meejah:
A few fixups, including making cryptobox work.


autobahn/wamp/component.py
```diff
@@ -547,12 +549,13 @@ def on_disconnect(session, was_clean):
                         was_clean=was_clean,
                     )
                     if not txaio.is_called(done):
-                        if was_clean:
-                            # eg the session has left the realm, and the transport was properly
-                            # shut down. successfully finish the connection
-                            txaio.resolve(done, None)
-                        else:
-                            txaio.reject(done, RuntimeError('transport closed uncleanly'))
```
oberstet: the deferred should be rejected in this case

autobahn/wamp/cryptobox.py
```diff
@@ -231,7 +231,7 @@ def decode(self, is_originating, uri, encoded_payload):
             if encoded_payload.enc_serializer != u'json':
                 raise Exception("received encrypted payload, but don't know how to process serializer '{}'".format(encoded_payload.enc_serializer))
 
-            payload = _json_loads(payload_ser)
+            payload = _json_loads(payload_ser.decode('utf8'))
```
oberstet: when does that apply? cryptobox works for me

oberstet:
merged via https://github.com/crossbario/autobahn-python/pull/909

autobahn/wamp/cryptobox.py
meejah: I couldn't get it working at all (on py3) without that line -- the other side is encoding to `utf8`-bytes and JSON wants strings not bytes...

autobahn/wamp/component.py
meejah: Yeah it does seem like that should be true -- but I changed this because it makes the shutdown wrong if the closing handshake timeout is hit. What happens then is that the component is trying to shut down, but then gets an error while closing -- so we either need a custom exception here that we catch in here: https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/component.py#L372 (and then stop trying to re-connect) .. "or something".

Basically I don't think we care at this point if the transport closed cleanly or uncleanly (as far as Component behavior) and just need to log it...

PR #912, opened 2017-09-25
Make it simpler to override transport_factory options

### LINKED ISSUE #838: Config/Settings to ApplicationRunner (openHandshakeTimeout)

frankwiles:
We recently ran into a situation when using Autobahn where we need to adjust the settings of `transport_factory`, specifically increasing the `autoPingTimeout`, in order for our application to function under certain loads.

We ended up having to sub-class ApplicationRunner and then cut-n-paste the entire `run` method to achieve this which is easily remedied with this patch. With this we would only need to override the `set_transport_protocol_options` method.  

This should fix #838 
Hope you'll consider this for inclusion in the next release.

### DIFF
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -147,6 +147,36 @@ class ApplicationRunner(object):
         self.proxy = proxy
         self.headers = headers
 
+    def set_transport_protocol_options(self, transport_factory):
+        """
+        Set default transport options
+        """
+        # client WebSocket settings - similar to:
+        # - http://crossbar.io/docs/WebSocket-Compression/#production-settings
+        # - http://crossbar.io/docs/WebSocket-Options/#production-settings
+        offers = [PerMessageDeflateOffer()]
+
+        # Function to accept permessage_delate responses from the server ..
+        def accept(response):
+            if isinstance(response, PerMessageDeflateResponse):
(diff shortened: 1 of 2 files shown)


meejah:
Thanks for the PR!

"In general" I think it would be best if nobody ever had to subclass ApplicationRunner -- what if instead of the above there was a ``set_protocol_options(self, **kwargs)`` or similar method that would remember any protocol options you set. You would have to call this method before calling ``.run()``, so code would look like:

```python
runner = ApplicationRunner(..)
runner.set_protocol_options(autoPingTimeout=6000)
runner.run(...)
```

I think the above would still address your use-case (i.e. "I want to set protocol options") but without the need to subclass. As far as implementation, I do like your re-factoring of that code to its own method but I'd keep that as an "internal" helper (i.e. prefixed with an underscore).

oberstet:
fwiw, I agree with @meejah on this comments ..

frankwiles:
Yeah @meejah that's a much better idea.  I'll adjust this and branch in that direction.

PR #913, opened 2017-09-28
fix formatting string

meejah:


om26er:
looks good, wonder if I should merge that change.

meejah:
If you're approving it, I can merge ;) No tests, but it's pretty obviously correct

om26er:
sure, go ahead :)

PR #914, opened 2017-09-29
[WIP] component.start method refactor: handle retries and fire connectfailure events

spr0cketeer:
Requesting comments on this approach before going further.

Only asyncio component done for now.

1. Component transport dict option will now work:
    max_retries: int # -1 for unlimited, 0 for a one time connect attempt
   Connect attempts will respect multiple transport dicts max_retries option.

2. Added 'connectfailure' component lifecycle event to return the connect exception as currently there's no way for a component user to be notified of connection errors.

3. Tweaked the max_retries handling code as it was actually working as a max_tries option

4. Added logging option to component.start() as logging was only working if using run([component]) but not when calling component.start if managing the event loop yourself.

First time tinkering with asyncio and txaio so a couple of things I wasn't sure of:
 
5. How to handle the gathered future returned from self.fire('connectfailure', fail)  if component user's component.on('connectfailure', func) handler raises an exception?  Right now any raised exception in the handler is swallowed and library goes on to handle the original failure exception - acceptable?

6. When there are no more transport connect attempts left I want to reject the done_f future with a meaningful exception, not just return the exception from the last failed connect attempt, so used the try/except approach to reject component future with a RuntimeError - is there a better exception to use?  TransportLost looks like a WAMP level exception for when an established session loses its transport.

### DIFF
--- a/autobahn/asyncio/component.py
+++ b/autobahn/asyncio/component.py
@@ -289,7 +289,7 @@ class Component(component.Component):
             assert(False), 'should not arrive here'
 
     # async function
-    def start(self, loop=None):
+    def start(self, loop=None, log_level=None):
         """
         This starts the Component, which means it will start connecting
         (and re-connecting) to its configured transports. A Component
@@ -304,6 +304,9 @@ class Component(component.Component):
             "done" or with an error if something went wrong.
         """
 
+        if log_level is not None:
+            txaio.start_logging(level=log_level)
+
         if loop is None:
             self.log.warn("Using default loop")
             loop = asyncio.get_event_loop()
@@ -312,92 +315,108 @@ class Component(component.Component):
         # specified in the docstring.
         done_f = txaio.create_future()
 
-        # transports to try again and again ..
-        transport_gen = itertools.cycle(self._transports)
+        # this is a 1-element list so we can set it from closures in
+        # this function
+        transport_gen = [None]
+
+        # Create a generator of transports that .can_reconnect()
+        def set_valid_transports():
+            valid_transports = [tp for tp in self._transports if tp.can_reconnect()]
+            transport_gen[0] = itertools.cycle(valid_transports)
+            return len(valid_transports)
+
+        set_valid_transports()
 
         # issue our first event, then start the reconnect loop
         f0 = self.fire('start', loop, self)
 
-        # this is a 1-element list so we can set it from closures in
-        # this function
-        reconnect = [True]
-
         def one_reconnect_loop(_):
             self.log.debug('Entering re-connect loop')
-            if not reconnect[0]:
-                return
 
-            # cycle through all transports forever ..
-            transport = next(transport_gen)
+            # cycle through all connectable transports until max_retries hit
+            transport = next(transport_gen[0])
 
-            # only actually try to connect using the transport,
-            # if the transport hasn't reached max. connect count
-            if transport.can_reconnect():
-                delay = transport.next_delay()
-                self.log.debug(
-                    'trying transport {transport_idx} using connect delay {transport_delay}',
-                    transport_idx=transport.idx,
-                    transport_delay=delay,
-                )
+            delay = transport.next_delay()
+            self.log.debug(
+                'trying transport {transport_idx} using connect delay {transport_delay}',
+                transport_idx=transport.idx,
+                transport_delay=delay,
+            )
 
-                delay_f = asyncio.ensure_future(txaio.sleep(delay))
+            delay_f = asyncio.ensure_future(txaio.sleep(delay))
 
-                def actual_connect(_):
-                    f = self._connect_once(loop, transport)
+            def actual_connect(_):
+                connect_f = self._connect_once(loop, transport)
 
-                    def session_done(x):
-                        txaio.resolve(done_f, None)
+                def session_done(x):
+                    txaio.resolve(done_f, None)
 
-                    def connect_error(fail):
-                        if isinstance(fail.value, asyncio.CancelledError):
-                            reconnect[0] = False
-                            txaio.reject(done_f, fail)
-                            return
+                def connect_error(fail):
+                    unrecoverable_error = False
+                    self.fire('connectfailure', self, fail.value)
 
-                        self.log.debug(u'component failed: {error}', error=txaio.failure_message(fail))
+                    if isinstance(fail.value, asyncio.CancelledError):
+                        unrecoverable_error = True
+
+                    self.log.debug(u'component failed: {error}', error=txaio.failure_message(fail))
+                    self.log.debug(u'{tb}', tb=txaio.failure_format_traceback(fail))
+                    # If this is a "fatal error" that will never work,
+                    # we bail out now
+                    if isinstance(fail.value, ApplicationError):
+                        if fail.value.error in [u'wamp.error.no_such_realm']:
+                            unrecoverable_error = True
+                            self.log.error(u"Fatal error, not reconnecting")
+
+                        self.log.error(u"{msg}", msg=fail.value.error_message())
+
+                    elif isinstance(fail.value, OSError):
+                        # failed to connect entirely, like nobody
+                        # listening etc.
+                        self.log.info(u"Connection failed: {msg}", msg=txaio.failure_message(fail))
+
+                    elif _is_ssl_error(fail.value):
+                        # Quoting pyOpenSSL docs: "Whenever
+                        # [SSL.Error] is raised directly, it has a
+                        # list of error messages from the OpenSSL
+                        # error queue, where each item is a tuple
+                        # (lib, function, reason). Here lib, function
+                        # and reason are all strings, describing where
+                        # and what the problem is. See err(3) for more
+                        # information."
+                        self.log.error(u"TLS failure: {reason}", reason=fail.value.args[1])
+                        self.log.error(u"Marking this transport as failed")
+                        transport.failed()
+                    else:
+                        self.log.error(
+                            u'Connection failed: {error}',
+                            error=txaio.failure_message(fail),
+                        )
+                        # some types of errors should probably have
+                        # stacktraces logged immediately at error
+                        # level, e.g. SyntaxError?
                         self.log.debug(u'{tb}', tb=txaio.failure_format_traceback(fail))
-                        # If this is a "fatal error" that will never work,
-                        # we bail out now
-                        if isinstance(fail.value, ApplicationError):
-                            if fail.value.error in [u'wamp.error.no_such_realm']:
-                                reconnect[0] = False
-                                self.log.error(u"Fatal error, not reconnecting")
-                                txaio.reject(done_f, fail)
-                                return
 
-                            self.log.error(u"{msg}", msg=fail.value.error_message())
-                            return one_reconnect_loop(None)
+                    if not unrecoverable_error:
+                        if not transport.can_reconnect():
+                            # transport failed or max_retries hit so remove
+                            # it from valid transports
+                            if not set_valid_transports():
+                                # No more valid transports to try
+                                unrecoverable_error = True
+                                err_msg = u"Connection failed: Exhausted all transport connect attempts"
+                                self.log.info(err_msg)
+                                try:
+                                    raise RuntimeError(err_msg)
+                                except RuntimeError:
+                                    fail = txaio.create_failure()
 
-                        elif isinstance(fail.value, OSError):
-                            # failed to connect entirely, like nobody
-                            # listening etc.
-                            self.log.info(u"Connection failed: {msg}", msg=txaio.failure_message(fail))
-                            return one_reconnect_loop(None)
+                    if unrecoverable_error:
+                        txaio.reject(done_f, fail)
+                        return
 
-                        elif _is_ssl_error(fail.value):
-                            # Quoting pyOpenSSL docs: "Whenever
-                            # [SSL.Error] is raised directly, it has a
-                            # list of error messages from the OpenSSL
-                            # error queue, where each item is a tuple
-                            # (lib, function, reason). Here lib, function
-                            # and reason are all strings, describing where
-                            # and what the problem is. See err(3) for more
-                            # information."
-                            self.log.error(u"TLS failure: {reason}", reason=fail.value.args[1])
-                            self.log.error(u"Marking this transport as failed")
-                            transport.failed()
-                        else:
-                            self.log.error(
-                                u'Connection failed: {error}',
-                                error=txaio.failure_message(fail),
-                            )
-                            # This is some unknown failure, e.g. could
-                            # be SyntaxError etc so we're aborting the
-                            # whole mission
-                            txaio.reject(done_f, fail)
-                            return
+                    return one_reconnect_loop(None)
 
-                    txaio.add_callbacks(f, session_done, connect_error)
+                txaio.add_callbacks(connect_f, session_done, connect_error)
 
             txaio.add_callbacks(delay_f, actual_connect, error)
 
--- a/autobahn/asyncio/component.py
+++ b/autobahn/asyncio/component.py
@@ -301,7 +301,7 @@ class Component(component.Component):
 
         if loop is None:
             self.log.warn("Using default loop")
-            loop = asyncio.get_default_loop()
+            loop = asyncio.get_event_loop()
 
         return self._start(loop=loop)
 
--- a/autobahn/asyncio/component.py
+++ b/autobahn/asyncio/component.py
@@ -351,9 +351,8 @@ class Component(component.Component):
                 def session_done(x):
                     txaio.resolve(done_f, None)
 
-                def connect_error(fail):
+                def handle_connect_error(fail):
                     unrecoverable_error = False
-                    self.fire('connectfailure', self, fail.value)
 
                     if isinstance(fail.value, asyncio.CancelledError):
                         unrecoverable_error = True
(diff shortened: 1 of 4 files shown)


spr0cketeer:
@meejah please review when you get a chance

autobahn/asyncio/component.py
```diff
@@ -312,101 +315,118 @@ def start(self, loop=None):
...
+
+                        self.log.error(u"{msg}", msg=fail.value.error_message())
+
+                    elif isinstance(fail.value, OSError):
+                        # failed to connect entirely, like nobody
+                        # listening etc.
+                        self.log.info(u"Connection failed: {msg}", msg=txaio.failure_message(fail))
+
+                    elif _is_ssl_error(fail.value):
+                        # Quoting pyOpenSSL docs: "Whenever
```
meejah: (Optional for this PR but ..) this code all mirrors the Twisted side -- if it could be moved to a base-class helper (or just bare function in autobahn.wamp.component) that would be ideal

autobahn/asyncio/component.py
```diff
@@ -312,101 +315,118 @@ def start(self, loop=None):
...
-                    txaio.add_callbacks(f, session_done, connect_error)
+                def notify_connect_error(fail):
+                    chain_f = txaio.create_future()
+                    handler_f = self.fire('connectfailure', self, fail.value)
+                    txaio.add_callbacks(
+                        handler_f,
+                        lambda _: txaio.reject(chain_f, fail),
+                        lambda _: txaio.reject(chain_f, fail)
+                    )
+                    return chain_f
```
meejah: This is an interesting idea -- a "connectfailure" notification. I think it should be in a separate PR, probably -- partly so we could discuss the idea itself without delaying fixing the re-connect loop for asyncio. Also, if we *do* add a ``connectfailure`` event, the Twisted side should get it too.

What would user-code do with this? Simply feedback to the user?

autobahn/wamp/component.py
```diff
@@ -279,13 +279,15 @@ def failed(self):
     def can_reconnect(self):
         if self._permanent_failure:
             return False
-        return self.connect_attempts < self.max_retries
+        if self.max_retries == -1:
+            return True
+        return self.connect_attempts < self.max_retries + 1
```
meejah: I think this is a good idea (a "try forever" option).

autobahn/asyncio/component.py
```diff
@@ -289,7 +289,7 @@ def _connect_transport(self, loop, transport, session_factory):
             assert(False), 'should not arrive here'
 
     # async function
-    def start(self, loop=None):
+    def start(self, loop=None, log_level=None):
         """
```
meejah: I actually don't think this belongs here (nor the start_logging call) -- it is already done in the ``run`` method and calling ``start`` on a component is (at least intended as) the lower-level API. That is, if you're already controlling your own event-loop. In this case, I would presume most users would also be controlling the log-starting themselves, too.

The idea is that "most" users should just call ``run([component0, component1])`` but if you wanted more control you'd call start yourself and deal with the deferreds/futures -- adding one more line in this case for ``start_logging`` yourself seems reasonable (since you'd also have to arrange for starting the event-loop anyway).

meejah:
As to your last two comments:

5: I think if the user ``connectfailure`` handler(s) raise exceptions they should be logged (but your approach of re-raising the underlying error is correct)

6: This probably does make sense (to reject with some meaningful exception). What will happen right now on the Twisted side is it'll just fall out of the whole ``@inlineCallbacks`` method, which is the same as returning/callback-ing the future with ``None``. A RuntimeError is probably okay, but perhaps we should consider a special ``ComponentFailed`` or similar exception (in ``autobahn.wamp.component.*``). **Ideally** this might indicate how each transport failed .. but just "all transports failed" in a ``ComponentFailed`` would be better than what we have now ;)

meejah:
Just for my understanding, the itches you're addressing in this PR are: "better retry-count handling, and option of infinite or one-time connects" and "add a hook to notify about individual connection failures" correct?

autobahn/asyncio/component.py
spr0cketeer: Yep I came to the same conclusion thinking about this today - if you're managing the loop you're also managing the logging so I will remove this.

autobahn/asyncio/component.py
```diff
@@ -312,101 +315,118 @@ def start(self, loop=None):
...
-        transport_gen = itertools.cycle(self._transports)
+        connectable_transports = dict()
+
+        # Create a generator of transports that .can_reconnect()
+        def set_connectable_transports():
+            valid_transports = [tp for tp in self._transports if tp.can_reconnect()]
+            connectable_transports['count'] =  len(valid_transports)
+            connectable_transports['gen'] = itertools.cycle(valid_transports)
+
+        set_connectable_transports()
```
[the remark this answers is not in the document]
spr0cketeer: Yes it can! Now I know about it  :smile:    I only saw ```can_reconnect``` not ```_can_reconnect``` and thought I had to keep track of remaining connectable transports in the generator so I was removing any that had reached max_retries, but I'll change to ```_can_reconnect```.

autobahn/asyncio/component.py
spr0cketeer: OK I'll see what can be consolidated into the base class.  Maybe most of the start method if the full mind-bending horrors of txaio can be faced in order to replace the @inlineCallbacks code in the twisted version?  Doable you think?

autobahn/asyncio/component.py
spr0cketeer: Yeah further user-side logging of transport failures for one.

If you start a component with multiple transports this will allow user-code to "do something" with the exception raised as each failed connect attempt happens.  Without this, if one of your transports is unconnectable but another succeeds and you get a working session, your component.start() error callback won't be called so you'll never know other than by looking in logs one of your transports didn't work.

Also could be handy for one-shot connect attempts with one transport and max_retries =  0.  'connectfailure' handler will fire and you'll know quickly there's a fundamental connection error without having to look at the exception type in the component.start() error callback which could be a session level exception.

spr0cketeer:
A dedicated ComponentFailed exception sounds a good idea. 

Yeah those are the two itches.  So I guess best do as you suggest and split this up into two PRs - one to get the max_retries working in asyncio and the other for connectfailure event discussion?

autobahn/asyncio/component.py
meejah: If it *is* doable, that would be awesome! Even if we can move "some" of the code to a common place, that's a big win IMO.

autobahn/asyncio/component.py
meejah: Okay, I think that makes sense. What do you think @oberstet? (That is, adding a ``connectfailure`` notification that you can listen to on ``Component`` on both twisted +  asyncio)

spr0cketeer:
@meejah new base class version of the start method for you to have a look at.  I've left the "connectfailure" event in until @oberstet decides whether it's wanted or not.

oberstet:
rgd a new lifecycle event "connectfailure", what is the current set of events we have (in component)? what exactly would be the announced semantics of the new event?

spr0cketeer:
@oberstet :

With this additional event there would be:

start
connectfailure
connect
join
ready
leave
disconnect

connectfailure will fire on each failed connect attempt with the exception raised for the transport failing.

This got lost in meejah's line comments conversation due to pushing additional commits - some reasons for adding this event were:

Yeah further user-side logging of transport failures for one.

If you start a component with multiple transports this will allow user-code to "do something" with the exception raised as each failed connect attempt happens. Without this, if one of your transports is unconnectable but another succeeds and you get a working session, your component.start() error callback won't be called so you'll never know other than by looking in logs one of your transports didn't work.

Also could be handy for one-shot connect attempts with one transport and max_retries = 0. 'connectfailure' handler will fire and you'll know quickly there's a fundamental connection error without having to look at the exception type in the component.start() error callback which could be a session level exception.

meejah:
I think something like `connectfailure` probably makes sense -- there's currently not a good way to know *which* (if any) transports failed when you've given multiple transports to `Component`.

As far as semantics: we should decide if you get a notification *every time* a connection fails (or e.g. only when Component has decided to give up on that transport, or only 0 or 1 times for each transport).

It looks to me like the current code will do the first thing (notify on every connection failure).

meejah:
It *might* also make sense to allow user-code to decide if this counts as a "fatal error" (i.e. don't try this transport again) via this hook. Currently, the Component code decides this.

spr0cketeer:
> It looks to me like the current code will do the first thing (notify on every connection failure).

Yes this is what it does currently.  I thought it best to do this and leave it up to the user to decide what to do about these events (if anything as adding a handler is optional).  If there's a network outage that slowly resolves itself over time, with the increasing backoff time on reconnects you may get different exceptions returned from this event over time with useful debug info.

> It might also make sense to allow user-code to decide if this counts as a "fatal error" (i.e. don't try this transport again) via this hook. Currently, the Component code decides this.

Could be useful.  Maybe allow adding a name/label key into the transports dict to identify each transport and have a `component.disable_transport(transport_name)` method?

autobahn/asyncio/component.py
```diff
@@ -49,19 +48,11 @@
...
 
 
-def _is_ssl_error(e):
-    """
-    Internal helper.
-    """
-    return isinstance(e, ssl.SSLError)
-
-
 def _unique_list(seq):
```
meejah: Is there a reason this was moved to be an instance-method instead of a bare helper function?

autobahn/asyncio/component.py
meejah: Bump; looks like `log_level=` is still in this diff?

autobahn/twisted/component.py
```diff
@@ -323,104 +317,23 @@ def _connect_transport(self, reactor, transport, session_factory):
         transport_endpoint = _create_transport_endpoint(reactor, transport.endpoint)
         return transport_endpoint.connect(transport_factory)
 
-    # XXX think: is it okay to use inlineCallbacks (in this
-    # twisted-only file) even though we're using txaio?
-    @inlineCallbacks
-    def start(self, reactor=None):
+    def start(self, reactor=None, log_level=None):
         """
```
meejah: remove `log_level=` kwarg

autobahn/wamp/component.py
```diff
@@ -460,11 +464,155 @@ def _can_reconnect(self):
                 return True
         return False
 
-    def start(self, reactor_or_loop):
-        raise RuntimeError('not implemented')
+    def _start(self, loop=None, log_level=None):
+        """
```
meejah: remove `log_level=` kwarg

spr0cketeer:
There's quite a few oddities with logging in Components ( and ApplicationRunner) stemming from txaio:
https://github.com/crossbario/txaio/issues/75
https://github.com/crossbario/autobahn-python/issues/674#issuecomment-225867970

The general advice for configuring logging in libraries is don't mess too much, just emit log events and let the app developer do their own logging configuration:
https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library
http://docs.python-guide.org/en/latest/writing/logging/#logging-in-a-library

Currently AB/txaio is messing too much with how stdlib logging works due to the choices made in txaio:

1.  `start_logging()` function attaches a handler to the root logger with default formatting and logging to stdout making app level logging configuration annoying e.g duplicate log events if you attach another handler to the root logger

2. in aio.py `make_logger()` always returns the root logger so all AB aio log events log to this object instead of an 'autobahn.*' namespace

3. info is the default loglevel even when not using `start_logging()` so `logging.debug()` is made a noop preventing `logger.setLevel(logging.DEBUG)` from working at app level.  This is contrary to what's stated in the last line of this section: https://autobahn.readthedocs.io/en/latest/wamp/programming.html#logging

I'll submit a couple of PRs to txaio repo to address 2 and 3.

Point 1 only applies if using `run([comp1, comp2], log_level='info'`) as it currently calls `start_logging()` whereas `comp1.start()` way of starting a component does not, but I think `run()` should be modified to remove the `log_level` argument and not use `start_logging()` either.  Yes it's a convenience method of running components but the AB lib shouldn't be starting logging too, that's an app level thing configured by the user via stdlib logging or tx's version.  

If run() is modified and the txaio PRs go in there would be no need to configure loglevels at the component level, both run() and start() users could then configure logging in the app like this:

```
logging.basicConfig(format=FORMAT, stream=sys.stdout)
myapp_logger = logging.getLogger(__name__)
myapp_logger.setLevel(logging.DEBUG)

ab_logger = logging.getLogger('autobahn')
ab_logger.setLevel(logging.INFO)
```
allowing independent logging levels of the app and AB.

Downside is no noop optimization of logleve calls in AB - how big of a performance issue is this?
Maybe we could create a module function like `optimize_logging(log_level=lvl)` which calls `txaio.set_global_log_level(lvl)` under the hood if needed.

autobahn/twisted/component.py
spr0cketeer: This was a workaround for point 3:
https://github.com/crossbario/autobahn-python/pull/914#issuecomment-349085777 but could be removed if the proposed changes to txaio are accepted.

autobahn/asyncio/component.py
spr0cketeer: So the tx/aio versions of _is_ssl_error() can be called by the component base class _start() method.

meejah: Ah, okay thanks that makes sense

autobahn/twisted/component.py
meejah: Hmmmmmmmm. That might take some more thinking. Because: allowing all `.debug()` calls to call all the way through to Python's logger if you've not got debug turned on will be (I believe) a fairly significant performance impact.

(Looking at this another way: if you *do* want debug messages to get through, why can't you call ``txaio.start_logging('debug')` or similar?)

The real point of *not* having `log_level=` kwarg in `start()` is because the rationale is: if you're calling `start()` yourself, you're "doing the logging stuff" (and managing the loop yourself). If you don't want to care about logging and loop-management, you'd use the high-level `run()` method of starting your components.

Also, consider the case where you have more than one `Component` -- what's supposed to happen if you pass different `log_level=` kwargs to each `start()` method? (With the code as it exists in this branch, whichever one got called most-recently will set the txaio global log level -- which I think is surprising and wrong).

meejah:
As to the question of "how much performance impact is `.debug()` not being a no-op?" I *think* the answer will be "significant" .. but I don't have hard figures for that. The reason I think this is that all internal logging is premised on the fact of it being no-op'd out if that log-level isn't turned on, and `.debug()` is not something you'd turn on in a production router for long periods of time.

(The amount of time "interpolating" strings -- that is, string-ifying and inserting variables into the format strings -- is non-trivial, and some of the debug logging is pretty low-level). That said, some hard figures would be *very nice* to have.

On the premise that most people will NOT want debugging messages in their logs, I think "by default it's off" is fine. If you *did* expect and want debug messages in the logs, you have to call `start_logging(...)` at some point .. but to use logging "properly" with Autobahn Python you have to do that anyway. Also, this WILL get called for you if you use the recommended, high-level `run()` API -- it only won't get called if you're opting in to calling `start()` on your components yourself (at which point, you should be arranging for `txaio.start_logging()` to be called precisely once anyway). Perhaps "this needs better documentation" is the real solution...?

spr0cketeer:
By attaching a handler to the root logger, start_logging() doesn't play nicely with stdlib logging.  Using set_global_level() achieves the same noop optimization without adding the handler.

If the performance hit to allowing debug calls through is significant, then how about:

https://github.com/crossbario/txaio/pull/122: delete, keeping info the default loglevel
https://github.com/crossbario/txaio/pull/121: merge to give 'autobahn' logging namespace 

`start()`:

- `log_level=` kwarg: remove, leaving log_level info the default in AB
- document method of changing loglevel to debug is to do both: 
  - `ab_logger.setLevel(logging.DEBUG)
  - `txaio.aio.set_global_log_level('debug')`

`run()`:

-` log_level=` kwarg: keep
- change `start_logging()` to `txaio.aio.set_global_level(log_level)`

That should ensure debug remains a noop and allow user to set app and AB level logging

oberstet:
I haven't read all above (and I wonder if the discussion is still related to the original issue), but I am against adding even more magic in the logging in txaio/AB/CB - this stuff is already too complicated.

Also, before trying to retrofit better interop with legacy logging (stdlib), IMO we should first reap the benefits of our advanced structured logging. 

That being said, I guess there are valid itches with the way logging is done in AB/CB. However, could we at least please move out the logging related discussion and proposals for change from the original issue here (which I understand was about uncontroversial, smaller improvements to the new component API)?

That way, we could merge the (uncontroversial) bits .. moving forward. And then first file a new issue that clearly describes the itch with logging first, and then discuss from there ..

> Rgd NOPs: this is important (for Crossbar.io), we will keep that. Any change that doesn't make log functions NOPs (depending on log level) isn't going to fly .. AB/CB has tons of `log.debug` calls deep inside at very granular level

spr0cketeer:
@oberstet yeah the discussion has spiraled a bit from the original PR and onto logging issues, but as it's where the rubber meets the road for end users of AB it's important to get the Component stuff right and not repeat any problem areas ApplicationRunner has.  I think it's worth continuing just a little more as it's feeling pretty close to being thrashed out (I hope!)

Current state:

log.debug calls will remain NOPs by default

The only change to txaio logging is to add namespace detection for aio side calls to make_logger():
 https://github.com/crossbario/txaio/pull/121

This will bring AB aio logging namespace in line with the existing code on the tx side https://github.com/crossbario/txaio/blob/master/txaio/tx.py#L242

Both ways of running/starting components now don't specify any log_level as logging configuration is driven app side via stdlib logging/tx logger modules.  The one caveat being the need to call `txaio.set_global_log_level('debug')` if debug logging is wanted.

@meejah to me the logging configuration is feeling much saner now and allows flexible ways of handling it app side.  I guess it comes down to whether you're happy that run() now won't start any logging and so means a couple more lines of logging code along with Component setup, but Component example code can take care of that (on aio side at least, WARNING levels and above are sent to stderr even if the user forgets to do any logging configuration).

spr0cketeer:
I've removed the changes to how logging works and will do a new PR for that - this PR now covers its original intent stated in the title.

spr0cketeer:
@meejah tweaked to add in your changes from 6bcca8e.  Can you do a review pls, be good to get this in before any other changes come in.

spr0cketeer:
@meejah ping

meejah:
Sorry, I just noticed your updates now! I will re-test/review tomorrow. Thanks!

Issue #875, opened 2017-07-26
add unregisterProducer to complement registerProducer


Dobatymo:
`WebSocketAdapterProtocol` implements `registerProducer`. It should contain `unregisterProducer` also. It just has to be passed through to twisted.

autobahn/twisted/websocket.py
```python
def unregisterProducer(self):
    self.transport.unregisterProducer()
```

autobahn/asyncio/websocket.py
```python
def unregisterProducer(self):
    raise Exception("not implemented")
```

PR #924, opened 2017-11-01
passthrough unregisterProducer to complement registerProducer

Dobatymo:
fixes https://github.com/crossbario/autobahn-python/issues/875

simply wrap `unregisterProducer` in the same style as `registerProducer`.


oberstet:
thanks for contributing!

PR #928, opened 2017-11-10
started adding WAMP message muxing

oberstet:


autobahn/wamp/serializer.py
```diff
@@ -262,16 +288,24 @@ class JsonSerializer(Serializer):
...
-            self.SERIALIZER_ID = u"json.batched"
+            if muxed:
+                self.SERIALIZER_ID = u"json.muxed.batched"
+            else:
+                self.SERIALIZER_ID = u"json.batched"
+        else:
+            if muxed:
+                self.SERIALIZER_ID = u"json.muxed"
+            else:
+                self.SERIALIZER_ID = u"json"
```
meejah: All these "if" statements (for "batched" vs. "muxed") make me think there should be more classes instead, perhaps ... like a `JsonBatchedSerializer` and `JsonMuxedSerializer`, etc.

Issue #907, opened 2017-09-23
component API reliance on twisted


singlecheeze:
See #295 If twisted is not loaded/wanted, opt of asyncio instead, autobahn-python/autobahn/wamp/component.py relies on two twisted imports (is this intentional?):
```python
from twisted.python.failure import Failure
from twisted.internet.error import ReactorNotRunning
```
```python
    def all_done(arg):
        log.debug("All components ended; stopping reactor")
        if isinstance(arg, Failure):
            log.error("Something went wrong: {log_failure}", failure=arg)
        try:
            reactor.stop()
        except ReactorNotRunning:
            pass
```
```text
Traceback (most recent call last):
  File "/home/user/PycharmProjects/app/backend/main.py", line 4, in <module>
    from autobahn.asyncio.component import Component, run
  File "/usr/lib/python3.6/site-packages/autobahn/asyncio/component.py", line 49, in <module>
    from autobahn.wamp import component
  File "/usr/lib/python3.6/site-packages/autobahn/wamp/component.py", line 34, in <module>
    from twisted.python.failure import Failure
ModuleNotFoundError: No module named 'twisted.python'
```

meejah:
Hmm, thanks -- yess any asyncio/twisted specific code should definitely only exist in e.g. `autobahn.twisted.component` (or the asyncio equivalent).

om26er:
I was just hit by this while I was updating one of my apps from ApplicationSession to Component.

om26er:
fixed by https://github.com/crossbario/autobahn-python/pull/930

PR #930, opened 2017-11-21
Remove twisted dependency from component API

om26er:
Fix for https://github.com/crossbario/autobahn-python/issues/907

### DIFF
--- a/autobahn/asyncio/component.py
+++ b/autobahn/asyncio/component.py
@@ -469,8 +469,11 @@ def run(components, log_level='info'):
     loop.add_signal_handler(signal.SIGINT, partial(nicely_exit, 'SIGINT'))
     loop.add_signal_handler(signal.SIGTERM, partial(nicely_exit, 'SIGTERM'))
 
+    def done_callback(reactor, arg):
+        reactor.stop()
+
     # returns a future; could run_until_complete() but see below
-    component._run(loop, components)
+    component._run(loop, components, done_callback)
 
     try:
         loop.run_forever()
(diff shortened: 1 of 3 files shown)


meejah:
This looks fine as-is .. but I wonder if we could move the "Twisted-specific" code to `autobahn.twisted.component` instead, somehow?

om26er:
> This looks fine as-is .. but I wonder if we could move the "Twisted-specific" code to autobahn.twisted.component instead, somehow?

Now made `done_callback` as a parameter which both backends must provide, kindly take a look again.

autobahn/asyncio/component.py
```diff
@@ -469,8 +469,11 @@ def nicely_exit(signal):
     loop.add_signal_handler(signal.SIGINT, partial(nicely_exit, 'SIGINT'))
     loop.add_signal_handler(signal.SIGTERM, partial(nicely_exit, 'SIGTERM'))
 
+    def done_callback(reactor, arg):
+        reactor.stop()
+
     # returns a future; could run_until_complete() but see below
```
meejah: very minor nit-pick (and I'm not even sure if we're consistent here already) but asyncio usually calls the reactor "loop"

meejah:
Great! This looks nicer to me; keeping any Twisted/asyncio-specific imports in `asyncio.twisted.*` is much better than inside `if` constructs :)

meejah:
(If you want to squash it all and force-push, go ahead .. otherwise good to merge from my perspective).

autobahn/asyncio/component.py
om26er: just pushed another commit that changes that.

PR #934, opened 2017-12-07
Component errors, component auth

meejah:
This fixes some error-handling on component authentication and adds "anonymous" and "ticket" authentication methods.


meejah:
Hmm, I have no idea why the py3.3 builders are sad

PR #935, opened 2017-12-08
Component-first documentation and separate listings files

meejah:


oberstet:
`pytest requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*' but the running Python is 3.3.6`

pytest has given up Py 3.3 ..

PR #940, opened 2018-01-25
Don't install default reactor unnecessarily

dvhwgumby:
On MacOS (at least) crossbar won't run because it tries to load a default reactor although it is already loaded.
This two-line fix to choosereactor.py simply checks to see if the module has been loaded before attempting to load it.
Possibly this check should be in twister/internet/main.py but that is marked as an obsolete compatibility method anyway so I put the fix here.

This addresses these corssbar issues:
crossbario/crossbar#1200
crossbario/crossbar#939
crossbario/crossbar#920
crossbario/crossbar#929

### DIFF
--- a/autobahn/twisted/choosereactor.py
+++ b/autobahn/twisted/choosereactor.py
@@ -65,7 +65,8 @@ def install_optimal_reactor(verbose=False):
         if current_reactor != 'KQueueReactor':
             try:
                 from twisted.internet import kqreactor
-                kqreactor.install()
+                if 'twisted.internet.reactor' not in sys.modules:
+                    kqreactor.install()
             except:
                 log.critical("Running on *BSD or MacOSX, but cannot install kqueue Twisted reactor")
                 log.warn("{tb}", tb=traceback.format_exc())


oberstet:
currently on mobile, keeping it short: first, welcome to crossbar/aurobahn:) and thanks for caring, andnfor contributing!rgd the specific patch you propose: pls see my comment on the crossbar issue: would be much preferred to actually find the twisted pkg import in crossbar that triggers that, and then postpone the import so that chosereactor can actually install the kqueue reactor..Am 25.01.2018 6:20 nachm. schrieb DV Henkel-Wallace <notifications@github.com>:On MacOS (at least) crossbar won't run because it tries to load a default reactor although it is already loaded.
This two-line fix to choosereactor.py simply checks to see if the module has been loaded before attempting to load it.
Possibly this check should be in twister/internet/main.py but that is marked as an obsolete compatibility method anyway so I put the fix here.
This addresses these corssbar issues:
crossbario/crossbar#1200
crossbario/crossbar#939
crossbario/crossbar#920
crossbario/crossbar#929



You can view, comment on, or merge this pull request online at:
  https://github.com/crossbario/autobahn-python/pull/940

Commit Summary
Don't install default reactor unnecessarily

File Changes

    M
    autobahn/twisted/choosereactor.py
    (3)
  

Patch Links:
https://github.com/crossbario/autobahn-python/pull/940.patchhttps://github.com/crossbario/autobahn-python/pull/940.diff

dvhwgumby:
Hallo Tobias, freut mich!  Well, in any case the installation instructions don't work on current Mac OS, which is a an open issue however you choose to address it it.  

FWIW, some of my colleagues are using my (admittedly crude) work around, and some others have chosen to use asyncio instead of twisted.  So a quick fix might be to change the default installation instructions on crossbar.io to suggest asyncio.

(I'm using C++ so will put my other questions there.  Enough Python for today 😀).

PR #946, opened 2018-02-07
twisted trunk doesn't work on python3.3

meejah:
See https://github.com/twisted/twisted/pull/943 Twisted dropped support for Python 3.3

This PR just removes it from the Tox matrix so builds pass. Can we add some magic in setup.py that a) means a python3.3 gets Twisted no newer than 17.9.x but b) others get newest Twisted? Or, crossbar drops python3.3 support as well (which then needs doc changes).


oberstet:
that was quick;)

PR #956, opened 2018-03-09
Wamp scram prototype

meejah:
See https://github.com/wamp-proto/wamp-proto/issues/135

### DIFF
--- a/autobahn/wamp/auth.py
+++ b/autobahn/wamp/auth.py
@@ -245,14 +245,15 @@ class AuthScram(object):
         # '$argon2i$v=19$m=512,t=2,p=2$5VtWOO3cGWYQHEMaYGbsfQ$AcmqasQgW/wI6wAHAMk4aQ'
 
         _, tag, ver, options, salt_data, hash_data = rawhash.split(b'$')
-        salted_password = hash_data  # KDF(Normalize(password), salt, params...)
+        salted_password = hash_data
         client_key = hmac.new(salted_password, b"Client Key", hashlib.sha256).digest()
         stored_key = hashlib.new('sha256', client_key).digest()
 
         client_signature = hmac.new(stored_key, auth_message.encode('ascii'), hashlib.sha256).digest()
-        client_proof = base64.b64encode(xor(client_key, client_signature))
+        client_proof = xor(client_key, client_signature)
+        print(client_proof)
 
-        return client_proof
+        return base64.b64encode(client_proof)
 
 
 IAuthenticator.register(AuthScram)
--- a/autobahn/wamp/auth.py
+++ b/autobahn/wamp/auth.py
@@ -36,13 +36,18 @@ import hmac
 import hashlib
 import random
 from struct import Struct
-from operator import xor
-from itertools import starmap
-from autobahn.util import public
+from passlib.utils import saslprep
+
+from autobahn.util import public, xor
 from autobahn.wamp.interfaces import IAuthenticator
+
+# make optional dependency?
+from argon2.low_level import hash_secret
+from argon2 import Type
 
 
 __all__ = (
+    'AuthScram',
     'AuthCryptoSign',
     'AuthWampCra',
     'pbkdf2',
@@ -65,8 +70,9 @@ def create_authenticator(name, **kwargs):
     """
     try:
         klass = {
-            AuthWampCra.name: AuthWampCra,
+            AuthScram.name: AuthScram,
             AuthCryptoSign.name: AuthCryptoSign,
+            AuthWampCra.name: AuthWampCra,
             AuthAnonymous.name: AuthAnonymous,
             AuthTicket.name: AuthTicket,
         }[name]
@@ -169,6 +175,87 @@ class AuthCryptoSign(object):
 
 
 IAuthenticator.register(AuthCryptoSign)
+
+
+class AuthScram(object):
+    """
+    Implements "wamp-scram" authentication for components.
+
+    NOTE: This is a prototype of a draft spec; see
+    https://github.com/wamp-proto/wamp-proto/issues/135
+    """
+    name = u'scram'
+
+    def __init__(self, **kw):
+        self._args = kw
+        self._client_nonce = None
+
+    @property
+    def authextra(self):
+        # is authextra() called exactly once per authentication?
+        if self._client_nonce is None:
+            self._client_nonce = base64.b64encode(os.urandom(16)).decode('ascii')
+        return {
+            u"nonce": self._client_nonce,
+        }
+
+    def on_challenge(self, session, challenge):
+        assert challenge.method == u"scram"
+        assert self._client_nonce is not None
+        required_args = ['nonce', 'salt', 'cost']
+        optional_args = ['memory', 'parallel']
+        # probably want "algorithm" too, with either "argon2id-19" or
+        # "pbkdf2" as values
+        for k in required_args:
+            if k not in challenge.extra:
+                raise RuntimeError(
+                    "WAMP-SCRAM challenge option '{}' is "
+                    " required but not specified".format(k)
+                )
+
+        channel_binding = ''  # fixme
+        server_nonce = challenge.extra[u'nonce']  # base64
+        salt = challenge.extra[u'salt']  # base64
+        cost = int(challenge.extra[u'cost'])
+        memory = int(challenge.extra.get(u'memory', 512))
+        parallel = int(challenge.extra.get(u'parallel', 2))
+        password = self._args['password'].encode('utf8')  # supplied by user
+        authid = saslprep(self._args['authid'])
+        client_nonce = self._client_nonce
+
+        auth_message = (
+            "{client_first_bare},{server_first},{client_final_no_proof}".format(
+                client_first_bare="n={},r={}".format(authid, client_nonce),
+                server_first="r={},s={},i={}".format(server_nonce, salt, cost),
+                client_final_no_proof="c={},r={}".format(channel_binding, server_nonce),
+            )
+        )
+
+        rawhash = hash_secret(
+            secret=password,
+            salt=base64.b64decode(salt),
+            time_cost=cost,
+            memory_cost=memory,
+            parallelism=parallel,
+            hash_len=16,  # another knob?
+            type=Type.ID,
+            version=19,
+        )
+        # spits out stuff like:
+        # '$argon2i$v=19$m=512,t=2,p=2$5VtWOO3cGWYQHEMaYGbsfQ$AcmqasQgW/wI6wAHAMk4aQ'
+
+        _, tag, ver, options, salt_data, hash_data = rawhash.split(b'$')
+        salted_password = hash_data  # KDF(Normalize(password), salt, params...)
+        client_key = hmac.new(salted_password, b"Client Key").digest()
+        stored_key = hashlib.new('sha256', client_key).digest()
+
+        client_signature = hmac.new(stored_key, auth_message.encode('ascii')).digest()
+        client_proof = base64.b64encode(xor(client_key, client_signature))
+
+        return client_proof
+
+
+IAuthenticator.register(AuthScram)
 
 
 class AuthWampCra(object):
--- a/autobahn/wamp/auth.py
+++ b/autobahn/wamp/auth.py
@@ -36,14 +36,22 @@ import hmac
 import hashlib
 import random
 from struct import Struct
-from passlib.utils import saslprep
+from operator import xor
+from itertools import starmap
 
-from autobahn.util import public, xor
+from autobahn.util import public
+from autobahn.util import xor as xor_array
 from autobahn.wamp.interfaces import IAuthenticator
 
-# make optional dependency?
-from argon2.low_level import hash_secret
-from argon2 import Type
+# if we don't have argon2/passlib (see "authentication" extra) then
+# you don't get AuthScram and variants
+try:
+    from argon2.low_level import hash_secret
+    from argon2 import Type
+    from passlib.utils import saslprep
+    HAS_ARGON = True
+except ImportError:
+    HAS_ARGON = False
 
 
 __all__ = (
@@ -187,6 +195,11 @@ class AuthScram(object):
     name = u'scram'
 
     def __init__(self, **kw):
+        if not HAS_ARGON:
+            raise RuntimeError(
+                "Cannot support WAMP-SCRAM without argon2_cffi and "
+                "passlib libraries; install autobahn['scram']"
+            )
         self._args = kw
         self._client_nonce = None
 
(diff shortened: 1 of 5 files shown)


oberstet:
Awesome! This looks quite good already. I am +1 on merging this soonish .. when the CI gets green;) So we can test it already early on and fix design gaps/issues as we eat our dog food;)

meejah:
Okay, what it needs:

 - some refactoring and unit-tests on the basic argon/scram stuff (and also re-use some of it in crossbar impl)
 - doesn't support PBKDF2 at all
 - doesn't do anything at all about channel-binding (but, this maybe needs more spec work too?)
 - do we want channel-binding to be optional? (I guess it has to be, "because JS", right?)

Things to note (but I think these are fine):

 - all the config values (for crossbar and client end) use hex-encoded binary (but most of the actual protocol-traffic encodes blobs using base64). Hex is more-consistent with existing crossbar/client things.

What needs to be thought about:

 - I'm using "authextra" for everything (which is NOT what the spec says); do we change the code or the spec? authextra is "kind-of more for user-level extra stuff" so are we abusing this by using it like this?
 - do we want a more-explicit "post auth" tie-in for all authentication types? (Right now, I'm adding my own on-join handler that double-checks the server signature -- this works fine, but it's possible some other on-join handler will run first and you'll do some "real work" before the server-signature is checked)
 - `argon2id`parameters and knobs: one "cost" thing for both pbkdf2 + argon2i, or expose everything
 - is `wamp-scram` plus an `algorithm` identifier (of either `argon2id-19` or `pbkdf2`) what we want to do? (same as my suggestion on the wamp-proto bug)
 - we're not quite following the SCRAM RFC here: "real" SCRAM says SHA1 *always* has to be supported (for example).
 - if we're "abusing" the RFC and using SHA256 anyway (which I think is fine) then maybe we want to consider using BLAKE2b instead? (Better/faster than SHA-3, has built-in support in Python3, not sure about Python2). This would be for all the HMAC calls and the SHA. Probably should discuss in the "wamp-proto" issue instead

meejah:
Oh, library considerations: I used `passlib` for its `saslprep` util to process usernames according to the RFC (because in my venvs it was already installed) but it appears we don't currently depend on that. Do we want to?

`argon2_cffi` looks like "the" good argon2 library. Do we want to blindly/always depend on it, or do more like how we handle NaCl (import the argon2i library only if something tries to do SCRAM and fail then if its an import error)?

oberstet:
@meejah rgd the new dependencies: I assume these are MIT or sth? no problem then, I guess if we want to stick to our pretty fine grained approach of "extras" (https://github.com/crossbario/autobahn-python/blob/master/setup.py#L192), we could either put the deps into "encryption", or into a new extra "authentication"?

oberstet:
> do we want channel-binding to be optional? (I guess it has to be, "because JS", right?)

yeah, at least JS from browsers cannot do that. no API. which is bad, because it means that any non-native authentication scheme cannot protect against the attacks that can be mounted when the authentication handshaking isn't tied to the underlying transport encryption .. I guess. maybe there is a different way of mitigating the same set of attacks, dunno.

oberstet:
> is wamp-scram plus an algorithm identifier (of either argon2id-19 or pbkdf2) what we want to do? (same as my suggestion on the wamp-proto bug)

if we can get away with 3 parts separated with 2 x "-", that is no "-" in the parts themselves, the 3rd part can totally be the algorithm. this allows the client to express a preference (by supplying multiple wamp-scram schemes, with different algos, in order), as well as let the server chose from that list (under server preference). as the client will get challenged with the full scheme chosen by the server, this allows both side to effectively negotiate for the algo to be used under wamp-scram as well. sounds good!

oberstet:
> we're not quite following the SCRAM RFC here: "real" SCRAM says SHA1 always has to be supported (for example).

so they really expect implementors to _must_ support deprecated hashes? mmh. as this scheme (wamp-scram) is used by wamp implementations, we can very well tighten that up, I'd say. so we just _require_ the hash to be used for wamp-scram to be from a tigher set.

meejah:
argon2_cffi is MIT, passlib is BSD

oberstet:
> I'm using "authextra" for everything (which is NOT what the spec says); do we change the code or the spec? authextra is "kind-of more for user-level extra stuff" so are we abusing this by using it like this?

yeah, this is abusing it;) it is meant for user stuff. mainly, to forward arbitrary data to a custom dynamic authenticator at the router side. so actually, these wamp-scram specific things should live on attributes of classes.

> do we want a more-explicit "post auth" tie-in for all authentication types? (Right now, I'm adding my own on-join handler that double-checks the server signature -- this works fine, but it's possible some other on-join handler will run first and you'll do some "real work" before the server-signature is checked)

not sure, can't follow. the checking should be done (and completed) in the auth class hooked into the WAMP handshake (called from protocol/session). the full checking should be definitely done before the on-join handlers are even fired - just to be double safe .. yeah, might be that making sure the final checking is in the first added handler, but huh, a tricky side condition to make sure. but maybe I don't get what you point at here ..

meejah:
The RFC says you can choose what hash to use for the HMAC, but "for interoperability" everything must support SHA1. I think in this case, it's fine to just not do that? (i.e. SHA256 only)

oberstet:
I should actually read the RFC .. long ago;) but in any case: these are already quite details, and the prototype looks already good! I think this is "working code" in the IETF sense. actually, half: the other bits obviously in crossbar.

meejah:
To be clear too, this has just "wamp-scram-argon2id19" or whatever (depending on what colour the "name" bikeshed is painted) in this PR, not "wamp-scram-pbkdf2"

oberstet:
> wamp-scram-argon2id19

sounds impressive;) like a borg drone ... merging!

oberstet:
wondering .. besides the use of SHA256, is there anything we should take back to the spec discussion https://github.com/wamp-proto/wamp-proto/issues/135 ?

meejah:
I think the current draft-spec used SHA256 as well. We should decide on "wamp-scram-*" vs. "wamp-scram" and an "algorithm" param.

There's the "authextra" vs. not thing, but I'm currently making it work with explicit attributes in `message.Welcome` etc instead of authextra (so we could be clear on that in the 135 discussion).

The actual SCRAM RFC announces support for channel-binding via the algorithm name, too, by adding `-plus` on the end.  So probably we should say "all WAMP SCRAM authentication methods start with `wamp-scram-` and the rest of the string is the algorithm name"

meejah:
If we do the above, we'd then have four SCRAM-type authentication methods:

 - wamp-scram-argon2id19
 - wamp-scram-argon2id19-plus
 - wamp-scram-pbkdf2
 - wamp-scram-pbkdf2-plus

ecorm:
I'm happy to see an early prototype, but I hope this doesn't blunt our efforts to reach consensus for the formal WAMP-SCRAM specification. Having said that, whatever practical issues you're come up against while implementing this early prototype would certainly be worth learning about.

@meejah 
> - wamp-scram-argon2id19
> - wamp-scram-argon2id19-plus
> - wamp-scram-pbkdf2
> - wamp-scram-pbkdf2-plus

The "-plus" suffix thing was a carry-over from the SCRAM RFC, and I don't think it's really needed for WAMP-SCRAM. Channel binding is something that has to be requested by the client, and there's already a `gs2_cbind_flag` property that indicates such a request (with `gs2_cbind_flag` to be possibly renamed to `channel_binding`.

Also, FYI, we're leaning towards "SRP" over "SCRAM" for our project because it's more secure against evesdropping. I was going to write an SRP proposal soon. A lot of the things we're working out for the WAMP-SCRAM spec can be used for a WAMP-SRC spec.

ecorm:
> The RFC says you can choose what hash to use for the HMAC, but "for interoperability" everything must support SHA1. I think in this case, it's fine to just not do that (i.e. SHA256 only) because we're not really writing "a SCRAM server that compiles with RFC5802"..?

RFC5802 is has the annotation "Updated by RFC7677", where RFC7677 defines the SCRAM mechanism that uses SHA-256 instead of the weaker SHA-1.

In RFC7677 section 4, we find:

> See [RFC4270] and [RFC6194] for reasons to move from SHA-1 to a strong security mechanism like SHA-256.

So they are indeed encouraging to move away from SHA-1.

SCRAM and other SASL authentication mechanisms are used by email and XMPP servers, so it makes sense for them to impose SHA-1 as the "least common denominator" for interoperability with email/chat clients. But that's not the situation for WAMP, so I think it's fine that we enforce the stronger SHA-256 hash.

If we add `SHA-1` as a possibility in our spec, we're going to get blasted by the security conscious folks who will say that `SHA-1` is no longer considered that secure.

> Oh, library considerations: I used passlib for its saslprep util to process usernames according to the RFC (because in my venvs it was already installed) but it appears we don't currently depend on that. Do we want to?

It's the availability of saslprep on browser-side Javascript that worries me more. I managed to find the [reklatsmasters/saslprep](https://github.com/reklatsmasters/saslprep) Javascript library, and the author [confirmed that it works on modern browsers](https://github.com/reklatsmasters/saslprep/issues/1).

We could loosen the normalization requirement to be only NFKC, but that would deviate from the SCRAM RFC.

oberstet:
> but I'm currently making it work with explicit attributes in message.Welcome etc instead of authextra

mmh, sorry @meejah, maybe I was hinting wrong here. rethinking:

we can't change the structure of the WAMP message, that is https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/message.py#L1057

we also cannot change (without breaking) the signature of `onChallenge` callback in the Autobahn session API https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/protocol.py#L1122

currently, the one parameter to `onChallenge` is just the raw dict I think, whereas it probably should have been a proper object, eg of class `ScramChallenge`, `CryptosignChallenge`. maybe.

but for now I think it would be best to follow wamp-cryptosign https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/auth.py#L140

looking over that code, I am wondering about https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/auth.py#L294

I think this is what you asked about? this is somewhat unexpected to me here, the cryptosign code just returns directly https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/auth.py#L182, not in the on-join event that follows. 

Actually, shouldn't the authenticator classes also have a onWelcome hook? And only that is then supposed to check the "scram_server_signature", and then trigger the whole on-join chain etc.?

oberstet:
In a way, there are corresponding classes in Crossbar.io that also have two hooks (onHello, onAuthenticate) .. and the client authenticator classes here could have two hooks too (onChallenge, onWelcome). Essentially, instances of these two classes are then called during the authentication handshake - over the wire.

oberstet:
I can't remember, but reasons? SCRAM seemed like popular on the street (and wide use is good for shaking out issues) https://news.ycombinator.com/item?id=10339676.

here is a nice overview https://www.certsi.es/en/blog/password-based-authentication

However, then there are people that say SPAKE2+EE is more secure than SRP;) https://www.reddit.com/r/netsec/comments/7xhxd0/the_scram_authentication_protocol/

Or AugPAKE https://tools.ietf.org/html/rfc6628. or ..

and then why be confined to this narrow class of schemes where poor users are required to remember ideally random numbers to help a machine do its work securely? ;)

ecorm:
> I can't remember, but reasons? SCRAM seemed like popular on the street (and wide use is good for shaking out issues .. eg SCRAM is a standard scheme used in SASL) https://news.ycombinator.com/item?id=10339676.

I'll put more details in https://github.com/wamp-proto/wamp-proto/issues/296 when I have the chance, but I think the main reason why SRP was avoided before was due to patent issues. This is no longer the case, as the relevant patents have expired.

meejah:
> Actually, shouldn't the authenticator classes also have a onWelcome hook? And only that is then supposed to check the "scram_server_signature", and then trigger the whole on-join chain etc.?

Yes, this is exactly what I was getting at -- I was thinking of adding a hook like above but hacked it in via "join" instead "because it's just a prototype" and we had that already. No other auth mechanism so far has something like this -- but I totally agree that "possibly running any other on-join handlers before we've confirmed the server" is bad.

@ecorm I find it easiest/best for myself to write prototype/PoC code while analyzing a spec, so .. don't worry I'm happy to make any changes to the code if we change the spec (in any case, the code doesn't actually exactly follow the spec, so .. ;). Also, I definitely have zero desire to use SHA1 for anything so I'm glad you read the RFCs more closely than I did :)

meejah:
@oberstet okay, I'm confused about `authextra`: we currently do use it for e.g. wamp-cryptosign to send a pubkey it seems -- was that also "abusing" the spec?

Looking at *just* the "client_nonce" for wamp-scram, there's two options: add this to `message.py` in the `Hello` class (like, e.g., `resume_token`) -- this doesn't change the WAMP message, as all those things are looked up in Details anyway **or** just use `authextra`.

I think what you mean by "don't use authextra" is that the client_nonce should go into the Hello as a top-level attribute (i.e. added to `__slots__`, etc.). This is also how the current draft spec from ecorm is written, too (i.e. the extra auth things go into Details).

oberstet:
@meejah sorry for confusion! indeed, after relooking, I think we should discuss this more deeply and take the chance to come up with a sustainable solution.

So on the wire, a HELLO message looks like this:

```
[HELLO, Realm|uri, Details|dict]
```

The interesting stuff is in the Details, eg here is a client offering two authmethods:

```
[HELLO, "realm1", {
   "authid": "alice";
   "authrole": "user",
   "authmethods": ["foo", "bar"],
   "authextra": {
      "param1": 23,
      "param2": 666
   }
}]
```

All existing autmethods pack their knobs into `authextra` (which must be a dict - if present) - eg WAMP-cryptosign is looking for a `pubkey` attribute in `authextra`.

Following that approach, WAMP-SCRAM knobs that a client wants to announce to a router should also go as attributes into `authextra` dict in the `details` dict of the HELLO message.

---

Now, looking at the structure of the `details` in above example, it seems actually weird that there is an option to have a _list_ of `authmethods`, but only _one_ `authextra` for all these.

What if I wanted to announce that I am (as a client) willing to authenticate under 2 specific parameter sets of WAMP-SCRAM - and none else?

This isn't possible in the current message structure, as the knobs are shared for the full list of authmethods.

What we could do is allow `authmethods` to take a more complex structure, eg

```
[HELLO, "realm1", {
   "authmethods": [
      {
         "authmethod": "foo",
         "authid": "alice",
         "authrole": "user",
         "authextra": {
            "param1": 23,
            "param2": 666
         }
      },
      {
         "authmethod": "bar",
         "authid": "bob",
         "authrole": "guest",
         "authextra": {
            "param1": 42,
            "param3": "foo"
         }
      }
   ]
}]
```

The semantics could be: if there is a knob at the authmethod level, take that, if not take the knob at the (global) authextra, if none, take default.

oberstet:
@meejah pls see https://github.com/wamp-proto/wamp-proto/issues/301 for what I think could be a sustainable structure for authextra/authmethods

meejah:
@oberstet 
> Actually, shouldn't the authenticator classes also have a onWelcome hook? And only that is then supposed to check the "scram_server_signature", and then trigger the whole on-join chain etc.?

So this would add a method to `ISession` like `onWelcome` (i.e. similar to `onChallenge`) and the Autobahn default implementation for `Session` would call out to `authenticator.on_welcome` or so...with I guess the argument being some kind of `Welcomed` (or similar?) object (like `Challenge`) so that we can add things in the future without wrecking the function signature? For now I think all it would need is "authmethod" and "authextra"...

PR #958, opened 2018-03-16
wamp-scram

meejah:
Continuation of https://github.com/crossbario/autobahn-python/pull/956

This adds an onWelcome/on_welcome hook for authenticators to use and SCRAM uses it to check the server-signature as well as a few other fixups. Now matches the draft-spec (again).

### DIFF
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -287,6 +287,9 @@ class Session(protocol._SessionShim):
     # XXX these methods are redundant, but put here for possibly
     # better clarity; maybe a bad idea.
 
+    def on_welcome(self, welcome_msg):
+        pass
+
     def on_join(self, details):
         pass
 
--- a/autobahn/twisted/component.py
+++ b/autobahn/twisted/component.py
@@ -41,8 +41,8 @@ try:
     from OpenSSL import SSL
 except ImportError as e:
     _TLS = False
-    if 'OpenSSL' not in str(e):
-        raise
+    # there's no optionsForClientTLS in older Twisteds or we might be
+    # missing OpenSSL entirely.
 
 import six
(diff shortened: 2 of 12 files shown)


oberstet:
looks good! has the new onWelcome stuff.

rgd the fallout from that: maybe we could add a default (empty) implementation of onWelcome here https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/protocol.py#L1129

oberstet:
@meejah I would like to land the current state of WAMP-SCRAM in both AB and CB - do you think it is "ready"? There is some CI fallout that needs fixing, but then I would merge. Could you give me ping when to do a final review for merge? Or just assign me as reviewer when you feel ready ..

meejah:
Okay I added the empty `onWelcome` and made a few fixups for the tests etc. Should be good to merge @oberstet

oberstet:
perfect! great work. this is a very complete "prototype" I'd say;) now, cool, we have a WAMP-SCRAM spec and client/router prototype=)

ecorm:
@meejah, @oberstet, except for channel binding, is WAMP-SCRAM now fully operational in Crossbar? Does it support dynamic authenticators?

ecorm:
Oh, and thank you for implementing this! :heart:

meejah:
It doesn't support dynamic authenticators (and no channel-binding, that's correct). Adding dynamic authenticators wouldn't be much more additional work but is not in master right now..

ecorm:
@meejah For internal use, we can do ticket auth with client-side pre-hashing for now. We still have a lot of work to do for user management and permission control. We'd like to "drop in" WAMP-SCRAM later so that we gain the property of an eavesdropper not being able to impersonate a client.

Issue #971, opened 2018-03-28
Asyncio: ApplicationRunner doesn't handle correctly restart of the event loop


rdebroiz:
Running a `asyncio.wamp.ApplicationSession` with `ApplicationRunner.run(session start_loop=True)`
two consecutive times fails the second one.

Minimal example: https://gist.github.com/rdebroiz/1854c2bab7da929e376c273d262c805a

Adding
``` python
if loop.is_closed():
    asyncio.set_event_loop(asyncio.new_event_loop())
```
Before to run the ApplicationRunner the second times make it works as expected.

This code snippet is coming from and is already called by `asyncio.wamp.ApplicationRunner.run`
see: https://github.com/crossbario/autobahn-python/blob/master/autobahn/asyncio/wamp.py#L246

Any ideas of what is going on?

om26er:
I can definitely reproduce the issue and after some initial debugging, I am a little confused as we are definitely checking if the loop is closed and setting a new one.

That's most probably a race condition, needs more digging.

om26er:
Commenting this out seems to fix the issue https://github.com/crossbario/autobahn-python/blob/master/autobahn/asyncio/wamp.py#L283

rdebroiz:
Just for the record, I got the same behavior with components, except that refresh the event loop between the two calls to run does not fix it.
```python
#!/usr/bin/env python
# usage: python autobahn_start_AS_two_times.py host port realm
import sys
from autobahn.asyncio.component import Component, run

import txaio
txaio.use_asyncio()

if __name__ == '__main__':
    url = f"ws://{sys.argv[1]}:{sys.argv[2]}/ws"
    realm = sys.argv[3]

    def main(session, details):
        pass

    component = Component(main=main, transports=url, realm=realm)

    run(component, 'debug')
    # let's do it again
    run(component, 'debug')
```
Output
``` python
  File "trash.py", line 19, in <module>
    run(component, 'debug')
  File "/local/ch196713/virtualenvs/****/lib/python3.6/site-packages/autobahn/asyncio/component.py", line 355, in run
    loop.add_signal_handler(signal.SIGINT, partial(nicely_exit, 'SIGINT'))
  File "/usr/lib64/python3.6/asyncio/unix_events.py", line 85, in add_signal_handler
    self._check_closed()
  File "/usr/lib64/python3.6/asyncio/base_events.py", line 357, in _check_closed
    raise RuntimeError('Event loop is closed')
[...]

PR #979, opened 2018-04-03
AIO/Component: create a new loop if already closed

om26er:
Fixes https://github.com/crossbario/autobahn-python/issues/971

### DIFF
--- a/autobahn/asyncio/component.py
+++ b/autobahn/asyncio/component.py
@@ -333,6 +333,10 @@ def run(components, log_level='info'):
     if log_level is not None:
         txaio.start_logging(level=log_level)
     loop = asyncio.get_event_loop()
+    if loop.is_closed():
+        asyncio.set_event_loop(asyncio.new_event_loop())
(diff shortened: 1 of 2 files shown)


meejah:
This might work (i.e. code looks plausible) but I'm concerned about batched-timers still -- how will those get to the new event-loop?

autobahn/asyncio/component.py
```diff
@@ -333,6 +333,10 @@ def run(components, log_level='info'):
     if log_level is not None:
         txaio.start_logging(level=log_level)
     loop = asyncio.get_event_loop()
+    if loop.is_closed():
+        asyncio.set_event_loop(asyncio.new_event_loop())
+        loop = asyncio.get_event_loop()
+        txaio.config.loop = loop
     log = txaio.make_logger()
```
meejah: I ran all the CTS tests finally, and nothing broke. However, it occurs to me that maybe the real bug is "we shouldn't close() the global event-loop in the first place"..?

PR #980, opened 2018-04-09
Bytes kwargs

meejah:
If you have a Python2 caller, and it's using the msgpack serializer, it'll turn ``str`` instances into ``bytes`` -- and if you make a call like ``call(u"some.uri", arg0="value")`` then the ``arg0`` will be a ``str`` instance in the ``kwargs`` dict

However, if the callee is Python3, it also wants "str" instances when doing ``**kwargs`` but it'll instead have ``bytes`` keys. The "real" fix is the one in "message.py" but I also included the ``except TypeError`` code on the callee side to be a little more robust..

### DIFF
--- a/autobahn/wamp/message.py
+++ b/autobahn/wamp/message.py
@@ -278,6 +278,25 @@ def check_or_raise_extra(value, message=u"WAMP message invalid"):
     return value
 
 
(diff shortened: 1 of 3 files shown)


oberstet:
cool! that was a nasty one ..

PR #990, opened 2018-04-13
try forever to reconnect by default

meejah:
"Try 15 times" is a weird default for reconnecting, IMO. This is kind-of a "breaking change" but .... I think this leads to a better default experience.


oberstet:
fwiw, the "15" comes from https://github.com/crossbario/autobahn-js/blob/master/lib/connection.js#L101

Issue #1007, opened 2018-05-23
WebSocketClientFactory's _batched_timer SHOULD use the passed event loop


karolyi:
Hey,

here's my second issue report in a week:

The `WebSocketClientFactory` can be passed an event loop:

https://github.com/crossbario/autobahn-python/blob/bba4ed8418e59a678012ecb466092bee250b906f/autobahn/asyncio/websocket.py#L288-L289

.... and `WebSocketClientFactory`'s `_batched_timer` **SHOULD** use it, which it doesn't now:
https://github.com/crossbario/autobahn-python/blob/bba4ed8418e59a678012ecb466092bee250b906f/autobahn/websocket/protocol.py#L3798-L3801

Why is this a problem? Because if someone doesn't have a main event loop, autoPings won't work. My application doesn't have a **main** event loop, it only has an event loop in a separate thread, which I initialize later, along with the websocket client.

As far as I've seen, `txaio` uses the main event loop, and the websocket client can't be told to use other than the main initialized one.

Correct me if I'm wrong, but now I have to have an asynchronous main event loop, just to have autopings working?

meejah:
Some overlap with https://github.com/crossbario/autobahn-python/issues/747

For context, this is actually a fairly deep issue since `txaio` is based around a "lowest common denominator" papering-over of the differences between Twisted and asyncio. Twisted has a single global reactor, and so does txaio.

In this particular case, I think passing `loop=` onwards to `make_batched_timer` is definitely "more correct", but also this needs a more comprehensive story, I think. Another option is to configure txaio to use your loop instead of passing a `loop=` kwarg to the factory -- that is: `txaio.config.loop = loop`?

(You're right that "txaio" has ONE event-loop, but it can be any eventloop; the default is `asyncio.get_event_loop()`...)

karolyi:
I agree with the deep statement. I took me a couple hours to get down to the bottom of this. I have it now in 'worksforme' state, but it can be confusing for newcomers, and I think it should be improved in the long run. Maybe my original issue comment will help those.

meejah:
Hmm, I was going to make the one-line fix but this is the base-class and the Twisted one doesn't have a `self.loop`..so not quite the oner-liner I was hoping :/

[...]

PR #1008, opened 2018-05-23
pass loop= kwarg for asyncio

meejah:
a preliminary fix for #1007


meejah:
txaio's make_batched_timer doesn't actually take a loop= kwarg anyway

PR #1012, opened 2018-05-27
implement abort argument in WebSocketAdapterProtocol._closeConnection

PaulosV:
When connection was to be forcefully closed (for example due to failed ping), the `abort` argument of `WebSocketAdapterProtocol._closeConnection` was not honored. The connection was just closed regularly, and it usually took a few minutes, until a `RST` packet came from either side of the connection.

This pull requests fixes that by calling `abort` when asked to.


oberstet:
thx!

jaydg:
When is this fix likely to be released?

Issue #1030, opened 2018-08-03
Connecting to crossbar router started after component regularly fails


aberres:
**Scenario**
A running application with pretty minimal code to just have a `Component` instance connecting the router.

The router is not running. It is started some seconds after the appliction.

**Expected behavior**
Initially the app obviously fails to connect to the router, when retrying it can eventually connect once the router is up.

**Observed behavior**
In a lot of cases no `join` event is fired at all. the component just stops to reconnect.

Find below the code and the log output.

```python
import os

from autobahn.asyncio.component import Component, run


def create_component():
    authentication = None

    component = Component(
        transports=[
            {
                "type": "websocket",
                "url": os.environ.get("AUTOBAHN_ROUTER", "ws://localhost:8080/ws"),
            }
        ],
        authentication=authentication,
        realm="realm1",
    )

    return component


class Application:
    def __init__(self, component):
        self.component = component

        component.on("join", self.on_join)
        component.on("leave", self.on_leave)

    async def on_join(self, session, details):
        print("JUST JOINED")

    async def on_leave(self, session, reason):
        print("JUST LEFT")


def main():
    component = create_component()

    app = Application(component)

    run([component], log_level="debug")


if __name__ == "__main__":
    main()
```

Log output:
```
2018-08-03T16:35:22 Entering re-connect loop
2018-08-03T16:35:22 trying transport 0 using connect delay 0
2018-08-03T16:35:22 connecting once using transport type "websocket" over endpoint "tcp"
2018-08-03T16:35:22 component failed: ConnectionRefusedError: [Errno 61] Connect call failed ('127.0.0.1', 8081)
2018-08-03T16:35:22 Traceback (most recent call last):
  File "/Users/berres/venv/mpptool/lib/python3.5/site-packages/txaio/aio.py", line 498, in done
    res = f.result()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/futures.py", line 294, in result
    raise self._exception
  File "/Users/berres/venv/mpptool/lib/python3.5/site-packages/txaio/aio.py", line 498, in done
    res = f.result()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/futures.py", line 294, in result
    raise self._exception
  File "/Users/berres/venv/mpptool/lib/python3.5/site-packages/txaio/aio.py", line 498, in done
    res = f.result()
[...]

Issue #1010, opened 2018-05-25
Asyncio Component Not Reconnecting After Failing WebSocket Opening Handshake


matthewtorgerson:
Hi, 

I am running Crossbar (17.12.1 on PyPy 5.10.0) and a handful of apps (Asyncio Autobahn 18.5.1 Components on Python 3.6.3) on a seperate server.  When we reboot the server that Crossbar is running on some of the component apps will not connect correctly and will not try to reconnect.  The following is logged:

`2018-05-24T21:54:17+0000 [INFO:autobahn.asyncio.component.Component] connecting once using transport type "websocket" over endpoint "tcp"
2018-05-24T21:54:17+0000 [INFO:autobahn.asyncio.component.Component] Connection failed: ConnectionRefusedError: [Errno 111] Connect call failed ('XXX.XXX.XXX.XXX', XXXX)
2018-05-24T21:54:20+0000 [INFO:autobahn.asyncio.component.Component] connecting once using transport type "websocket" over endpoint "tcp"
2018-05-24T21:54:20+0000 [INFO:autobahn.asyncio.component.Component] Connection failed: ConnectionRefusedError: [Errno 111] Connect call failed ('XXX.XXX.XXX.XXX', XXXX)
2018-05-24T21:54:23+0000 [INFO:autobahn.asyncio.component.Component] connecting once using transport type "websocket" over endpoint "tcp"
2018-05-24T21:54:23+0000 [INFO:autobahn.asyncio.websocket.WebSocketClientProtocol] failing WebSocket opening handshake ('WebSocket connection upgrade failed (404 - NotFound)')
2018-05-24T21:54:23+0000 [WARNING:autobahn.asyncio.websocket.WebSocketClientProtocol] dropping connection to peer tcp:XXX.XXX.XXX.XXX:XXXX with abort=True: WebSocket connection upgrade failed (404 - NotFound)` 

At this point nothing happens.  It does not try to reconnect and it does not exit.  I'm not exactly sure why the connection upgrade is failing, however I do believe the component should try to either reconnect at this point, or shutdown.

Thanks,
Matthew

PaulosV:
Minimal viable code to reproduce:
```
from autobahn.wamp.types import ComponentConfig
from autobahn.asyncio.wamp import ApplicationSession, ApplicationRunner

url = "ws://example.com/NOTwebsocket"
runner = ApplicationRunner(url, "abc")
config = ComponentConfig("abc", {})
[...]

PR #1034, opened 2018-08-07
Fix auto-reconnect in case of WS handshake failure

omorandi:
When using the component API with asyncio, the client component doesn't try to reconnect after a failing WS handshake, as pointed out in issues #1010 and #1030. 

The issue is caused by some naming mismatch between the twisted and asyncio implementations of the `WebSocketAdapterProtocol` class. In particular the wamp component (wamp/component.py) tries to wrap the `connectionLost()` method of the underlying protocol, but the asyncio implementation only has a `connection_lost()` method. 

Curiously enough, the affected code path doesn't seem to be invoked when using the twisted counterpart.

This PR aims at fixing these issues, enabling the component to proceed with the usual exponential backoff retry mechanism when failing to properly connect to the other endpoint.

### DIFF
--- a/autobahn/asyncio/component.py
+++ b/autobahn/asyncio/component.py
@@ -269,7 +269,7 @@ class Component(component.Component):
                 server_hostname=tls_hostname,
             )
             time_f = asyncio.ensure_future(asyncio.wait_for(f, timeout=timeout))
-            return time_f
+            return self._wrap_connection_future(time_f)
 
         elif transport.endpoint[u'type'] == u'unix':
             path = transport.endpoint[u'path']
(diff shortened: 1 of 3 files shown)


oberstet:
No, this is mixing things up. The Twisted implemention is here https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/component.py

omorandi:
@oberstet can I ask you more details about why "this is mixing things up", and the reason for closing the PR without any further comment? 

This PR has never been about the Twisted component implementation, which indeed I didn't touch. It is about its asyncio counterpart, and its integration with the wamp component.
Issues #1010 and #1030 are real: even in a simple router restart scenario, a reconnection is never retried more than once from an asyncio based client, which is kind of a showstopper for any production application based on asyncio and this module.

I have no problem with the fact that this PR is not being accepted as it is, however I think the underlying issue is quite critical and still requires a solution.

PR #1040, opened 2018-08-20
Fix asyncio auto-reconnect in case of WS handshake failure

omorandi:
This PR is a new take on https://github.com/crossbario/autobahn-python/pull/1034, which was closed without further comments.

The PR fixes the way the asyncio component handles the failure to connect to the other endpoint, as reported in issues #1010 and #1030.

In particular it fixes:
* a method naming mismatch (`connection_lost` vs `connectionLost`) in autobahn/asyncio/websocket.py
* an argument type mismatch in the chaining of asyncio connection success callbacks (see autobahn/asyncio/component.py)
* the need transform the 'loss of connection' exception (which could be None), before passing it to upper layers (where the exception cannot be None)

### DIFF
--- a/autobahn/asyncio/component.py
+++ b/autobahn/asyncio/component.py
@@ -269,7 +269,7 @@ class Component(component.Component):
                 server_hostname=tls_hostname,
             )
             time_f = asyncio.ensure_future(asyncio.wait_for(f, timeout=timeout))
-            return time_f
+            return self._wrap_connection_future(time_f)
 
         elif transport.endpoint[u'type'] == u'unix':
             path = transport.endpoint[u'path']
@@ -280,10 +280,30 @@ class Component(component.Component):
                 path=path,
             )
             time_f = asyncio.ensure_future(asyncio.wait_for(f, timeout=timeout))
-            return time_f
+            return self._wrap_connection_future(time_f)
 
         else:
             assert(False), 'should not arrive here'
+
+    def _wrap_connection_future(self, conn):
+
+        # The result of asyncio's AbstractEventLoop.create_connection() and
+        # AbstractEventLoop.create_unix_connection() is a (transport, protocol) pair,
+        # while in the Twisted counterpart, the IStreamClientEndpoint.connect() method
+        # only returns a protocol object. We need to reconcile the two, since the
+        # the transport connection success callback in wamp.component expects only the protocol
+
+        f = txaio.create_future()
+
+        def on_success(result):
+            (transport, proto) = result
+            txaio.resolve(f, proto)
+
+        def on_failure(err):
+            txaio.reject(f, err)
+
+        txaio.add_callbacks(conn, on_success, on_failure)
+        return f
 
     # async function
     def start(self, loop=None):
--- a/autobahn/asyncio/websocket.py
+++ b/autobahn/asyncio/websocket.py
@@ -97,6 +97,10 @@ class WebSocketAdapterProtocol(asyncio.Protocol):
         self._connectionMade()
 
     def connection_lost(self, exc):
+        self.connectionLost(exc)
+
+    # keep naming consistent with the corresponding function in twisted/websocket.py
+    def connectionLost(self, exc):
         self._connectionLost(exc)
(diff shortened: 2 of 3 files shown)


meejah:
Hi, thanks for the pull-request!

I'm trying to understand the changes and repeat referenced other bugs; is the `connection_lost` vs. `connectionLost` "just" some twisted vs asyncio naming thing and/or a case of "too much code is in the 'general' class instead of framework-specific ones"..?
Can you explain why the exceptions need to be wrapped? Thanks!

Also, I know the unit-test coverage isn't great for some of this stuff, but any additional such tests would be *awesome* to have for these kinds of things :)

omorandi:
Hi @meejah, thank you for looking into this.

The root of the issue lies in this portion of code in the wamp component: 
https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/component.py#L726-L751
which assumes  that the underlying protocol implementation has a `connectionLost()` method. That code wraps around `connectionLost()` in order to to "capture" its invocation and thus triggering the rejection of the future returned by `_connect_once()`.

The first problem is that the asyncio implementation of `WebSocketAdapterProtocol` has no method named `connectionLost`, but it has one named `connection_lost`, overridden from the `asyncio.Protocol` class, which is "Called when the connection is lost or closed" (https://docs.python.org/3/library/asyncio-protocol.html#asyncio.Protocol).

So, the first change here: https://github.com/crossbario/autobahn-python/pull/1040/files#diff-05ecf29bd3a5812b4f0b10f58ff172bbR100 renames the old `connection_lost` as `connectionLost` and redirects  the former to the latter, working around the naming mismatch.

With this change in place, another problem emerges: the `on_connect_success()` callback (https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/component.py#L726-L751)  expects a `protocol` argument. That callback is invoked as part of a chain starting in `autobahn.asyncio.component._connect_transport()` (https://github.com/crossbario/autobahn-python/blob/master/autobahn/asyncio/component.py#L264-L283). The problem is that both `loop.create_connection()` and `loop. create_unix_connection()` invocations return a `(transport, protocol)` pair, which clashes with the protocol argument expected by `on_connect_success()`. Here is where the change in autobahn/asyncio/component.py comes into play. I wrap the future in order to transform the tuple result into a single proto argument.

Finally, the exception wrapping: the docs for `BaseProtocol.connection_lost(exc)` (https://docs.python.org/3/library/asyncio-protocol.html#asyncio-protocol) state that "The argument is either an exception object or None". At the end of the failure callback chain there is the `handle_callback_error()" function defined here: https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/component.py#L512. When the `fail` argument (which is still the `exc` coming from `connection_lost`) is None, we fall here: https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/component.py#L543, where `txaio.failure_message(fail)` isn't much happy to receive a None argument.

Regarding test coverage, I can try adding some tests for this.

meejah:
Okay, great thanks! All that makes sense (and does indeed sound like a case of "refactored too much into the 'base' class" which was Twisted-specific).

I really don't like the "wrap methods just to 'listen'" approach, but I don't see a better one here -- if you have any ideas around this, I'd be happy to hear them :) and sorry for the delays getting this looked at and merged.

meejah:
(Just as a heads up, I likely won't have time to confirm and merge this until tomorrow..)

meejah:
Ahh, I see why the `TransportLost()` wrapping: asyncio will/can call `connection_lost` with `None` as an argument. So I believe the correct answer here is that the "listen for connection lost" code should be framework-specific -- because for Twisted we always want to just forward the "fail" object (it'll always be a `Failure`) whereas on asyncio we might want to create a `TransportLost()` exception in case `connection_lost` was called with `None`.

I'll see if I can come up with a better concrete suggestion on how to accomplish this...

meejah:
please see #1041 for a possible solution

omorandi:
Thanks, I commented on #1041. I guess we can close this PR in favor of yours

meejah:
Continued in #1041

PR #1044, opened 2018-08-28
Cancel calls

meejah:
(Note: depends on https://github.com/crossbario/txaio/pull/137)


meejah:
All CI will fail until there's a txaio release with cancel support

Issue #1057, opened 2018-10-04
AttributeError: 'ConnectionResetError' object has no attribute 'value'


npeditto:
Hi,

I noticed this "bug" with autobahn (18.9.2) and Python 3.5.2 in Ubuntu 16.04.5 LTS (Xenial Xerus):

2018-10-03 13:06:43.102 29258 ERROR asyncio [-] Exception in callback WebSocketAdapterProtocol.connection_lost(ConnectionRes...eset by peer'))
handle: <Handle WebSocketAdapterProtocol.connection_lost(ConnectionRes...eset by peer'))>: AttributeError: 'ConnectionResetError' object has no attribute 'value'
2018-10-03 13:06:43.102 29258 ERROR asyncio Traceback (most recent call last):
2018-10-03 13:06:43.102 29258 ERROR asyncio   File "/usr/lib/python3.5/asyncio/events.py", line 125, in _run
2018-10-03 13:06:43.102 29258 ERROR asyncio     self._callback(*self._args)
2018-10-03 13:06:43.102 29258 ERROR asyncio   File "/usr/local/lib/python3.5/dist-packages/autobahn/asyncio/component.py", line 308, in lost
2018-10-03 13:06:43.102 29258 ERROR asyncio     rtn = orig(fail)
2018-10-03 13:06:43.102 29258 ERROR asyncio   File "/usr/local/lib/python3.5/dist-packages/autobahn/asyncio/websocket.py", line 100, in connection_lost
2018-10-03 13:06:43.102 29258 ERROR asyncio     self._connectionLost(exc)
2018-10-03 13:06:43.102 29258 ERROR asyncio   File "/usr/local/lib/python3.5/dist-packages/autobahn/websocket/protocol.py", line 3363, in _connectionLost
2018-10-03 13:06:43.102 29258 ERROR asyncio     WebSocketProtocol._connectionLost(self, reason)
2018-10-03 13:06:43.102 29258 ERROR asyncio   File "/usr/local/lib/python3.5/dist-packages/autobahn/websocket/protocol.py", line 1096, in _connectionLost
2018-10-03 13:06:43.102 29258 ERROR asyncio     reason_string = _maybe_tls_reason(reason.value)
2018-10-03 13:06:43.102 29258 ERROR asyncio AttributeError: 'ConnectionResetError' object has no attribute 'value'
2018-10-03 13:06:43.102 29258 ERROR asyncio

oberstet:
ok, fixed on master via above PR

PR #1059, opened 2018-10-08
Fix AttributeError when ConnectionResetError does not contain "reason" attribute

bashlakov:
This patch fix issue #1057


oberstet:
thanks for contributing!

PR #1066, opened 2018-10-16
Don't eat Component.stop() request when crossbar not connected

om26er:
Currently, if autobahn is not able to connect to a WAMP router (e.g. Crossbar), calling Component.stop() doesn't really exit the process, this PR attempts to fix that.


autobahn/wamp/component.py
```diff
@@ -605,16 +605,20 @@ def transport_check(_):
...
         txaio.add_callbacks(start_f, transport_check, error)
         return done_f
 
     def stop(self):
-        return self._session.leave()
+        if self._session.is_attached():
+            return self._session.leave()
+        else:
+            if getattr(self, '_delay_f', None):
+                return txaio.as_future(txaio.cancel, self._delay_f)
```
meejah: You shouldn't need the `as_future` here I don't think .. just `txaio.cancel(..)`.

Also it's probably better to avoid the `getattr` call and just do `self._delay_f = None` in `__init__` or so? Also, it should be set back to `None` in `attempt_connect` I believe.

I haven't actually *tried* the code yet though :)

om26er: Update the PR with the changes suggested above.

om26er: Also how do we feel about a single method (Component.stop()) return different type of results. The user code would have to deal with a `Future` and None currently, thoughts ?

meejah: Ah, yes missed that detail. Yes we should always return a Future/Deferred from `stop` .. so the `as_future` deals with that.

om26er: Ok, brought that line back ;-)

meejah:
Okay, I think we'd need to attach an errback to `._delay_f` and ignore cancels (maybe "only-if stop called"). I can fix this up later on if you don't get to it first :)

meejah:
Actually, we probably don't need to bother about checking .. we're the only ones that will call `cancel` so we can assume that's because of the `stop()` code-path. Might be worth a comment though :)

meejah:
@om26er can you please rebase this on latest master? (Also i have some more fixes, but a PR looks weird against your branch because i rebased first :/ ... see https://github.com/meejah/AutobahnPython/tree/fix-component-stop

om26er:
> can you please rebase this on latest master?

done. Also added a comment to code

autobahn/wamp/component.py
```diff
@@ -501,9 +506,19 @@ def _start(self, loop=None):
             "done" or with an error if something went wrong.
         """
 
+        # we can only be "start()ed" once before we stop .. but that
+        # doesn't have to be an error we can give back another future
+        # that fires when our "real" _done_f is completed.
+        if self._done_f is not None:
+            d = txaio.create_future()
+            def _cb(arg):
```
meejah:
```suggestion

            def _cb(arg):
```

PR #1071, opened 2018-10-26
unconditionally unregister producer in dropConnection

### LINKED ISSUE #1064: dropConnection appears to leave connections around
Found by ashleyk_ and reported in IRC.

In `failHandshake` in `autobahn/websocket/protocol.py` we call `self.dropConnection(abort=False)`. This then ultimately calls `transport.loseConnection()`. If using TLS, these connections don't actually close (in Twisted at least).

An easy fix is to pass `abort=True` at `protocol.py:3020`. Could even be a Twisted issue..`ashleyk_` had made an example: https://paste.debian.net/plain/1047301

meejah:
should fix #1064 (at least).

### DIFF
--- a/autobahn/websocket/protocol.py
+++ b/autobahn/websocket/protocol.py
@@ -838,6 +838,7 @@ class WebSocketProtocol(object):
         """
         Drop the underlying TCP connection.
         """
+        self.unregisterProducer()
         if self.state != WebSocketProtocol.STATE_CLOSED:
 
             if self.wasClean:


oberstet:
small CI glitch

```
  File "/home/travis/build/crossbario/autobahn-python/autobahn/twisted/websocket.py", line 192, in unregisterProducer
    self.transport.unregisterProducer()
builtins.AttributeError: 'FakeTransport' object has no attribute 'unregisterProducer'
autobahn.twisted.test.test_protocol.Hixie76RejectionTests.test_handshake_fails
```

PR #1075, opened 2018-11-06
component.stop() should not return session.leave() result (if session exists)

npeditto:
In component.py the stop() should do session.leave() and return the a future object if session exists (not return session.leave() result only).

Otherwise we are not able to call again the component.start(...) method after stop() method called.

### DIFF
--- a/autobahn/wamp/component.py
+++ b/autobahn/wamp/component.py
@@ -655,7 +655,7 @@ class Component(ObservableMixin):
     def stop(self):
         self._stopping = True
         if self._session and self._session.is_attached():
-            return self._session.leave()
+            self._session.leave()
         elif self._delay_f:
             # This cancel request will actually call the "error" callback of
             # the _delay_f future. Nothing to worry about.


oberstet:
actually, looks good and merging (as I am pushing a release) --

meejah:
I don't understand what problem you're trying to solve here? `session.leave` does return a future/deferred (that fires when the session actually closes). If you don't care about that, just ignore the return value from

meejah:
(If you *do* care about waiting for the disconnect to actually happen there won't be any way to do that now)

oberstet:
@meejah uups, yes, right .. we do want to return the deferred from leave. lack of sleep, too much hurry;)

npeditto:
In a scenario where I call "component.start(loop)" after a "component.stop()" with autobahn v18.9.2 an application is able to reconnect. Through the latest version (18.10.1) where the component.py has been updated (in particular, _start() and stop() methods) this application is not able to reconnect to Crossbar.

In particular, the attribute "self._done_f" in "_start(...)" method is not reset to "None". I tried also another workaround: in stop() method I set "self._done_f" to "None" before the "return self._session.leave()". So this could be another solution.

meejah:
`Component` already knows how to re-connect, you shouldn't have to call `start()` again (and `_start` is an internal method, you should only call `.start()` the public API). Can you show some example code about what you're trying to accomplish?

npeditto:
> I don't understand what problem you're trying to solve here? `session.leave` does return a future/deferred (that fires when the session actually closes). If you don't care about that, just ignore the return value from `.stop()`

Yes, I was wrong about future object return. Maybe the problem, that I described in the previous post, is related to the "self._done_f" attribute that is not set to "None" in stop() method.

meejah:
Maybe, I'm not sure I ever anticipated someone calling `.start()` more than once. What are you trying to do..?

npeditto:
> `Component` already knows how to re-connect, you shouldn't have to call `start()` again (and `_start` is an internal method, you should only call `.start()` the public API). Can you show some example code about what you're trying to accomplish?

Yes of course. I used in my code the start() method (not _start()).

In autobahn 18.9.2 if I call
`component.stop()`
(because I need for my reasons to close the wamp connection)

and then call 
`component.start(loop)`

my application was able to reconnect to Crossbar again.

meejah:
(And yes, it's probably more-correct to reset `_done_f` to `None` at some point or at least make it an error to call `.start()` a second time if we really can't/won't support that ... but I do want to understand your use-case..)

So essentially you have a reason to want to disconnect your `Component` for some amount of time and then re-connect it..?

npeditto:
> (And yes, it's probably more-correct to reset `_done_f` to `None` at some point or at least make it an error to call `.start()` a second time 
It should be fine to reset _done_f to None

> So essentially you have a reason to want to disconnect your `Component` for some amount of time and then re-connect it..?
Yes, I need to disconnect my application and reconnect it. Resetting the `_done_f` before `return session.leave()` I was able to do that. 
So for me is very important to avoid to "make it an error to call .start() a second time".

meejah:
I think the correct fix is probably for `_done_f` to be reset after the `leave()` deferred fires. So your application code then might look something like:

```
my_component.start(..)
# ...
await my_component.stop()
# ...
my_component.start(..)
```

npeditto:
> ```
> my_component.start(..)
> # ...
> await my_component.stop()
> # ...
> my_component.start(..)
> ```

Exactly that is my workflow.

> I think the correct fix is probably for `_done_f` to be reset after the `leave()` deferred fires. 

I agree! Should the `_done_f` reset be insert into the stop() method (in component.py)?

meejah:
Probably what should be done is anywhere we `.resolve()` the `_done_f` it should be reset to `None`..? (but maybe only on `_stopping`). Might need more thought for the different cases.

meejah:
Moving to a ticket: https://github.com/crossbario/autobahn-python/issues/1077

PR #1080, opened 2018-11-08
Added parameter `dumps_default` to JsonObjectSerializer.__init__

marc1n:
Autobahn-python current implementation does not allow to serialize to JSON non standard types of value in WAMP message kwargs/args (eg. `kwargs={'amount': decimal.Decimal('10.25')}`).

I added parameter `dumps_default` to `JsonObjectSerializer.__init__` in order to allow customize JSON dumping of non-standard types.

Thanks to this change we will be able to do the following:

```python
def dumps_default(obj):
    if isinstance(obj, decimal.Decimal):
        return str(obj)
    raise TypeError('Unable to convert object of {} to JSON'.format(type(obj)))

serializers = [JsonSerializer(dumps_default=dumps_default)]
server = WampRawSocketServerFactory(factory, serializers)
```

### DIFF
--- a/autobahn/wamp/serializer.py
+++ b/autobahn/wamp/serializer.py
@@ -145,13 +145,20 @@ import json
 import base64
 
 
-class _WAMPJsonEncoder(json.JSONEncoder):
+class _JsonDefaultDumper(object):
+    def __init__(self, custom_dumps=None):
+        self._custom_dumps = custom_dumps
 
-    def default(self, obj):
+    def __call__(self, obj):
+        if self._custom_dumps:
+            try:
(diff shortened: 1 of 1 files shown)


oberstet:
WAMP does not support non-standard, serializer specific types, because those cannot be transparently converted between serializers.

oberstet:
one more note: Autobahn/Crossbar support a feature called "WAMP payload transparency". using that you can use any custom types: https://github.com/crossbario/crossbar-examples/tree/master/payloadcodec

this feature is not yet in the official WAMP spec ..

PR #1089, opened 2018-12-05
Allow optional activation_code to be passed in "authextra"

oddjobz:
I've been playing with a Python version of the UI's network discovery code and started it as a WAMP client using the Component class. (I know, I don't need to use Cryptosign for this, but it seemed like a useful exercise to see how it would work) .. I came unstuck when needing to pass an activation code, this addresses the issue. My code looks like this;
```python
        self._key = SigningKey.from_key_bytes(binascii.a2b_hex(PRIVATEKEY))
        extra = {u'pubkey': self._key.public_key(), u'activation_code': ACTIVE }
        self.privkey_hex = self._key._key.encode(encoder=HexEncoder)
        authenticator = create_authenticator(
            u"cryptosign",
            authid=AUTHID,
            authrole='user' if realm == 'com.crossbario.fabric' else 'owner',
            privkey=self.privkey_hex,
            authextra=extra
        )
        self._component = Component(
            transports=[{'url': self._url}],
            realm=realm,
            authentication={'cryptosign': authenticator}
        )
```
Which works with this patch .. without it, I'm not sure how I can apply an activation code with component level cryptosign authentication. (in reality, the activation code is only ever passed on the second run through, first pass generates the activation code request, third and subsequent passes are fully verified)

### DIFF
--- a/autobahn/wamp/auth.py
+++ b/autobahn/wamp/auth.py
@@ -148,6 +148,7 @@ class AuthCryptoSign(object):
 
     def __init__(self, **kw):
         # should put in checkconfig or similar
+
         for key in kw.keys():
             if key not in [u'authextra', u'authid', u'authrole', u'privkey']:
                 raise ValueError(
@@ -159,11 +160,12 @@ class AuthCryptoSign(object):
                     "Must provide '{}' for cryptosign".format(key)
                 )
         for key in kw.get('authextra', dict()):
-            if key not in [u'pubkey']:
+            if key not in [u'pubkey', u'activation_code']:
                 raise ValueError(
                     "Unexpected key '{}' in 'authextra'".format(key)
                 )
 
+        activation_code = kw.get('authextra', {}).get('activation_code')
         from autobahn.wamp.cryptosign import SigningKey
         self._privkey = SigningKey.from_key_bytes(
             binascii.a2b_hex(kw[u'privkey'])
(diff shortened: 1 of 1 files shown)


meejah:
In case this wasn't obvious: you don't have to call `create_authenticator` -- you can just pass all the configuration into `Component`. Like:

```
        authenticator = create_authenticator(
            u"cryptosign",

        )
        self._component = Component(
            transports=[{'url': self._url}],
            realm=realm,
            authentication={
                'cryptosign': {
                    authid=AUTHID,
                    authrole='user' if realm == 'com.crossbario.fabric' else 'owner',
                    privkey=self.privkey_hex,
                    authextra=extra
                }
            }
        )

meejah:
Hrmmm, I see CrossbarFX does accept an `activation_code` in `authextra`.

Arguably, that's maybe a completely new authentication type that's "a lot like cryptosign, but isn't" -- i.e. even if we conclude keeping the activation code "inline" like that is good, we'd want a new client auth-type in CrossbarFX code, not here.

oberstet:
activation codes are crossbarfx specific, and not WAMP cryptosign. essentially, it is a method for trust-on-first-encounter for the pubkey. upcoming, we might get rid of that and move to blockchain based pairing (which I already implemented .. eg nodes can even read their config from IPFS).

PR #1093, opened 2018-12-13
return the wrapped function from component decorators

### LINKED ISSUE #1092: Decorator returning None
Hi,

I am not sure it is a bug but I would like to understand why as soon as I register a function, it becomes impossible to access it. Here is a simple example

```
@comp.register("com.toto")
def toto():
    print("toto")

print(toto) # returns None
```

Why does toto disappear to become None? Is it intended and if yes why ? I'm asking that as I write some pytest and I would like to test functions without the crossbar.io stack in a first approach.
How broken would it be to use wraps from functools to be able to unwrap original functions ?

Thank you for this amazing lib !

meejah:
fixes #1092


meejah:
@oberstet this is a simple fix, should go in next release

PR #1116, opened 2019-02-19
fix #1115: Await tasks after they've been cancelled in asyncio nicely_exit.

### LINKED ISSUE #1115: Tasks are cancelled but never awaited after a shutdown while runing `autobahn.asyncio.component.run`
While runing an asyncio component (using `run`), all tasks are cancelled after the process received a SIGINT or SIGTERM signal, but they are never awaited then the loop is immediately closed

To test it, run the following snippet and kill it with `ctrl+c`

```python
import asyncio
from autobahn.asyncio.component import Component, run

transports = []
realm=""

comp = Component(transports=transports, realm=realm)

@comp.on_ready
async def sleep_forever(*args, **kwargs):

    async def sleep():
        while True:
            print("sleeping")
            await asyncio.sleep(1)

    def wake_up(t):
        try:
            t.result()
        except asyncio.CancelledError:
            print("task cancel")    # never reached
        print("waked up!")         # never reached

    task = asyncio.ensure_future(sleep())
    task.add_done_callback(wake_up)
    await asyncio.wait_for(task, timeout=None)

run(comp)
```
What I expect:
```
2019-02-19T13:01:45 connecting once using transport type "websocket" over endpoint "tcp"
sleeping
sleeping
2019-02-19T13:01:47 Shutting down due to SIGINT
task cancel
waked up!
```

What I get: 
```
2019-02-19T13:01:45 connecting once using transport type "websocket" over endpoint "tcp"
sleeping
sleeping
2019-02-19T13:01:47 Shutting down due to SIGINT
2019-02-19T13:01:47 Task was destroyed but it is pending!
task: <Task pending coro=<sleep_forever() done, defined at /home/rdebroiz/trash.py:22> wait_for=<Task cancelled coro=<sleep_forever.<locals>.sleep() done, defined at /home/rdebroiz/trash.py:25>> cb=[gather.<locals>._done_callback(0)() at /usr/lib/python3.6/asyncio/tasks.py:622]>
```
My understanding of the problem

```python
@asyncio.coroutine
    def exit():
        return loop.stop()

    #  nicely_exit should be a coro
    def nicely_exit(signal): 
        log.info("Shutting down due to {signal}", signal=signal)
        for task in asyncio.Task.all_tasks():  # filter the current task to not cancel it as well
            task.cancel()   

        # here you should await for the previously cancelled task to be done

        asyncio.ensure_future(exit())  # just close the loop or call exit here no need of a task.
    try:
        # here the partials should return a nicely_exit coro  wrapped in a task
        loop.add_signal_handler(signal.SIGINT, partial(nicely_exit, 'SIGINT'))  
        loop.add_signal_handler(signal.SIGTERM, partial(nicely_exit, 'SIGTERM'))
    except NotImplementedError:
        # signals are not available on Windows
        pass
```

One last thing, It would be neat  to have a way to inject some codes that should be executed at shutdown via the run method.

something like `run([my_components],  on_shutdown=my_teardown_coro)`

credit to https://www.roguelynn.com/words/asyncio-we-did-it-wrong-pt-2/

rdebroiz:
Address issue #1115

### DIFF
--- a/autobahn/asyncio/component.py
+++ b/autobahn/asyncio/component.py
@@ -30,7 +30,7 @@ from __future__ import absolute_import, print_function
 import six
 import ssl  # XXX what Python version is this always available at?
 import signal
-from functools import partial, wraps
+from functools import wraps
 
 try:
     import asyncio
@@ -319,7 +319,7 @@ def run(components, log_level='info'):
 
     XXX fixme for asyncio
 
-    -- if you wish to manage the loop loop yourself, use the
+    -- if you wish to manage the loop yourself, use the
     :meth:`autobahn.asyncio.component.Component.start` method to start
     each component yourself.
 
@@ -350,18 +350,31 @@ def run(components, log_level='info'):
     #   signal.signal(signal.SIGINT, signal.SIG_DFL)
 
     @asyncio.coroutine
-    def exit():
-        return loop.stop()
-
     def nicely_exit(signal):
         log.info("Shutting down due to {signal}", signal=signal)
-        for task in asyncio.Task.all_tasks():
-            task.cancel()
-        asyncio.ensure_future(exit())
+
+        tasks = asyncio.Task.all_tasks()
+        for task in tasks:
+            # Do not cancel the current task.
+            if task is not asyncio.Task.current_task():
+                task.cancel()
+
+        def cancel_all_callback(fut):
+            try:
+                fut.result()
+            except asyncio.CancelledError:
+                log.debug("All task cancelled")
+            except Exception as e:
+                log.error("Error while shutting down: {exception}", exception=e)
+            finally:
+                loop.stop()
+
+        fut = asyncio.gather(*tasks)
+        fut.add_done_callback(cancel_all_callback)
 
     try:
-        loop.add_signal_handler(signal.SIGINT, partial(nicely_exit, 'SIGINT'))
-        loop.add_signal_handler(signal.SIGTERM, partial(nicely_exit, 'SIGTERM'))
+        loop.add_signal_handler(signal.SIGINT, lambda: asyncio.ensure_future(nicely_exit("SIGINT")))
+        loop.add_signal_handler(signal.SIGTERM, lambda: asyncio.ensure_future(nicely_exit("SIGTERM")))
     except NotImplementedError:
         # signals are not available on Windows
         pass


rdebroiz:
CI says:
```
 ERROR collecting .tox/py27-asyncio/lib/python2.7/site-packages/autobahn/wamp/test/test_component_aio.py 
.tox/py27-asyncio/lib/python2.7/site-packages/_pytest/python.py:500: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
.tox/py27-asyncio/lib/python2.7/site-packages/py/_path/local.py:668: in pyimport
    __import__(modname)
.tox/py27-asyncio/lib/python2.7/site-packages/_pytest/assertion/rewrite.py:296: in load_module
    six.exec_(co, mod.__dict__)
.tox/py27-asyncio/lib/python2.7/site-packages/six.py:709: in exec_
    exec("""exec _code_ in _globs_, _locs_""")
<string>:1: in <module>
    ???
.tox/py27-asyncio/lib/python2.7/site-packages/autobahn/wamp/test/test_component_aio.py:36: in <module>
    from autobahn.asyncio.component import Component
E     File "/home/travis/build/crossbario/autobahn-python/.tox/py27-asyncio/lib/python2.7/site-packages/autobahn/asyncio/component.py", line 360
E       yield from asyncio.gather(*tasks)
E                ^
E   SyntaxError: invalid syntax
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
```
Don't know what syntax is expected in place of `yield from`

oberstet:
> Don't know what syntax is expected in place of `yield from`

You have to write Deferred-only code (no "yield from" syntax) if the code is to be run on both py2 and py3 without modification - like this unit test ..

rdebroiz:
Thanks, I am not familiar at all with the Deferred-only syntax :S. 

Also what do you think about my suggestion of adding a hook to the `run` method to let the user execute teardown code before the shutdown actually take place (ie. before cancelling all tasks in `exit_nicely`) ?
(properly handle shutdown of our applications is a strong prerequisite for us)

Is it something that could be part of the API or should I use directly the `start` method. 

I'd really like to continue to use `run` as auto re-connection is something hard to deal with and it is only supported with `run` right ?

autobahn/asyncio/component.py
```diff
@@ -350,18 +350,24 @@ def run(components, log_level='info'):
...
-
     def nicely_exit(signal):
         log.info("Shutting down due to {signal}", signal=signal)
-        for task in asyncio.Task.all_tasks():
-            task.cancel()
-        asyncio.ensure_future(exit())
+        tasks = asyncio.Task.all_tasks()
+        tasks.remove(asyncio.Task.current_task())
+        [task.cancel() for task in tasks]
+
```
meejah: This "bare" list-comprehension is a little weird (and I'm not sure it'll "do the right thing" in PyPy..?). I think it would be simpler/more-clear to make this a "for task in tasks: task.cancel()"

autobahn/asyncio/component.py
```diff
@@ -350,18 +350,24 @@ def run(components, log_level='info'):
...
+        try:
+            yield from asyncio.gather(*tasks)
+        except asyncio.CancelledError:
```
meejah: (Further to @oberstet's comment) this will have to be in an inner function that returns a `Future`, and the `nicely_exit` function can return this `Future` using `add_done_callback` to add the cleanup/success handlers...and/or you could use the `txaio` helpers here (i.e. `txaio.gather` and `txaio.add_callbacks`)

"In general" it's best to write code using the `txaio` helpers in Autobahn, but in this particular case it's okay to use asyncio APIs directly (if that's easier/more-clear) because we're already in an asyncio-specific module.

meejah:
@rdebroiz to answer your question: `run` should correctly shut down on a `SIGTERM`, so yes any `Component`s started via `run(...)` should shut down properly, you shouldn't need to use the `.start()` API to get this behavior (IMO).

rdebroiz:
I I did the changes as suggested.

About you're answer on IRC: 
>  on_leave should get called if you exit via a SIGTERM, IMO .. so sounds like a legit bug

**Note** that the components are still abruptly stopped (ie: they won't finish their life cycle, on_leave, on_diconects, etc. won't be called). But at least all asyncio tasks will be "really" cancelled before the asyncio loop is stopped. 
To do it you need to properly stop the component before cancelling all tasks in `nicely_exit`.

Anyway, people probably want to execute some custom cleanup code before components are stopped and all tasks cancelled. My suggestion was to extend the API of the `run`method with a `on_shutdown` (None by default) keyword argument. If `on_shutdown` is provided with a coroutine,  it is first awaited in `nicely_exit` before stopping components, canceling leftover tasks, stopping the loop/reactor, etc. 

It would be a nice way to inject cleanup code to the one executed after the process received a SIGINT/SIGTERM signal.

meejah:
> It would be a nice way to inject cleanup code to the one executed after the process received a SIGINT/SIGTERM signal.

I agree with this .. but what's wrong with "`on_leave` should be called"...? That is, why do we need a new API here? AFAIR the behavior on SIGTERM in Twisted will end up calling `on_leave` (I have not double-checked that recently, however).

oberstet:
> Also what do you think about my suggestion of adding a hook to the `run` method to let the user execute teardown code before the shutdown actually take place

the thing is:

* when the shutdown happens because the router sent GOODBYE, the WAMP client is not allowed to do anything WAMP app wise anymore - but needs to finish the closing WAMP handshake.
* when the shutdown happens because the client shuts down, it could have done any specific cleanup before anyway - because it initiated the shutdown

so that means, we talk about having a hook for custom shutdown code that runs after the actual WAMP session is gone, but before the Future resolved that is returned from `Component.start`?

ideally, we should have this as a dedicated issue .. also to discuss aspects like ^

> `Component.start` returns a Future which will resolve (to ``None``) when we are "done" or with an error if something went wrong.

---

rgd changelog source, that would be https://github.com/crossbario/autobahn-python/blob/master/docs/changelog.rst ;) however, I'll merge it now - will add the changelog when doing a release

rdebroiz:
@meejah 
> I agree with this .. but what's wrong with "`on_leave` should be called"...? That is, why do we need a new API here?

The difference is that `on_leave`/`on_disconnect` is called each time a particular component's session left the realm or the transport got disconnected. which can happen several times during the lifetime of the application holding the component(s). 

on_shutdown would be called only when the application receive a SIGTERM/SIGINT. 
One may not want to do the same things in those two situations. For instance you could want to cancel long-time running task when the application is shutting down but not when one particular component's session left the realm.

PR #1126, opened 2019-02-26
Add on_error callback for application code

mplebanski:
What I want to achieve with this PR is giving application code a way to react to `Component failed: Exhausted all transport connect attempts`. Currently I couldn't find an appropriate way to terminate the execution when it happens. 

Open to discussion and hints.

### DIFF
--- a/autobahn/wamp/component.py
+++ b/autobahn/wamp/component.py
@@ -466,6 +466,7 @@ class Component(ObservableMixin):
                 'leave',        # fired by ApplicationSession
                 'disconnect',   # fired by ApplicationSession
                 'connectfailure',   # fired by base class
+                'error'
             ]
         )
 
@@ -669,6 +670,7 @@ class Component(ObservableMixin):
 
             if not self._can_reconnect():
                 err_msg = u"Component failed: Exhausted all transport connect attempts"
+                self.fire('error', err_msg)
                 self.log.info(err_msg)
                 try:
                     raise RuntimeError(err_msg)
@@ -879,6 +881,11 @@ class Component(ObservableMixin):
         """
         self.on('ready', fn)
 
+    def on_error(self, fn):
+        """
+        A decorator as a shortcut for listening for 'ready' events.
+        """
+        self.on('error', fn)
 
 def _run(reactor, components, done_callback):
     """


oberstet:
sorry for the late answer .. I'm not sure we have the lifecycle of Component completely covered with user extension points / hooks. but we should, I agree;) because the lifecycle of the wamp session and transport is sth an advanced application will want to react to often.

@meejah Does Component has such a hook? also, do we have sth similar to `ApplicationSession.onUserError` in Component?

meejah:
@oberstet there is `on("connectfailure", ..)` to hook into reasons your individual connections are failing. The future from `Component.start` will, I think, errback in the case of all transports being un-connectable ... but there's not a direct hook/listener for that. Since neither `ready` nor `start` have a "pair" (like leave/join connect/disconnect) maybe one of those would be a good name (start/`end` or ready/`failed` or ...?).

If the hook is *just* for "we are giving up, there's no more transport" then on_error sounds maybe too generic?

oberstet:
> If the hook is just for "we are giving up, there's no more transport" then on_error sounds maybe too generic?

yep, agreed. how about _also_ firing `onconnectfailure` - with arguments that identify the connection error as final ("all connection attempts exhausted")

meejah:
> how about also firing onconnectfailure - with arguments that identify the connection error as final ("all connection attempts exhausted")

Yeah, that sounds good too, with a custom `Exception` subclass I guess, right? (So a handler can distinguish "some connection failure" and "Component is giving up now")

meejah:
Hmm, what if we name this new signal `finish` (to mirror `start`) and it is called with a single argument: `None` if everything was fine or an `Exception` subclass if it was not. It will be signaled with the same `Exception` instance as `connectfailure` will now be as well -- say `class AllTransportsFailed(Exception):`

Bonus points: internally, the asyncio vs Twisted implementations rely on being able to pass an "all done" callback to the shared-code `_run` method; they could instead use `on('finish', ...)` after this is done.

oberstet:
> what if we name this new signal finish (to mirror start) 

Ok, I like this, as "not overloading" / being explicit is good, and "all connection attempts exhausted, I am done and won't try again" is different enough from "_one_ connection attempt failed or succeeded .. but I will continue".

The current list of lifecycle events [reads](https://autobahn.readthedocs.io/en/latest/wamp/programming.html#session-lifecycle):

1. `connect`: transport connected
2. `join`: session has successfully joined a realm
3. `ready`: indicates that the realm has been joined and all “join” handlers have completed (including async ones)
4. `leave`: session has left a realm
5. `disconnect`: transport has disconnected

So the `start` event you mention: that is just missing in the docs? and would be sorted before 1. in above list.

And the new proposed event `finish` would be sorted after 5. and when a client has exhausted the connection attempts, the last connection attempt will materialize in one last firing of `disconnect` and then one final firing `finish`?

mplebanski:
> And the new proposed event `finish` would be sorted after 5. and when a client has exhausted the connection attempts, the last connection attempt will materialize in one last firing of `disconnect` and then one final firing `finish`?

How I understand this is `disconnect` will be fire even though the component never connected, right? I believe in a case where one never actually connected to anything then only `finish` should be fired.

meejah:
Yes, `start` is at 0 (and I would propose `finish` would come at 6). I'd have to check if this is *actually* what happens, but as @mplebanski says if you never `connect`-id properly there would be no `disconnect`.

There's also a `connectfailed` which is signaled on any connection failure. So, e.g. a `Component` with two transports that are "just wrong" and won't connect might look like this:

1. start
2. connectfailed
3. connectfailed
4. connectfailed (not fired now, this would be `AllTransportsFailed` or whatever)
5. finish (fired with `AllTransportsFailed`)

meejah:
@mplebanski it seems we have a better handle on this now, as per my last comment -- do you want to tackle this and update your PR (or file a new one if you prefer). Force-pushes are okay on branches, too FYI.

PR #1130, opened 2019-03-05
deactivate websocket opening and ping/pong timeouts - there is sth wr…

oberstet:
…ong ..

### DIFF
--- a/autobahn/websocket/protocol.py
+++ b/autobahn/websocket/protocol.py
@@ -1051,7 +1051,7 @@ class WebSocketProtocol(object):
         self.autoPingPendingCall = None
 
         # set opening handshake timeout handler
-        if self.openHandshakeTimeout > 0:
+        if False and self.openHandshakeTimeout > 0:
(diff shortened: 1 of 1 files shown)


oberstet:
The unit tests for websocket ping/pong fail (expected), like this `autobahn.websocket.test.test_protocol.WebSocketServerProtocolTests.test_auto_ping`

meejah:
I don't understand; what is the problem?

oberstet:
I'm not exactly sure .. I have left this PR (not to be merged) as a reminder to investigate. what I observed is: under high load (in particular many new connections/sec), connections sometimes get killed by the timeout very soon / right after the connection is finally established. I deactivated the timeouts in my testing, and that "solved" the problem. Connections succeed.

My best guess at how we might reproduce: we need a load test that creates many new connections/sec, keeping each for say a couple of ping/pong cycles, and driving up the load so that the router worker is at 100% CPU

PR #1142, opened 2019-03-18
hotfix rel for missing flatbuffers import guards

oberstet:


jberends:
Thanks! Great responsiveness!

PR #1156, opened 2019-03-27
wheel push deploy

meejah:
This needs some Travis setup before it'll work: set `WAMP_PRIVATE_KEY` in the context (similar to the AWS keys).


oberstet:
so `WAMP_PRIVATE_KEY` is set on Travis: gonna merge it

Issue #1158, opened 2019-03-27
Updating WebSocket headers on every connection attempt


meejah:
As brought up in `#autobahn`, currently a WebSocket connection gets its headers from the Factory (via. `self.factory.headers`). However, a Factory may make many connections and it is desirable to have a hook to change the headers for each connection (for example, when doing re-connections).

One way would be to mutate the `.headers` attribute in the Factory. This will "probably" work as there is likely only one "active" protocol instance created per-factory. However, it's not very explicit.

@exarkun suggested: "maybe something like having `WebSocketClientProtocol.onConnect` allowed to return a `ConnectionRequest` (or similar) that can specify headers?

meejah:
There are other options that get set from the factory too, so it probably makes sense to at least consider them in any "return value from `onConnect`" solution.

Perhaps what we could do is internally use "the thing that `onConnect` returns" to get all the options, and the default one can itself fall back into the `Factory` (which provides backwards-compat too). New code can either make their entirely own object (that has all the options) or something like:

```
def onConnect(...):
    options = super().onConnect(..)
    options.headers["foo"] = "bar"
    return options
```

...so our implementation would have to take care to copy any mutables that are in the factory (like `.headers`).

oberstet:
I left some pointers replying to JP on IRC - which I can't access right now .. but at least for WebSocket servers, this is already there.

meejah:
Note that for servers, `onConnect` can already return custom headers, see https://github.com/crossbario/autobahn-python/blob/master/autobahn/websocket/protocol.py#L2442

This feature would be for the client side.

exarkun:
Here's some application code that I wrote that I would like to be better:

https://github.com/exarkun/gridsync/blob/a807f9310700941fc8cc8577e82e95838f94c782/gridsync/streamedlogs.py#L83-L98

As written, the ClientService will drive reconnections always using the same API token.  I would like to be able to refresh the API token on every new connection attempt.  As meejah mentioned, I could probably do this now using prepareConnection and mutation of the factory's headers attribute:

```
def update_key(get_api_token, protocol):
    protocol.factory.headers['authorization'] = "tahoe-lafs {}".format(get_api_token())

...

client_service = ClientService(..., prepareConnection=partial(update_key, get_api_token))
```

But it has the drawbacks that meejah mentioned.  It's quite easy to see how the API example meejah gave (overriding `onConnect`) could be used to do this more nicely.

oberstet:
[...]

PR #1170, opened 2019-04-03
onConnecting implementation

meejah:
As per latest discussion in https://github.com/crossbario/autobahn-python/issues/1158

### DIFF
--- a/autobahn/websocket/interfaces.py
+++ b/autobahn/websocket/interfaces.py
@@ -419,6 +419,22 @@ class IWebSocketChannel(object):
 
     @public
     @abc.abstractmethod
+    def onConnecting(self, transport_details):
+        """
+        This method is called when we've connected, but before the handshake is done.
+
+        :param transport_details: information about the transport.
+        :type transport_details: :class:`autobahn.websocket.types.TransportDetails`
+
+        :returns: A
+            :class:`autobahn.websocket.types.ConnectingRequest`
+            instance is returned to indicate which options should be
+            used for this connection. If you wish to use the default
+            behavior, `None` may be returned (this is the default).
+        """
+
+    @public
+    @abc.abstractmethod
     def onOpen(self):
         """
         Callback fired when the initial WebSocket opening handshake was completed.
--- a/autobahn/websocket/types.py
+++ b/autobahn/websocket/types.py
@@ -152,16 +152,16 @@ class ConnectingRequest(object):
         'protocols',
     )
 
-    def __init__(self, host, port, resource, headers=None, useragent=None, origin=None, protocols=None):
+    def __init__(self, host=None, port=None, resource=None, headers=None, useragent=None, origin=None, protocols=None):
         """
         :param headers: HTTP headers to send in the opening handshake
         :type headers: dict
 
         """
         # required
-        self.host = host
-        self.port = port
-        self.resource = resource
+        self.host = host if host is not None else "localhost"
+        self.port = port if port is not None else 80
+        self.resource = resource if resource is not None else "/"
         # optional
         self.headers = headers if headers else dict()
         self.useragent = useragent
--- a/autobahn/websocket/types.py
+++ b/autobahn/websocket/types.py
@@ -193,24 +193,37 @@ class TransportDetails(object):
     __slots__ = (
         'peer',  # .getPeer() on Twisted, .get_extra_info('peername') on asyncio
         'host',  # .getHost() on Twisetd, .get_extra_info('sockname') on asyncio
+        'is_secure',  # bool
+        'secure_channel_id',  # bytes
     )
     # possibly useful:
     # is_secure
     # peer_certificate  # getPeerCertificate(), .get_extra_info('peercert')
 
-    def __init__(self, peer, host):
+    def __init__(self, peer, host, is_secure, secure_channel_id):
         """
-        :param peer: the address to which we are connected
+        :param str peer: the address to which we are connected
 
-        :param host: our local address
+        :param str host: our local address
+
+        :param bool is_secure: using TLS or not
+
+        :param dict secure_channel_id: information about our
+            channel-binding or None if not using TLS. The only
+            channel-binding currently is `tls-unique` so if this is a
+            `dict` it will be: `{u'tls-unique': bytes}`
         """
         self.peer = peer
         self.host = host
+        self.is_secure = is_secure
+        self.secure_channel_id = secure_channel_id
 
     def __json__(self):
         return {
             'peer': self.peer,
             'host': self.host,
+            'is_secure': self.is_secure,
+            'secure_channel_id': self.secure_channel_id,
         }
 
     def __str__(self):
--- a/autobahn/websocket/types.py
+++ b/autobahn/websocket/types.py
@@ -136,7 +136,10 @@ class ConnectingRequest(object):
     proceeded.
 
     `host`, `port`, and `resource` are all required, everything else
-    is optional
+    is optional. Note that these are values that will be seen by the
+    client and should represent the public-facing host, port and
+    resource to which the client is connecting (not necessarily the
+    action host/port being used).
     """
 
     __slots__ = (
--- a/autobahn/websocket/types.py
+++ b/autobahn/websocket/types.py
@@ -128,6 +128,50 @@ class ConnectionRequest(object):
 
 
 @public
+class ConnectingRequest(object):
+    """
+    Thin-wrapper for WebSocket connection request information provided in
+    :meth:`autobahn.websocket.protocol.WebSocketClientProtocol.onConnecting`
+    after a client has connected, but before the handshake has
+    proceeded.
+    """
+
+    __slots__ = (
+        'headers',
+    )
+
+    def __init__(self, host, port, resource, headers=None, useragent=None, origin=None, protocols=None):
+        """
+        :param headers: HTTP headers to send in the opening handshake
+        :type headers: dict
+
+        """
+        # required
+        self.host = host
+        self.port = port
+        self.resource = resource
+        # optional
+        self.headers = headers if headers else dict()
+        self.useragent = useragent
+        self.origin = origin
+        self.protocols = protocols if protocols else []
+
+    def __json__(self):
+        return {
+            'host': self.host,
+            'port': self.port,
+            'resource': self.resource,
+            'headers': self.headers,
+            'useragent': self.useragent,
+            'origin': self.origin,
+            'protocols': self.protocols,
+        }
+
+    def __str__(self):
+        return json.dumps(self.__json__())
+
+
+@public
 class ConnectionResponse(object):
     """
     Thin-wrapper for WebSocket connection response information provided in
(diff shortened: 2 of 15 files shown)


oberstet:
had a quick look: rgd the new type classes `TransportDetails` and `ConnectingRequest`: I think we might want to make all ctor arguments optional, as certain transports (twisted/asyncio level) such as serial, pipes and UDS don't have host/port ..

meejah:
Okay, I was just looking at the asyncio and Twisted "transport" interfaces and it looked like they always provide `getHost` .. i.e. `ITransport` has `getHost` (but NOT port, so that should be optional).

Looks like `getHost` on a Unix socket returns `address.UNIXAddress(self.socket.getsockname())` and similar for `getPeer`. It seems like asyncio aped these interfaces too (i.e. any "socket" has this stuff).

Examining how we used this currently/before, `request += "Host: %s:%d\x0d\x0a" % (self.factory.host, self.factory.port` so `ConnectingRequest` does need a host + port (of "whatever is visible on the Internet" I guess?) but `TransportDetails` doesn't. Currently, `TransportDetails` (in this PR) just has `host` and `peer` which are defined as "whatever the low-level framework returned". I believe these will always work (and if you're using Twisted, you'll get an `IAddress` for both, which would be a `UNIXAddress` if you've got a unix socket etc).

Put another way, if we want to use `ConnectingRequest` in place of "things in the factory", we currently do blindly use (and thus require) `host`, `port`, and `resource` -- these being "what the client will see" , not necessarily what is *actually* being used. I added a little clarification in the docs.

autobahn/websocket/types.py
```diff
@@ -127,6 +127,96 @@ def __str__(self):
...
+        'host',
+        'port',
+        'resource',
+        'headers',
+        'useragent',
+        'origin',
+        'protocols',
+    )
+
+    def __init__(self, host, port, resource, headers=None, useragent=None, origin=None, protocols=None):
```
oberstet: With WebSocket over non-IP based transports (such as UDS, pipes, serial connections) there is no host or port at the bytes transport level. WebSocket itself however requires a fully qualified HTTP URL - and hence host+port are "needed" in as for creating the HTTP URL for the initial HTTP request that start the WebSocket opening handshake. Eg WebSocket-over-UDS, I often simply use `http://localhost` (because it really doesn't matter .. it's just to make WebSocket happy).

Because of that, I'm not sure if we want to require host and port in the ctor here .. we might make it optional, and fill in defaults automatically (host=="localhost", port==80, resource=="/")

autobahn/websocket/types.py
```diff
@@ -127,6 +127,96 @@ def __str__(self):
...
+
+    __slots__ = (
+        'peer',  # .getPeer() on Twisted, .get_extra_info('peername') on asyncio
+        'host',  # .getHost() on Twisetd, .get_extra_info('sockname') on asyncio
+    )
+    # possibly useful:
+    # is_secure
+    # peer_certificate  # getPeerCertificate(), .get_extra_info('peercert')
+
+    def __init__(self, peer, host):
```
oberstet: Similar to my above comment: what about UDS, pipes, serial, ..?

autobahn/websocket/types.py
meejah: Well, I think we do: I simply copied everything we used from `WebSocketClientFactory` -- and `host, port, resource` were all used without checking. The others might have been null. My reading of the code is that these are all "whatever is public"...? Or something? e.g. here is the original code: https://github.com/crossbario/autobahn-python/blob/master/autobahn/websocket/protocol.py#L3378

autobahn/websocket/types.py
meejah: I believe they all return "something" on their respective platforms (i.e. `ITransport` in Twisted has `getHost()` and it returns an `IAddress` which is .. basically "transport specific" (e.g. TCPAddress or UNIXAddress, etc). That said, I'll double-check pipes though.

autobahn/websocket/types.py
meejah: Oh, wait, I see what you mean, maybe: if they're `None` in ctor then we just make something up (presuming the caller means "I don't care")...? (where "something" == "ws://localhost:80/")

meejah: Oh, yeah, that's exactly what the old code would have done: https://github.com/crossbario/autobahn-python/blob/master/autobahn/websocket/protocol.py#L3838

oberstet:
not sure if that is of use, but we have 2 helpers:

* `autobahn.twisted.util.peer2str`
* `autobahn.asyncio.util.peer2str`

meejah:
Ahhh, interesting. I wonder if it's better to use those (and then `TransportDetails` always has strings for those two items) or if it's better to just declare "you will get whatever your framework returns for the peer and host"...? (These will be `IAddress` implementations in Twisted, not sure what asyncio returns/promises)

oberstet:
>  I wonder if it's better to use those (and then TransportDetails always has strings for those two items) 

yeah, that would be better IMO. we can have only one item: `peer|str == peer2str(..)` => this is what we need in CB and AB (WAMP)

we also need this https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/util.py#L94

this is the channel ID over the underlying TLS (when the transport is running TLS)

I don't know how to get it on asyncio

also, looking at `def transport_channel_id(transport, is_server, channel_id_type):`:

* `transport` and `is_server` => those are available (implicitly) for `onConnecting` to fill
* `channel_id_type` => for now, the only value it accepts is `"tls-unique"`, so practically we could kick it as it is not available in `onConnecting`
* we could instead have `TransportDetails.channel_ids` be a dict with a `"tls-unique"` key filled - so that we can fill in other bindings
* I am not sure 100% if the latter will really work, as I haven't worked with other bindings
* the relevant RFC for TLS channel bindings is this https://tools.ietf.org/html/rfc5929

meejah:
Okay, so convert `peer` and `host` to strings, add `channel_id` (maybe "`None`" if asyncio can't get it?). And maybe `channel_id_type` but if so it's just hard-coded to `tls-unique`? (or maybe we just leave it out for now, and document that it's always `tls-unique` currently and if that changes it'll grow a new attribute?). Do we want `is_secure` (basically "we used TLS" or not?) too? (i.e. if `is_secure` is `False` then `channel_id` and `channel_id_type` [if it exists] is `None` then I guess?)

oberstet:
How about this?

```
TransportDetails.peer: str
TransportDetails.is_secure: bool
TransportDetails.secure_channel_id: dict or None
```

and

```
TransportDetails.secure_channel_id[u'tls-unique']: bytes
```

meejah:
Sounds good. Drop `host`? Or it also gets run through `peer2str` as well ? (it's also an `IAddress` so that should work)

oberstet:
Yeah, this looks all good for merge from my side now! Awesome. Hope it'll do for JP too;) So if don't want anything else to push on this PR from your side, please merge

meejah:
@exarkun do you see anything missing as far as the use-case for Tahoe-LAFS? (It has `"headers"` so I think we're good...)

meejah:
I added some documentation, a changelog entry, and put the `onConnecting` method in the interface.

oberstet:
going to merge it now, as I'd like to run it with ongoing development of crossbar/crossbarfx - to see if there is any unexpected fallout. should @exarkun have any itches with how this now is landed, we can do that via another PR if necessary ..

autobahn/websocket/types.py
```diff
@@ -127,6 +127,109 @@ def __str__(self):
...
+        :param headers: HTTP headers to send in the opening handshake
+        :type headers: dict
+
+        """
+        # required
+        self.host = host if host is not None else "localhost"
+        self.port = port if port is not None else 80
+        self.resource = resource if resource is not None else "/"
+        # optional
+        self.headers = headers if headers else dict()
```
exarkun: Kind of an anti-pattern.  If `headers` is an empty dict then this sets `self.headers` to a _different_ empty dict.  It doesn't really matter for simple, straightforward code but it creates quite a surprise for nasty code that mutates that dict later on.

I'd suggest preserving identity:

```
self.headers = headers if headers is not None else dict()
```

And likewise for `protocols` below.

autobahn/websocket/interfaces.py
```diff
@@ -417,6 +417,22 @@ def onConnect(self, request_or_response):
...
+        :param transport_details: information about the transport.
+        :type transport_details: :class:`autobahn.websocket.types.TransportDetails`
+
+        :returns: A
+            :class:`autobahn.websocket.types.ConnectingRequest`
+            instance is returned to indicate which options should be
+            used for this connection. If you wish to use the default
+            behavior, `None` may be returned (this is the default).
+        """
+
```
exarkun: Are all non-subclassing `IWebSocketChannel` implementations broken now, because they lack this method?  Or is the only supported use through subclassing the ABC?

autobahn/websocket/types.py
oberstet: ah, right! fwiw, the classes in "types" were envisioned as pure value holding classes - but the code as it stands is neither nor. if we'd go with "pure value types", I guess we should (deep)copy everything coming in via ctors. lets see  @meejah 's opinion ..

autobahn/websocket/interfaces.py
oberstet: yeah, people that implemented their own sub-classing stuff have broken classes now until they add the new callback (which is strictly expected by the calling library code). not sure if anyone has done that (wrote own classes for channel). adding dynamic attribute checking to work around that .. not sure if it is worth.

meejah: Well, but also the only way to "implement" an ABC is by subclassing it, isn't it? So then they'll get this empty method (which returns `None` and thus "asks for defaults"). I should double-check that, but I'm fairly certain that's how ABCs work...

autobahn/websocket/types.py
meejah: We can definitely do what @exarkun suggests (which I agree is a good idea). At some point I had some "XXX" comments about this; we can "slow everything down" by deep-copying or make it "up to the caller" to make sure they're not mutating (or, only mutating in a way that makes sense?). We should definitely do the `is not None` thing ... I'm not so sure about deep-copying everything.

oberstet: ok, so no deep copy, and preserve identity for dict: +1

while discussing this, why is `dict()` better than `{}`?

meejah: I prefer `dict()` to avoid confusion with `set()`s. A dict with one item: `{"foo": "bar"}` a set with two items: `{"foo", "bar"}`...you can't make an empty set with `{` and `}` AFAIK, but ....

autobahn/websocket/interfaces.py
meejah: Hrm, I think I'm wrong here ... "A class containing at least one method declared with this decorator that hasn't been overridden yet cannot be instantiated. Such methods may be called from the overriding method in the subclass (using super or direct invocation)."

That's from PEP3119. So, I *think* what that means is that yes, anyone who created their own classes from scratch and declared the `IWebSocketChannel` "interface" via subclassing will now be broken (because they haven't implemented `onConnecting`). A quick test in the repl confirms this -- you'll get `TypeError: Can't instantiate abstract class Foo with abstract methods onClose, onConnect, onConnecting, onMessage, onOpen, onPing, onPong, sendClose, sendMessage, sendPing, sendPong`.

PR #1161, opened 2019-03-28
WIP: ticket 1159: proof-of-concept

meejah:
To make the discussion in #1159 more concrete, here is a proof-of-concept / strawman branch which I believe represents the discussion up until now.

Note: might not even be valid syntax ;) at this point

Issue #1159, opened 2019-03-27
introduce and encourage use of an interface that is more easily testable


exarkun:
I've had some difficulty writing good unit tests for WebSocketClientProtocol-based implementations.  The tools available seem to encourage one of two paths:

* Directly instantiate the relevant objects and then call methods on them _as though_ they were in use by the real Autobahn WebSocket implementation.
* Set up real network servers and clients that interact over real sockets so that the real Autobahn WebSocket implementation is driving them.

The first of these options has the drawback that there's no guarantee the tests will actually drive the application code correctly (and even if it is initially correct the chances of it diverging from reality increase as Autobahn is maintained and changed).

The second of these options has the drawback that it is prone to many different kinds of spurious failure (most code that takes this approach doesn't even manage to get the part where you start listening on a port correct :cry: ).  It also draws in a lot more code and so failures can become harder to spot.  Failures can also end up being caused by other code that's not the main focus of the test.  It's also a lot slower/more expensive as a result of running so much more code.

I think it would be very useful if Autobahn itself offered and encouraged the use of a higher-level interface that allows details about connections/sockets/networks to be contained in an object that can easily be substituted for a testing fake.  A model for this which appears to be working well is the regular HTTP client interface in Twisted, `twisted.web.client.IAgent`.  This has one primary network-enabled implementation in Twisted, `twisted.web.client.Agent` and the third-party(-ish) treq library provides a test-focused implementation, `treq.testing.RequestTraversalAgent`.

Application code written against `IAgent` can accept an implementation and do what it needs to do.  Real-world use of this application code passes in an `Agent` instance (or possibly a wrapper but that's not directly relevant here).  Test use of this same application code passes in a correctly initialized `RequestTraversalAgent`.  The application code doesn't know or care which implementation it gets.  `Agent` goes off and makes real requests over a real network.  `RequestTraversalAgent` decomposes the request and serves up a response to it using a `twisted.web.resource.IResource` provider - with no network, only Python method calls and correct copying of bytes from one place to another in memory.

meejah:
Okay, this sounds good. I agree with your "drawbacks" at the top. I also agree that while there are unit-tests in Autobahn they tend to be of the second variety (and thus lean on `Mock` a lot) -- it would be great to improve these.

There also isn't a great story for *users* of Autobahn: how should they test their websocket-using application?

[...]

PR #1174, opened 2019-04-16
WIP: websocket Agent interfaces

meejah:
Updated version of https://github.com/crossbario/autobahn-python/pull/1161

An implementation of https://github.com/crossbario/autobahn-python/issues/1159

(also note there was an example with DOS line-endings that I fixed in here too, but not really part of this)

### DIFF
--- a/autobahn/twisted/test/test_websocket_agent.py
+++ b/autobahn/twisted/test/test_websocket_agent.py
@@ -1,13 +1,37 @@
 
 from twisted.trial import unittest
-from twisted.internet.defer import inlineCallbacks
+from twisted.internet.defer import inlineCallbacks, Deferred
 from autobahn.twisted.testing import create_memory_agent
 from autobahn.twisted.websocket import WebSocketServerProtocol
+from autobahn.twisted.websocket import WebSocketClientProtocol
+
+from twisted.internet import reactor
 
 
 class SpammingWebSocketServerProtocol(WebSocketServerProtocol):
     def onMessage(self, *args, **kw):
         print("SERVER MESSAGE: {} {}".format(args, kw))
+
+
+class LoggingWebSocketServerProtocol(WebSocketServerProtocol):
+
+    messages = []
+    reactor = None
+
+    def onOpen(self):
+        # "break the loop" in synchonous agent behavior that means
+        # we'll "do" all the message-sending before anything gets a
+        # chance to listn ... but XXX THINK do we actually need to do
+        # this "in autobahn" or is this a produce of "all in-memory
+        # transports" which take zero time?
+        def send_messages():
+            for msg in self.messages:
+                print("sending {}".format(msg))
+                self.sendMessage(msg)
+        self.reactor.callLater(0, send_messages)
+
+    def onMessage(self, *args, **kw):
+        print("SERVER GOT MESSAGE: {} {}".format(args, kw))
 
 
 class TestAgent(unittest.TestCase):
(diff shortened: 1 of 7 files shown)


meejah:
Okay, this is a lot closer to what's wanted in the original ticket I think. What's missing is the "in-memory" version of the agent. I'm a little less-clear on how to proceed here as far as "the state-of-the-art in Twisted for this". I believe that answer is `IOPump`, but I've not built anything with that yet -- maybe @exarkun has better ideas?

oberstet:
minor side note: changes to `examples/twisted/websocket/echo/server.py` are -- whitespace only? do we need/want those?

meejah:
I will take out the `server.py` changes (yes, they're just whitespace; it has DOS line-endings still)

meejah:
Okay, I've put in some PoC code for "how would the memory-agent look". This can now get a client -> server message through, in this test: https://github.com/crossbario/autobahn-python/pull/1174/commits/f69f36f425bbdcdabe976901a394605d878f2454#diff-82145de931975660c372ad98a4c5bd8aR12

So, I guess the idea would be that we could provide some "test" server protocols or client protocols and use them to test "the other side". E.g. a simple "server" protocol could be given a list of replies to send back for each message it receives -- this would be leveraged to test client-side behavior.

The main "win" here is that the "real" twisted WebSocket agent is being used under the hood, and "real" websocket client and server protocols are being used -- instead of all the `Mock()`s etc in many of the existing tests.

The other "win" is that these primitives can be leveraged by *other* users to create tests in their test-suites. Which is exactly the original use-case here for Tahoe.

@exarkun am I on the right track with https://github.com/crossbario/autobahn-python/pull/1174/commits/f69f36f425bbdcdabe976901a394605d878f2454 ?

meejah:
There's a couple `XXX FIXME` comments in there regarding "am I using `MemoryReactor` correctly" if anyone has insight as well...basically, it doesn't implement the `Clock` stuff or domain-resolution things.

meejah:
I think we probably want a `protocol=None` option for `open()` for those cases where code is already using subclassing .. that is, we tell `open` what client protocol we want to establish (or we pass a `Factory`?) By default it can use `WebSocketClientProtocol` (and/or `WebSocketClientFactory`).

oberstet:
I have merged https://github.com/crossbario/autobahn-python/pull/1170 which seems to have resulted in some conflicts in this PR ..

autobahn/twisted/test/test_websocket_agent.py
```diff
@@ -0,0 +1,93 @@
...
+
+class TestAgent(unittest.TestCase):
+
+    def setUp(self):
+        pass
+
+    @inlineCallbacks
+    def test_foo(self):
+        agent = create_memory_agent(SpammingWebSocketServerProtocol, None)
+        proto = yield agent.open("ws://localhost:1234/ws", dict())
```
exarkun: I guess the path component of the URI is ignored here.  It might make sense to respect it at some point.  Requiring that the URI be completely correct for the agent to actually find a protocol is a really convenient way to have your tests prove you've gotten your URI handling correct.  Otherwise you want a separate set of tests for that.

autobahn/twisted/testing/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def send_server_message_to_client(self, client, msg):
+        """
+        Is this an API that agent should have? (I'm playing with ways to
+        build tests .. and tests will want to control what the server
+        sends, and probably wait for client messages...
+        """
+        server_proto = self._servers[client]
+        server_proto.sendMessage(msg)
+        self.flush()
```
exarkun: Hm.  Interesting.  I can certainly imagine this style of test.  It seems like it's closer to the ad hoc end of the testing spectrum.  Test methods call this method as they please to trigger certain codepaths in the client that the test wants to verify.

The upside is plenty of flexibility and minimal per-message/test coding overhead.  The downside is that it's unverified server behavior and the more tests you have, the more you repeat yourself.

The obvious alternative at the other end of the spectrum is to use your real server protocol class.  Then your real server protocol has to play nicely ... and you don't duplicate messaging logic in the tests because you're relying on the real implementation.

I think it might make sense to omit this API from the initial introduction of the agent.  If it proves to be too useful to leave out then it's easy to add later.

autobahn/twisted/websocket.py
```diff
@@ -75,6 +81,144 @@
...
+    """
+    return _TwistedWebSocketClientAgent(reactor)
+
+
+def check_transport_config(transport_config):
+    """
+    raises a ValueError if `transport_config` is invalid
+    """
+    # XXX move me to "autobahn.websocket.util"
+    if not isinstance(transport_config, six.text_type):
```
exarkun: I guess this is partly about compatibility with existing autobahn APIs ... but sure would be nice to have a structured URI type instead of a string :/

autobahn/websocket/interfaces.py
```diff
@@ -36,6 +36,32 @@
...
 
 
+@six.add_metaclass(abc.ABCMeta)
+class IWebSocketClientAgent(object):
+    """
+    Instances implementing this interface create WebSocket
+    connections.
+    """
+
+    def open(self, transport_config, options, protocol_class=None):
```
exarkun: I guess `transport_config` is again for compatibility with existing autobahn APIs... a real type here instead of a dict would be pretty nice though.

autobahn/websocket/interfaces.py
```diff
@@ -36,6 +36,32 @@
...
+        """
+        Open a new WebSocket connection.
+
+        :returns: a future which fires with a new
+            WebSocketClientProtocol instance which has just completed the
```
exarkun: "`protocol_class` or `WebSocketClientProtocol` instance" here I guess.

meejah:
See https://github.com/crossbario/autobahn-python/pull/1186 instead

PR #1180, opened 2019-04-19
Ticket1158 onconnecting followup

meejah:


meejah:
From discussion in https://github.com/crossbario/autobahn-python/pull/1170

PR #1185, opened 2019-05-01
onConnecting updates

meejah:


meejah:
This re-instates the `onConnecting` code, with fixes for "tls failure" cases. This passes all `cbtests` (which are Twisted-based -- but I manually tested asyncio, which doesn't require fixes here). Currently, no `cfctests` are working for me so I'm not confident those pass.

oberstet:
> This passes all cbtests and cfctests

sounds good: I trigger a CI run on crossbarfx ..

oberstet:
@meejah crossbarfx is broken again https://travis-ci.com/crossbario/crossbarfx - could you please have a look? the fails on cbfunc-tests: strange .. I don't know why it happens: have only looked quickly, but I get the same fails also when running locally. did you test `tox -e functests-cb,functests-cfc,py37-cli-1,py37-cli-2` in crossbarfx?

meejah:
Yes, I'll look -- weird, I ran both -cb and -cfc tests locally (but, not I don't have a 3.7 installed currently -- will change that :)

---oh, "pip install failed" is this more of that "pip 19.1 can't install crossbar" fallout?

meejah:
Hmm, I can get the tests to pass locally with 3.7.3 as well ... but the error looks like "the pip 19.1 problem". Maybe https://github.com/crossbario/crossbarfx/pull/289 fixes it?

meejah:
Yes, that PR above fixes test-breakage (but not Sphinx, which may or may not be related?)

PR #1186, opened 2019-05-02
WebSocket testing support, via Agent-style interface

meejah:
This implements #1159 to introduce an "agent-style" interface for creating websocket connections, and a "memory agent" that sets up in-memory connections (instead of real TCP connections). This allows for nicer client/server tests for libraries or applications making use of Autobahn WebSockets.


meejah:
@exarkun an updated PR for websocket testing. I believe the addresses all [1] the comments from the previous PR, and re-bases it to current master.

[1] This code as it exists still doesn't look at the path-part of the WebSocket resource..

PR #1189, opened 2019-05-06
add rawsocket + twisted example

meejah:
A Twisted + ApplicationRunner + rawsocket + unix-socket example (see also https://github.com/crossbario/autobahn-python/issues/409)


oberstet:
ah, great! this covers a transport that is important in crossbar

PR #1197, opened 2019-05-21
convert to python2.7-compatible code

### LINKED ISSUE #1196: Consider making WAMP command line tool python 2.7+ compatible
The WAMP command line tool

https://github.com/crossbario/autobahn-python/blob/c3365f31e5e22aad19b9d054090be93d237a6091/setup.py#L267

is implemented in

https://github.com/crossbario/autobahn-python/blob/master/autobahn/__main__.py

and uses python 3.5+ syntax ("async def" etc).

as far as I can see, using txaio instead (and plain deferred approach instead of co-routines) would be possible, and would make the command line tool work on python 2.7+

meejah:
Fixes #1196

### DIFF
--- a/autobahn/__main__.py
+++ b/autobahn/__main__.py
@@ -25,6 +25,8 @@
 ###############################################################################
 
 from __future__ import absolute_import
+from __future__ import print_function
+
 
 # this module is available as the 'wamp' command-line tool or as
 # 'python -m autobahn'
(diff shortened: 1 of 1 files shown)


meejah:
In hindsight, it was short-sighted to use python3-only syntax. This makes it work with python2.7 OR python3 -- but still requires Twisted.

It could be possible to make this also support python3 + asyncio, but that will require deeper re-factoring (as txaio lacks anything like "a coroutine or @inlineCallbacks" compatibility-piece). And also we'd have to do some work to figure out which platform you're on (basically, "is twisted installed or not").

Not impossible to do the above, but at least this gets python2.7 working "now"...

oberstet:
cool, looks good! yeah, we _could_ also make it work for asyncio (and twisted) using non-coroutine / deferred-only style .. but this PR already fixes the issue at hand (py2)

Issue #1097, opened 2018-12-23
Segmentation fault using large iteration count in autobahn.wamp.auth.derive_key() 


spr0cketeer:
>>> from autobahn.wamp import auth
>>> auth.derive_key('x', 'x', 150000, 32)
[1]    27137 segmentation fault (core dumped)  python

This happens on an autobahn client when using dynamic authentication in crossbar with Django as the backend - Django now using 150000 iterations in its pbkdf2 hasher https://github.com/django/django/blob/master/django/contrib/auth/hashers.py#L239

oberstet:
yeah, I can repro the issue:

```
(cpy371_1) oberstet@intel-nuci7:~/scm/crossbario/crossbar$ python
Python 3.7.1 (default, Dec  6 2018, 11:02:00) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from autobahn.wamp import auth
>>> auth.derive_key('x', 'x', 150000, 32)
Speicherzugriffsfehler (Speicherabzug geschrieben)
```

oberstet:
still there with the latest cpy 3.7.3. 100k iterations works fine though:

``` 
(cpy373_1) oberstet@intel-nuci7:~/scm/crossbario/crossbarfx$ python
Python 3.7.3 (default, Apr 20 2019, 14:09:01) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from autobahn.wamp import auth
>>> auth.derive_key('x', 'x', 15000, 32)
b'sm4ODgt0zxj/ycd36HTpCv/ev7TilE1iyAKzdllZ0vk='
>>> auth.derive_key('x', 'x', 20000, 32)
b'u66hPM8BSa9mBO7wiWaSLR0AmFgHaAASTfsimR4QTiU='
>>> auth.derive_key('x', 'x', 50000, 32)
b'9NMYF8WkdyY2Gc+GH5HFqjY0Na4E4/XFnNmpzoVjE8Q='
>>> auth.derive_key('x', 'x', 100000, 32)
b'7Y0kKpWAJloO5JARvAxqc2E0QVVJSyC/kG7yaJTg16o='
>>> auth.derive_key('x', 'x', 120000, 32)
Speicherzugriffsfehler (Speicherabzug geschrieben)
(cpy373_1) oberstet@intel-nuci7:~/scm/crossbario/crossbarfx$ 
```

[...]

PR #1198, opened 2019-05-22
Ticket1097 pbkdf2

meejah:
Fixes: #1097


oberstet:
yeah, using cryptography is good! not sure why I didn't use that in the first place .. maybe it's new. anyways, cool

Issue #1214, opened 2019-06-18
RegisterOptions: allow sending custom options when registering RPC


samangh:
When using the autobahn, how does one set a custom Option when registering an RPC end point?

The RegisterOptions object throws an exception when using an options that is not specified:

https://github.com/crossbario/autobahn-python/blob/fe70ceebe04849470b0c81b26ff6e65e709b0d33/autobahn/wamp/types.py#L930-L935

I note that [Section 6.1](https://wamp-proto.org/_static/gen/wamp_latest.html#extensibility) of  the WAMP specification specifies that if an option is not understood by a specification then it must be ignored.  My understanding is that this means that that the client should transmit the option to the router but not otherwise act on it.

> **6.1 Extensibility**
> 
> Some WAMP messages contain `Options|dict` or `Details|dict` elements. This allows for future extensibility and implementations that only provide subsets of functionality by ignoring unimplemented attributes. Keys in `Options` and `Details` MUST be of type string and MUST match the regular expression `[a-z][a-z0-9_]{2,}` for WAMP predefined keys. Implementations MAY use implementation-specific keys that MUST match the regular expression _[a-z0-9_]{3,}. Attributes unknown to an implementation MUST be ignored.
>

samangh:
Any thoughts on this?

oberstet:
> When using the autobahn, how does one set a custom Option when registering an RPC end point?

This is not exposed in Autobahn currently.

> What router are you using?

samangh:
I'm using a fork of https://github.com/darrenjs/wampcc. I'm implementing some additional functionality by using the `Options` element, as specified in the WAMP implementation.

oberstet:
The extensibility section in the paragraph https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#rfc.section.6.1 relates to the WAMP community extending the protocol itself - it is _not_ about _user_ extensibility.

User extensibility - aka "custom attributes" or "implementation specific attributes" - is not in the spec, but discussed here: https://github.com/wamp-proto/wamp-proto/issues/345

Thus, the first thing is to nail consensus on above issue on the WAMP spec adding `x_` based custom attributes as an advanced feature to the WAMP spec.

When that is done, we need implementation .. and your PR #1216 is essentially what needs to be done here (it must then be reworked to use `x_` attributes - not a custom dict).

oberstet:
[...]

PR #1216, opened 2019-07-02
Enable passing arbitrary options on CALL, REGISTER, SUBSCRIBE and PUB…

samangh:
This closes issue #1214.

This allows for passing of arbitrary options when doing CALL, REGISTER,
SUBSCRIBE and PUBLISH. This is done passing the 'custom_options'
variable when creating `RegisterOptions`, `CallOptions`,
`SubscribeOptions` or `PublishOptions`.

### DIFF
--- a/autobahn/wamp/message.py
+++ b/autobahn/wamp/message.py
@@ -4662,6 +4662,7 @@ class Register(Message):
         'concurrency',
         'force_reregister',
         'forward_for',
+        'custom_options'
     )
 
     def __init__(self,
@@ -4671,7 +4672,8 @@ class Register(Message):
                  invoke=None,
                  concurrency=None,
                  force_reregister=None,
-                 forward_for=None):
+                 forward_for=None,
+                 custom_options=None):
         """
 
         :param request: The WAMP request ID of this request.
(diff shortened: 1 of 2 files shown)


oberstet:
Oh, cool - thanks for the initiative! Please see my comments https://github.com/crossbario/autobahn-python/issues/1214#issuecomment-507986417

oberstet:
see also the discussion https://github.com/wamp-proto/wamp-proto/issues/345

there are 2 aspects we need for the implementation:

1. extending the library user API to allow users to provide custom attributes (eg extending `ApplicationSession.call`)
2. marshaling custom attributes to the wire

---

rgd 2: as an example. a WAMP call message `[CALL, Request|id, Options|dict, Procedure|uri]` should marshal custom attributes to the wire as follows:

```python
if self.custom_options:
   for k in self.custom_options:
      self.custom_options[u'x_{}'.format(k)] = self.custom_options[k]
```

https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/message.py#L4147

rgd 1:

```python
@public
class CallOptions(object):
    def __init__(self,
                 on_progress=None,
                 timeout=None,
                 caller=None,
                 caller_authid=None,
                 caller_authrole=None,
                 forward_for=None,
                 correlation_id=None,
                 correlation_uri=None,
                 correlation_is_anchor=None,
                 correlation_is_last=None,
                 details=None,
                 custom_options=None):
```

`custom_options` must be dict with strictly string keys. the prefixing with `x_` is done the message class `marshal` function (so the user does provide unprefixed keys)

PR #1232, opened 2019-08-02
Prefer the new name of proto_helpers

exarkun:
Twisted deprecated the `twisted.test.proto_helpers` module and introduced a new `twisted.internet.testing` module with all of the same contents.


meejah:
Thanks!

PR #1234, opened 2019-08-16
cast snappy end message to bytes

asgoel:
The compression caller expects the result of `end_compress_message` to be a byte string, but it was a standard string.

How production-ready is the snappy compression module? Should it just work out of the box?

@oberstet let me know if this is ok.

### DIFF
--- a/autobahn/websocket/compress_snappy.py
+++ b/autobahn/websocket/compress_snappy.py
@@ -412,7 +412,7 @@ class PerMessageSnappy(PerMessageCompress, PerMessageSnappyMixin):
         return self._compressor.add_chunk(data)
 
     def end_compress_message(self):
-        return ""
+        return b""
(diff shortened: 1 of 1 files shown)


oberstet:
> How production-ready is the snappy compression module? Should it just work out of the box?

It should be fully usable in production - however, keep in mind: using snappy as a compression algo is not in the WebSocket standard - the only algo standardized is deflate. I am not aware of any other WebSocket library supporting snappy compression. AutobahnPython supports deflate, bzip2 and snappy ..

asgoel:
thanks @oberstet! That's fine, we use autobahn for both our client and server :) Do you know when you're going to be able to release this patch?

oberstet:
@asgoel we'll likely cut a new release beginning of september ..

PR #1241, opened 2019-09-04
Testnet4

oberstet:


om26er:
That might be conflicting with https://github.com/crossbario/autobahn-python/pull/1242 big time ;-)

oberstet:
mmh, yeah=(

Issue #1193, opened 2019-05-14
time.clock is removed in Python 3.8


tirkarthi:
`time.clock` was deprecated in 3.3 and was removed in 3.8 with PR https://github.com/python/cpython/pull/13270 . I think below line might cause incompatibility in 3.8 on Windows so I am creating an issue for the same.

https://github.com/crossbario/autobahn-python/blob/87ea1cca817dbc41a783b6b27483197bfa69ac78/autobahn/util.py#L466

Thanks

ivankravets:
Hi @oberstet 
Do you plan to fix this issue? We receive a lot of issues related to `time.clock` when Python 3.8 is installed.

oberstet:
from quick reading, seems like

* `time.clock` is depr. since 3.8 on all platforms
* autobahn only uses `time.clock` when on windows

so we should extend this line (and only that)

https://github.com/crossbario/autobahn-python/blob/87ea1cca817dbc41a783b6b27483197bfa69ac78/autobahn/util.py#L466

to use

https://docs.python.org/3.3/library/time.html#time.perf_counter

when on >=3.8 (and on windows, which is already the case for the code line I linked)

ivankravets:
Yes, we patched `autobahn-python` with `time.perf_counter` for pre-compiled packages.

om26er:
fixed via https://github.com/crossbario/autobahn-python/pull/1259/files

PR #1258, opened 2019-10-17
add python 3.8 compatibility

om26er:
Fixes https://github.com/crossbario/autobahn-python/issues/1193


oberstet:
no, this is wrong: it exports new symbols (get_clock) which are then used on linux too - where is (if any) should use the existing rtime

oberstet:
pls see my comment here https://github.com/crossbario/autobahn-python/issues/1193#issuecomment-543565756

om26er:
@oberstet do we also want to deal with examples as well, because they use `time.clock()`, if not, then yes, the change can be as simple as that.

oberstet:
ah, right. the examples should use `autobahn.util.rtime` ..

PR #1259, opened 2019-10-18
Add py38 compatibility

om26er:
Take 2 at https://github.com/crossbario/autobahn-python/issues/1193

### DIFF
--- a/autobahn/util.py
+++ b/autobahn/util.py
@@ -455,6 +455,17 @@ def generate_serial_number():
     return generate_token(char_groups=6, chars_per_group=4, chars=DEFAULT_TOKEN_CHARS, sep=u'-', lower_case=False)
 
 
+def _get_clock_impl():
+    if sys.version_info >= (3, 3):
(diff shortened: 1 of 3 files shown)


autobahn/util.py
```diff
@@ -463,7 +463,10 @@ def generate_serial_number():
     # first call to this function, as a floating point number, based on the
     # Win32 function QueryPerformanceCounter(). The resolution is typically
     # better than one microsecond
-    _rtime = time.clock
+    if sys.version_info >= (3, 8):
```
ivankravets: `perf_counter` was added in Python 3.3. Why we can't use it for all Python 3.3+ line? 
`time.clock` produces a warning on `<Python 3.8`

oberstet:
> perf_counter was added in Python 3.3. Why we can't use it for all Python 3.3+ line?

because we don't want to when on linux

ivankravets:
@oberstet what do you mean?

PR #1263, opened 2019-11-02
Fix argument type check in sendMessage

0xfadeef:


0xfadeef:
Are these type checks really necessary though?

om26er:
I peaked inside the code and the change looks good to me :+1:

meejah:
yeah, also `assert` goes away if you happen to use "optimized". And the checks should probably use `isinstance` not comparing to types ... but also they do provide some checks / sanity.

*Ideally*, we'd probably convert these all to mypy or similar these days (bear in mind Autobahn started in Python2 era).

meejah:
thanks for the patch!

PR #1277, opened 2019-12-22
Remove usage of six

om26er:
Based on the information on the [forum](https://forum.crossbar.io/t/call-for-comments-retire-python-2-support/1660/2) I believe we'd have to drop Python2 support at one stage.

Since this was mostly done using `bash` (find, sed), lets see if the CI likes the changes.

Note: this branch may not land at all, so don't panic if you depend on Python2.


oberstet:
had a quick look (obviously a whole bunch of changed lines .. but that is expected of course): looks good. I have created a proper issue: https://github.com/crossbario/autobahn-python/issues/1282 and sticked that to a milestone for the next release

PR #1291, opened 2020-01-14
[WIP] CI fix playground

om26er:


om26er:
The playground is closed. CI seems to be OK

PR #1307, opened 2020-02-04
add procedure parameter validation

om26er:
Optionally validate incoming call parameter types

* Only validates parameters with type hints


om26er:
We could potentially just introduce a "helper" decorator, as proposed here https://forum.crossbar.io/t/using-type-hinting-for-parameter-type-validation/1670, instead of changing `protocol.py`. But having something "central" could save lots of parameter validations we do in Crossbar/CrossbarFX

autobahn/wamp/protocol.py
```diff
@@ -58,6 +59,17 @@ def is_method_or_function(f):
     return inspect.ismethod(f) or inspect.isfunction(f)
 
 
+def check_args_types(func, *args, **kwargs):
+    # Converge both args and kwargs into a dictionary
+    arguments = inspect.getcallargs(func, *args, **kwargs)
+    response = []
+    for name, kind in typing.get_type_hints(func).items():
```
om26er: We could do without importing `typing` there as well by using `func.__annotations__` instead.

autobahn/wamp/component.py
```diff
@@ -381,7 +396,7 @@ def do_subscription(session, details):
             return fn
         return decorator
 
-    def register(self, uri, options=None):
+    def register(self, uri, options=None, validate=False):
```
om26er: That parameter could be used to toggle the behavior, making it optional per procedure and "off" by default

autobahn/wamp/component.py
```diff
@@ -343,6 +344,20 @@ def describe_endpoint(self):
...
 
 
+def type_check(func):
+    async def _type_check(*args, **kwargs):
+        # Converge both args and kwargs into a dictionary
+        arguments = inspect.getcallargs(func, *args, **kwargs)
+        response = []
+        for name, kind in func.__annotations__.items():
+            if name in arguments and type(arguments[name]) != kind:
+                response.append("'{}' required={} got={}".format(name, type(arguments[name]).__name__, kind.__name__))
```
meejah: Should this test be `isinstance` instead? (I think it should; if I demand `Base` as the type, I'd expect my subclass to be accepted)

autobahn/wamp/component.py
meejah: `Component.register` should mirror the args to the "real" register .. so, this is fine for a PoC but `session.register` should probably also take `validate=` if we're going to integrate this fully. I think most users are probably using `Session.register(..)` right now.

om26er: Fixed that now; moved the validate code to protocol.py.

autobahn/wamp/component.py
om26er: done

om26er:
Well, I guess that is no longer a proof of concept ;-)

oberstet:
> Well, I guess that is no longer a proof of concept ;-)

ok, so if we really wanna have it more than "experimental", I think it would be wise to give it a little more thought .. rgd public API changes etc. here is my view / couple of thoughts:

* the type checking _can_ activated on a registered procedure ("register case") or an event handler ("subscribe case")
* to activate type checking, a user needs to enable a flag either when doing register/subscribe or when adding the respective decorator (https://github.com/crossbario/autobahn-python/blob/108408cb1c274570e7e06f31427058a7b04c72fc/autobahn/wamp/uri.py#L319)
* any change to the public API should be mirrored in both the new and recommended component API, as well as the old ApplicationSession API
* in particular, a change to `register()` should be reflected here https://github.com/crossbario/autobahn-python/blob/108408cb1c274570e7e06f31427058a7b04c72fc/autobahn/wamp/interfaces.py#L564
* the next question would be: add the new "enable type check flag" as another keyword argument to register/subscribe or add this flag to RegisterOptions/SubscribeOptions?
* and the final question: name of the flag? fwiw, I think I'd prefer `typecheck` or `enable_typecheck` rather than `validate` as the latter to me implies some kind of checking against _app level_ rules rgd app payload .. rather than checking against technical types.

oberstet:
I think `RegisterOptions` and `SubcribeOptions` should remain "WAMP wire options" only, and the flag should be (as it is in the PR) a new parameter to `register` (and `subscribe`):

```python
def register(self, endpoint, procedure=None, options=None, prefix=None, typechecks=None):
    """
    ...
        :param typechecks: Enable automatic type checking against (Python 3.5+) type hints specified on the ``endpoint`` callable.
        :type typechecks: bool
    ...
    """
```

oberstet:
above extension of register/subscribe interface could then reflect in both component and ApplicationSession

and finally, the decorators

https://github.com/crossbario/autobahn-python/blob/108408cb1c274570e7e06f31427058a7b04c72fc/autobahn/wamp/uri.py#L319

could be extended

```python
@public
def register(uri, options=None, typechecks=None):
    """
    Decorator for WAMP procedure endpoints.

    :param uri:
    :type uri: str

    :param options:
    :type options: None or RegisterOptions

    :param typechecks: Enable automatic type checking against (Python 3.5+) type hints specified on the ``endpoint`` callable. Types are checked at run-time on each invocation of the ``endpoint`` callable. When a type mismatch occurs, an :class:`autobahn.wamp.error.TypeError` is raised automatically and returned to the callee (via the router).
    :type typechecks: bool
    """
```

om26er:
> I think `RegisterOptions` and `SubscribeOptions` should remain "WAMP wire options" only, and the flag should be (as it is in the PR) a new parameter to `register` (and `subscribe`):

Done

> above extension of register/subscribe interface could then reflect in both component and ApplicationSession

That is done as well

> and similar for `autobahn.wamp.uri.subscribe`

This is done as well.

Regarding `autobahn.wamp.error.TypeError`, that class does not exist, shall I create a new class and a new python package under autobahn/wamp, called `error`, or just create a new class called `TypeError` that inherits from `Error` class ?

om26er:
>     * and the final question: name of the flag? fwiw, I think I'd prefer `typecheck` or `enable_typecheck` rather than `validate` as the latter to me implies some kind of checking against _app level_ rules rgd app payload .. rather than checking against technical types.

How about naming that variable `check_types` instead, since that implies of an "action"

oberstet:
> Regarding autobahn.wamp.error.TypeError ..

sorry, mixed up the package name .. best would be derive a new class `autobahn.wamp.exception.TypeError` from `ApplicationError` using the WAMP error URI `wamp.error.type_error` 

https://github.com/crossbario/autobahn-python/blob/108408cb1c274570e7e06f31427058a7b04c72fc/autobahn/wamp/exception.py#L91

autobahn/wamp/interfaces.py
```diff
@@ -672,6 +679,13 @@ def subscribe(self, handler, topic=None, options=None):
         :param options: Options for subscribing.
         :type options: An instance of :class:`autobahn.wamp.types.SubscribeOptions`.
 
+        :param check_types: Enable automatic type checking against (Python 3.5+) type hints
+            specified on the ``endpoint`` callable. Types are checked at run-time on each
+            invocation of the ``endpoint`` callable. When a type mismatch occurs,
+            an :class:`autobahn.wamp.error.TypeError` is raised automatically and returned
+            to the callee (via the router).
```
oberstet: rethinking this, we should probabyl:

* for the RPC case (register): return the error to the _caller_ and additionally raise the error in `onUserError` (within the _callee_ side)
* for the PubSub case (subscribe): only raise the error in `onUserError` (within the _subscriber_ side, as there is no message back to the _publisher_ nor to the router - until we have "acknowledged event delivery" in WAMP)

oberstet:
@om26er apart from my last review comment ^ (sorry, slight change) this now looks cool for me (in terms of what we want on the public api side .. impl isn't complete obviously)!

@meejah ?

om26er:
> sorry, mixed up the package name .. best would be derive a new class `autobahn.wamp.exception.TypeError` from `ApplicationError` using the WAMP error URI `wamp.error.type_error`

So I introduced `TypeCheckError` instead, since `TypeError` over shadows the builtin exception. We can agree on a different name, if required.

oberstet:
> So I introduced TypeCheckError instead, since TypeError over shadows the builtin exception.

yep, perfect!

autobahn/wamp/interfaces.py
om26er: Latest commit tries to address that, kindly take a look.

om26er: Improved the docstrings a bit...

oberstet:
so for me, this is good to merge. we can add an example, doc page, CI tests and changelog in a follow up ..

autobahn/wamp/protocol.py
```diff
@@ -457,7 +457,7 @@ def onUserError(self, fail, msg):
         """
         Implements :func:`autobahn.wamp.interfaces.ISession.onUserError`
         """
-        if isinstance(fail.value, exception.ApplicationError):
+        if hasattr(fail, 'value') and isinstance(fail.value, exception.ApplicationError):
             self.log.warn('{klass}.onUserError(): "{msg}"',
```
meejah: When does `fail` ever not have a `.value` attribute?

autobahn/wamp/protocol.py
```diff
@@ -490,6 +490,32 @@ def _swallow_error(self, fail, msg):
...
+        :return_error: Whether the error should be returned to the caller in
+                       addition to being notified to the user code.
+        :type: bool
+        """
+        async def _type_check(*args, **kwargs):
+            # Converge both args and kwargs into a dictionary
+            arguments = inspect.getcallargs(func, *args, **kwargs)
+            response = []
+            for name, kind in func.__annotations__.items():
+                if name in arguments and not isinstance(arguments[name], kind):
```
meejah: If `func` has no annotations at all, `__annotations__` is just empty, right? So that brings up a sort of meta-question, then: if you've asked for type-checking, but the procedure you've registered doesn't have any annotations should that be a `ValueError` or similar? (That is, an error because you're using the API improperly?)

That is, maybe `type_check` itself should check if `func.__annotations__` has any entries and throw a `ValueError` if it doesn't?

autobahn/wamp/protocol.py
```diff
@@ -490,6 +490,32 @@ def _swallow_error(self, fail, msg):
...
+                    response.append(
+                        "'{}' required={} got={}".format(name, kind.__name__, type(arguments[name]).__name__))
+            if response:
+                error = TypeCheckError(', '.join(response))
+                self.onUserError(txaio.create_future_error(error))
+                if return_error:
```
meejah: I haven't completely thought through this yet, but don't we just want to `raise TypeCheckError()` here, and let the framework deal with the error the way it normally does? That would imply getting rid of `return_error` too since then it would always  be `True`. (Maybe I'm just asking: why the distinction here?)

autobahn/wamp/uri.py
```diff
@@ -142,6 +142,10 @@ def __init__(self, uri, target, options=None):
 
         :param options: An optional options object
         :type options: None or RegisterOptions or SubscribeOptions
+
+        :param check_types:
+        :type check_types: bool
+
```
meejah: This should have a description too, i guess (here and below).

autobahn/wamp/protocol.py
om26er: My IDE's code analyzer had detected that. That issue happens here https://github.com/crossbario/autobahn-python/blob/8cba0623b6891d253a75ea16e95d49374e0010ec/autobahn/wamp/protocol.py#L825

autobahn/wamp/protocol.py
om26er: I am fine either way. What do you think @oberstet ?

autobahn/wamp/uri.py
om26er: Added that now

autobahn/wamp/protocol.py
meejah: Ah, good catch! That should be fixed too then .. I guess maybe a separate PR? But yeah I think `onUserError` should always be receiving `Failure`s

autobahn/wamp/protocol.py
om26er: I had changed that per this comment https://github.com/crossbario/autobahn-python/pull/1307#discussion_r375840597

autobahn/wamp/protocol.py
oberstet: my view:

if I added a type-check annotator to an endpoint procedure (for callee) or event handler I want _all_ parameters of the callable to be type checked. if I missed a type annotation for one or more parameters, the type-check annotator should throw an error.

one can argue for that error to be a ValueError or TypeError - depending on perspective (a Value is just a Type-0 type .. blabla) .. I don't care which one;)

autobahn/wamp/protocol.py
om26er: @oberstet ^^

oberstet: yes, just raise TypeCheckError and let existing framework code handle the rest - so that we don't duplicate behavior in the type-check decorator

oberstet:
so this has received extensive review, comments and respective changes - and I'm gonna merge it now;) @om26er awesome work! this is potentially a really powerful feature. lets give it some real world exposure .. I am gonna use it in the xbrnetwork backend. should come handy as all the boilerplate for tight type checks get boring quickly. this feature should do away with that =)

autobahn/wamp/protocol.py
om26er: Fixed that with a following commit https://github.com/crossbario/autobahn-python/pull/1314

autobahn/wamp/protocol.py
om26er: Opened https://github.com/crossbario/autobahn-python/issues/1317 for both suggestions above

autobahn/wamp/protocol.py
om26er: ^ Reported that as https://github.com/crossbario/autobahn-python/issues/1318

Issue #1327, opened 2020-02-25
Websocket transport disconnects if PONG arrives late even if other (data frames) are arriving


DZabavchik:
When client is on slow connection and large payload is scheduled PONG responses end up at the end of the send queue.  Server socket schedules PONG timeout and closes the connection if PONG does not arrive on time. 

I guess proper fix is for all client libraries to use priority queue or schedule PONG responses ahead of any pending data frames.

oberstet:
> When client is on slow connection and large payload is scheduled PONG responses end up at the end of the send queue.

yeah, I can see the problem

> I guess proper fix is for all client libraries to use priority queue or schedule PONG responses ahead of any pending data frames.

rgd the latter, probably needs more:

* configure auto-fragment size (we have that in autobahn-python and crossbar)
* since autobahn-python is single-thread async, and when using twisted: consumer-producer pattern

for the latter, here is an example https://github.com/crossbario/autobahn-python/tree/master/examples/twisted/websocket/streaming

----

having said that, I just saw your PR (https://github.com/crossbario/autobahn-python/pull/1328): I like that a lot!

it is yet another way (different from above routes) - which isn't strictly RFC6455 compliant to the letter (I think .. without re-reading the spec), but I wouldn't care, because it seems like a pragmatic and simple solution and does not harm (rgd what the spec really cares about)

@meejah thoughts?

DZabavchik:
Thank you. That was the intent.  Since that PONG is not used for anything except for:
a. Keep-alives and preventing load-balancers from dropping silent connections
b. Detecting lost route to client (no FIN ACK) 

And traffic arriving from client ensures both + we are actually sending PING, just don't care if PONG arrives on time and with same content.

meejah:
Yeah, I like the idea: after all, the intent is to know "is this connection alive" and if we're getting data, it is...

PR #1328, opened 2020-02-25
Fix for #1327 - cancel Auto Ping Timeout 

DZabavchik:
and use inbound data frames in leu of PONG response

### DIFF
--- a/autobahn/websocket/protocol.py
+++ b/autobahn/websocket/protocol.py
@@ -1673,14 +1673,15 @@ class WebSocketProtocol(ObservableMixin):
                 self.logRxFrame(self.current_frame, self.control_frame_data)
             self.processControlFrame()
         else:
-            self._cancelAutoPingTimeoutCall()
-            
             if self.state == WebSocketProtocol.STATE_OPEN:
                 self.trafficStats.incomingWebSocketFrames += 1
             if self.logFrames:
                 self.logRxFrame(self.current_frame, self.frame_data)
 
             self._onMessageFrameEnd()
+
+            if self.autoPingTimeoutCall:
+                self._cancelAutoPingTimeoutCall()
 
             if self.current_frame.fin:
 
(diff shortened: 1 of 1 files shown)


oberstet:
so as discussed on https://github.com/crossbario/autobahn-python/issues/1327, this is a good approach we would like to have and merge

rgd the PR code

trivial issue is flake8 complaining https://travis-ci.org/crossbario/autobahn-python/jobs/655024730#L330

then, did you test this in real world? does it actually solve the issue (connection do indeed stay alive)?

my main worries beyond that: how does this change behave at scale? 

the reason is, that it that calls `self._cancelAutoPingTimeoutCall()` in the hot path of frame processing (on each received frame):

https://github.com/crossbario/autobahn-python/pull/1328/files#diff-83d22d16d23d299b646c94c41445c1aaR1676

also, Autobahn has sophisticated ping/pong handling ("batching") that originated from this discussion:

https://github.com/crossbario/autobahn-python/issues/642

we have to be sure this PR works with the batching ..

ideally we'd have mass-coms/scaling tests in our CI so we'd immediately know answers to above .. but we don't:(

DZabavchik:
> then, did you test this in real world? does it actually solve the issue (connection do indeed stay alive)?
We had this fix in production since April 2017 (autobahn-python 16.1) .  I just noticed that PR was never filed for a well tested fix. https://github.com/NinjaRMM/autobahn-python/commit/6de1b43350bb4e6ce25c998076d2b2dfc38db472 

I participated in #642 also :)

There was no performance impact, as most of the time there is no pending PONG timeout.   We did see a huge improvement in reliability.

A slight improvement can be made by moving `if self.autoPingTimeoutCall:` outside of the function to save a call.

DZabavchik:
Forgot to mention that because this is targeting large transfers and to minimize impact it is only done in `onFrameEnd`.   Updated PR with minor optimizations and fixes for travis-ci

oberstet:
> We had this fix in production since April 2017 (autobahn-python 16.1)  ... We did see a huge improvement in reliability.

ah, perfect. thanks!

> I participated in #642 also :)

huh, right, sorry;)

thanks for the patch, and also your latest improvements to the PR!

gonna merge it now=)

PR #1337, opened 2020-03-03
Fix confusion between paying and payment channel

disalvoquentin:
In seller.py, balance calls payment instead of paying.


oberstet:
huh, good catch! thx=)

PR #1339, opened 2020-03-03
resource wrapper fix

meejah:
See https://github.com/crossbario/crossbar/issues/1719


om26er:
Tested, fixes the issue.

PR #1351, opened 2020-03-30
added ability to disable TLS channel binding

sgoudelis:
I added the ability to disable TLS channel binding.

PR #1353, opened 2020-04-05
make channel_id_type optional in transport_channel_id()

oberstet:


sgoudelis:
I added a comment here: https://github.com/crossbario/autobahn-python/pull/1351

PR #1368, opened 2020-04-23
added ability to disable TLS channel binding

### LINKED ISSUE #1355: missing channel_binding handling in authenticators
channel_binding attribute is indeed missing in https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/auth.py#L145 - that's a bug in component API (which is using authenticators) vs ApplicationSession (as it is used in the example ^, and which isn't using authenticators, but onChallenge callback)

jameshilliard:
I rebased #1351 which seems to now fix #1355 in master.


oberstet:
ah right, thanks for following this up, and fixing!

PR #1369, opened 2020-04-24
Split xbr modules into PEP 420 style package.

jameshilliard:
This should fix a bug where xbr modules were being installed when not requested.


jameshilliard:
For example when installing with `pip install .`

With this change xbr modules are correctly not installed:
```python
>>> from autobahn.xbr._blockchain import SimpleBlockchain
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'autobahn.xbr'
```

Without this change they are incorrectly installed:
```python
>>> from autobahn.xbr._blockchain import SimpleBlockchain
No module named 'mnemonic'Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jameshilliard/autobahn-python/autobahn/xbr/_blockchain.py", line 27, in <module>
    import web3
ModuleNotFoundError: No module named 'web3'
```

oberstet:
this works as designed:

* if you want just the bare minimum, do `pip install .`
* if you want xbr (and everything), do `pip install .[all]`
* there are more install flavors, eg you can do `pip install .[accelerate,asyncio,twisted,encryption,compress,serialization,scram,xbr]` (see setup.py)

just tested ^ in a fresh venv:

https://gist.github.com/oberstet/3a258c78b6589cff0dba187f2388b637

jameshilliard:
> if you want just the bare minimum, do `pip install .`

That's what I was doing which was resulting in xbr modules still being installed.

oberstet:
no, the log shows that xbr is not installed when you dont provide the install flavor

https://gist.github.com/oberstet/3a258c78b6589cff0dba187f2388b637#file-gistfile1-txt-L44

and the log shows that xbr is installed when the option was given

https://gist.github.com/oberstet/3a258c78b6589cff0dba187f2388b637#file-gistfile1-txt-L188

jameshilliard:
See [here](https://gist.github.com/oberstet/3a258c78b6589cff0dba187f2388b637#file-gistfile1-txt-L47-L48) where the installed xbr module is trying to load `web3`.

oberstet:
the error is not the installed file, but the missing dependency:

ModuleNotFoundError: No module named 'web3'

the reason is the missing install flavor

all good, nothing to fix

jameshilliard:
> the error is not the installed file, but the missing dependency:

If the `xbr` module wasn't being installed you would not see that type of error at all, the error originates from the `_blockchain.py` xbr module which shouldn't be installed unless `xbr` is requested.

> the reason is the missing install flavor

The issue is the install flavor does not control `xbr` module installation at all, this PR fixes that.

oberstet:
ok, now I see what you mean. yes, true. it controls the installation of deps. but anyhow: this is a super intrusive change that will very likely break some things down the line. and on the other hand, I don't see a big upside. what is the problem fixed? sorry, we won't merge this. if you disagree, best would be to file an issue describing what's the actual problem (and link this PR as a proposed solution) for discussion

jameshilliard:
> this is a super intrusive change that will very likely break some things down the line.

From my understanding of how PEP 420 namespacing works this is effectively the exact sort of use case it was designed for. What I'm doing probably looks more complicated than it really is, it's essentially just installing modules from 2 packages to the same place.

> I don't see a big upside. what is the problem fixed?

There's a number of issues, I brought up the main ones I could think of off the top of my head in #1371. I think this change will actually allow for simplification of the codebase due to being able to remove all the conditional runtime import logic.

oberstet:
>  it's essentially just installing modules from 2 packages to the same place.

I _think_ PEP420 only covers code imports?

as in: my fear is that this will produce fallout in_downstream_ code build/CI/CD stuff that depends on the current approach.

for example: pip allows to install directly from a github URL - if that repo has 1 python package in a standard structure. 

https://github.com/crossbario/crossbar/blob/218783000b0a52c723820e7100118e7b1cded076/.travis-deploy.sh#L67

if we change to 2 packages (using namespaces) in 1 repo, I'm not sure this would still work.

it will need adjustments definitely. and we have more build/CI/CD like dependencies ^ ..

note that I don't know if this specific pip thing still works with pep420 .. just an example of the kind of issues I would expect (on our side, resulting from such a change)

so it's a big PITA, triggering work on our side, and we don't have time to clean up such fallout now. this is my main itch, even though I can understand why you want it changed like proposed. mmmh ..

jameshilliard:
> I think PEP420 only covers code imports?

Well it allows this sort of approach to work easier by autoimporting I think.

> my fear is that this will produce fallout in_downstream_ code build/CI/CD stuff that depends on the current approach.

Most third party CI/CD is going to use pypi at least which should effectively work the same.

> if we change to 2 packages (using namespaces) in 1 repo, I'm not sure this would still work.

Well since that isn't installing xbr I don't see why it wouldn't work as before.

But if you do need xbr from the master branch then you could install using the [subdirectory flag with git](https://pip.readthedocs.io/en/stable/reference/pip_install/#vcs-support).

> so it's a big PITA, triggering work on our side, and we don't have time to clean up such fallout now. this is my main itch, even though I can understand why you want it changed like proposed. mmmh 

Well since xbr isn't really in production yet from my understanding this would probably be the best time to do this as you shouldn't really have much in the way of third parties depending on xbr, and this should have a fairly low change of causing any major regressions on the autobahn side IMO. I can help fix up any CI/CD regressions if that helps.

PR #1373, opened 2020-04-24
trollius is gone

meejah:
noticed this recently, but trollius is gone (as in we don't use it/care) right?


oberstet:
yes, trollius is gone, as we deprecated python 2 support

PR #1374, opened 2020-04-25
Add environmental variable that strips xbr.

jameshilliard:
This will create distributions and installations without xbr when the env variable `AUTOBAHN_STRIP_XBR` is set.


oberstet:
ok, cool! this looks like it addresses all concerns .. the added (amount of) lines: we can handle that .. not too much hassles/work.

so as you mentioned somewhere, yeah, xbr isn't yet broadly used .. should have little external effect. our own stuff: I don't expect much now .. it's non-intrusive .. let's see;)

thanks for your time and work in contributing!

jameshilliard:
Thanks, by the way it's probably still a good idea to split the contracts out into a separate repo in addition to this change, mostly just so that autobahn downloads are faster as autobahn is a dependency for many projects, that's more of an issue for non-buildroot usages of autobahn(docker images for example) however as buildroot cross compiles packages on a host system and installs to the target image.

Buildroot downloads a large amount of code during a build anyways so the size doesn't make a huge difference on the download size, it's the target installation size that's the big issue, buildroot tightly controls what actually ends up on the target image to keep things small.

oberstet:
>  it's probably still a good idea to split the contracts out into a separate ..

.. package

https://github.com/crossbario/autobahn-python/issues/1375

PR #1376, opened 2020-04-25
Add environmental variable that strips tests.

jameshilliard:
This option is handy for further reducing installation size.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -146,25 +146,19 @@ extras_require_all = extras_require_twisted + extras_require_accelerate + extras
 
 packages = [
     'autobahn',
-    'autobahn.test',
     'autobahn.wamp',
     'autobahn.wamp.gen',
     'autobahn.wamp.gen.wamp',
     'autobahn.wamp.gen.wamp.proto',
-    'autobahn.wamp.test',
     'autobahn.websocket',
-    'autobahn.websocket.test',
     'autobahn.rawsocket',
-    'autobahn.rawsocket.test',
     'autobahn.asyncio',
     'autobahn.twisted',
-    'autobahn.twisted.testing',
     'autobahn.nvx',
-    'autobahn.nvx.test',
     'twisted.plugins',
 ]
 
-package_data = {'autobahn.asyncio': ['./test/*']}
+package_data = {}
 
 entry_points = {
     "console_scripts": [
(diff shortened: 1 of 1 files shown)


oberstet:
sorry, this is exactly what I was afraid of: you are pushing your app specific optimizations upstream. this is not maintainable for us. how about creating an env var for every subset of files? no.

oberstet:
I think the issue that you seem actually trying to address is: "Minimize the _installed_ code in a complete product". If you extend that into every library, this will be lots of work. In any case, this issue is product-specific and IMO doesn't make sense to force each and every library to add knobs to _their_ install infra for the exact set of files your app needs. This should be done in (complete) product packaging, not app packaging, not library packaging.

jameshilliard:
Yeah, this isn't super important but I figured it might be handy to reduce the size a little more. These aren't really app specific optimizations so much as distribution/distro optimizations though.

jameshilliard:
> If you extend that into every library, this will be lots of work.

For the most part this is what is actually what is done for embedded systems as maintaining downstream patches is rather difficult, although we do this for some things such as python itself which currently has [33 downstream patches](https://github.com/buildroot/buildroot/tree/master/package/python3) to aggressively reduce the installation size. Python's setuptools build system is just not that great in general when it comes to handling optional feature installation.

meejah:
If you're worried about size, you maybe shouldn't install any `.py` files at all, just the `.pyc` files. I do agree w/ @oberstet here, though: we can't maintain things that very different use-cases need.

Debian requires tests to be installed, for example, and anyone who downloads the tarball expecting to "get Autobahn" would be surprised if it didn't have any tests at all.

When you mention that "maintaining downstream patches is rather difficult" think about our side: now when something breaks with tests or the installer we have dozens more lines to deal with -- there is maintenance burden for us too.

Maybe a useful way to proceed would be if you shared your scripts to strip / reduce / etc the build -- other embedded platforms might benefit by using this as a base for their own stripping.

jameshilliard:
> If you're worried about size, you maybe shouldn't install any .py files at all, just the .pyc files.

Already been doing that actually(it's a buildroot build config option that gets applied to all python packages essentially).

> Debian requires tests to be installed, for example, and anyone who downloads the tarball expecting to "get Autobahn" would be surprised if it didn't have any tests at all.

Yeah, that's why I made it so that this could be applied at install time rather than at dist build time.

> When you mention that "maintaining downstream patches is rather difficult" think about our side: now when something breaks with tests or the installer we have dozens more lines to deal with -- there is maintenance burden for us too.

Well these sorts of things are usually easier to maintain upstream than downstream, it's a burden for both I agree, it's just a lot harder to do downstream.

> Maybe a useful way to proceed would be if you shared your scripts to strip / reduce / etc the build -- other embedded platforms might benefit by using this as a base for their own stripping.

These are generally just part of the buildroot project, I try to upstream all change that apply to the base buildroot project so that I don't have to maintain my own fork.

PR #1378, opened 2020-04-30
Removed externalPort assignment when not set

### LINKED ISSUE #1360: Consider enabling strict externalPort optionally
Hello,
I had noticed that autobahn strictly enforces the RFC https://tools.ietf.org/html/rfc7230#section-5.4. As far as I can tell(I could be wrong) the externalPort isn't used anywhere else apart from the check at https://github.com/crossbario/autobahn-python/blob/master/autobahn/websocket/protocol.py#L2611-L2633. 
The problem arises when this is used inside a system such as Kubernetes, where there's an optional SSL termination at edge, all requests are routed via an Ingress node. This means all users have to set the external port to 80. (Even though outside it's 443). 
I found the issue from https://github.com/RobotWebTools/rosbridge_suite/pull/468, by internally switching the library, it broke compatibility with no way to maintain backward compatibility. 
Would you consider making the externalPort check non-default and having a strict enforce mode?

pramodhkp:
I went ahead and created a PR that fixes #1360.  We needed it to fix something on our side.

Let me know if it makes sense/ if there are any other changes that's required to get it merged.


pramodhkp:
Hey @oberstet , were you able to give this some thought?

oberstet:
hey, thanks for contributing!

ok, so I had a closer look .. couple of comments below. if you have a chance to modify the PR, I'd merge. this is a deeply buried, subtle change, but I think this should be good.

1) we can simplify the code 

https://github.com/crossbario/autobahn-python/blob/0e1983826a734d0b0adf7d190d4808dcc95a0e99/autobahn/websocket/protocol.py#L3198

by replacing this whole block with just

```python
self.externalPort = externalPort
```

2) we should change the comment

``` 
# do port checking only if externalPort or URL was set
``` 

to

```
# do port checking only if externalPort was set
```

3) and finally, I think it also makes sense to remove the `else` block here

https://github.com/crossbario/autobahn-python/blob/0e1983826a734d0b0adf7d190d4808dcc95a0e99/autobahn/websocket/protocol.py#L2627

---

all of this should render the external port checking optional, and only apply when `externalPort` was indeed set.

oberstet:
great! merging this, since

"autobahn/wamp/serializer.py:314:135: E225 missing whitespace around operator"

is unrelated (flake8 itself changed a bit ..), and is fixed on the upcoming https://github.com/crossbario/autobahn-python/pull/1382

pramodhkp:
thank you!

pramodhkp:
@oberstet hey, one question. when is the release for this PR scheduled for?

pramodhkp:
@oberstet could you please give me an ETA for this, if possible? ^

oberstet:
v20.5.1 should be ready to shi till sunday latest .. and this will have above change

PR #1400, opened 2020-07-07
Fix xbr strip for contracts

jameshilliard:
Make package data installation conditional for xbr contracts and move `MANIFEST.in` package data installation into setup.py.

Seems I hadn't properly tested `AUTOBAHN_STRIP_XBR` against an autobahn distribution containing the xbr contracts which were actually being installed by `MANIFEST.in` and not `package_data['xbr'] = ['./xbr/contracts/*.json']` as I had assumed which I had copied from [here](https://github.com/crossbario/autobahn-python/blob/v20.3.1/setup.py#L254). Lets just eliminate `MANIFEST.in` and move all package data installation into setup.py.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,0 @@
-include LICENSE
-include autobahn/nvx/_utf8validator.c
-recursive-include autobahn/wamp/gen/schema *
-recursive-include autobahn/xbr/contracts *
--- a/setup.py
+++ b/setup.py
@@ -164,7 +164,12 @@ packages = [
     'twisted.plugins',
 ]
 
-package_data = {'autobahn.asyncio': ['./test/*']}
+package_data = {
+    '': ['LICENSE'],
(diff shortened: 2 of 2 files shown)


oberstet:
fwiw, I think removing the MANIFEST.in will also remove it from source distributions (not sure .. there was a reason why both is included .. forgot what exactly).

jameshilliard:
It seems like everything was still ending up in the source distribution from what I could tell.

jameshilliard:
According to the [documentation](https://packaging.python.org/guides/using-manifest-in/#how-files-are-included-in-an-sdist) `package_data` is included in source distributions.

oberstet:
won't fix

jameshilliard:
@oberstet Was there an issue with this fix you were seeing? Without this the install size for a non-xbr install of autobahn from pypi is many times larger than it should be, by around 8.8MB last I checked.

Issue #1322, opened 2020-02-20
Specific type check behavior


oberstet:
consider this registered procedure:

```python
    @wamp.register('network.xbr.console.get_member_logins', check_types=True)
    def get_member_logins(self, member_oid: bytes, details: Optional[CallDetails] = None) -> list:
```

when this procedure is called with **2 (!)** arguments, the resulting traceback is:

```
2020-02-20T21:59:49+0100 [Router      19483] XbrNetworkApi.onUserError(): "AttributeError: __name__"
Traceback (most recent call last):
  File "/home/oberstet/scm/crossbario/autobahn-python/autobahn/wamp/protocol.py", line 1020, in onMessage
    on_reply = txaio.as_future(endpoint.fn, *invoke_args, **invoke_kwargs)
  File "/home/oberstet/scm/crossbario/txaio/txaio/tx.py", line 365, in as_future
    return ensureDeferred(fun(*args, **kwargs))
  File "/home/oberstet/cpy381_3/lib/python3.8/site-packages/twisted/internet/defer.py", line 911, in ensureDeferred
    return _cancellableInlineCallbacks(coro)
  File "/home/oberstet/cpy381_3/lib/python3.8/site-packages/twisted/internet/defer.py", line 1529, in _cancellableInlineCallbacks
    _inlineCallbacks(None, g, status)
--- <exception caught here> ---
  File "/home/oberstet/cpy381_3/lib/python3.8/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/home/oberstet/scm/crossbario/autobahn-python/autobahn/wamp/protocol.py", line 508, in _type_check
    "'{}' required={} got={}".format(name, kind.__name__, type(arguments[name]).__name__))
  File "/home/oberstet/cpy381/lib/python3.8/typing.py", line 757, in __getattr__
    raise AttributeError(attr)
builtins.AttributeError: __name__
```

this is super-obscure and unexpected ..

when turning off type checks, it will run into the assert within the procedure body:

```
2020-02-20T22:02:07+0100 Traceback (most recent call last):
  File "/home/oberstet/cpy381_3/lib/python3.8/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/oberstet/cpy381_3/lib/python3.8/site-packages/twisted/internet/defer.py", line 1475, in gotResult
[...]

Issue #1332, opened 2020-02-29
TypeError: Subscripted generics cannot be used with class and instance checks


oberstet:
**Code:**

```python
    @wamp.register('network.xbr.console.find_markets', check_types=True)
    def find_markets(self,
                     created_from: Optional[int] = None,
                     limit: Optional[int] = None,
                     include_owners: Optional[List[bytes]] = None,
                     include_actors: Optional[List[bytes]] = None,
                     include_titles: Optional[List[str]] = None,
                     include_descriptions: Optional[List[str]] = None,
                     include_tags: Optional[List[str]] = None,
                     include_apis: Optional[List[bytes]] = None,
                     details: Optional[CallDetails] = None) -> List[bytes]:
        pass
```

**Traceback:**

```
2020-02-29T22:27:08+0100 [Router      26746] XbrNetworkApi.onUserError(): "TypeError: Subscripted generics cannot be used with class and instance checks"
Traceback (most recent call last):
  File "/home/oberstet/scm/crossbario/autobahn-python/autobahn/wamp/protocol.py", line 1020, in onMessage
    on_reply = txaio.as_future(endpoint.fn, *invoke_args, **invoke_kwargs)
  File "/home/oberstet/scm/crossbario/txaio/txaio/tx.py", line 365, in as_future
    return ensureDeferred(fun(*args, **kwargs))
  File "/home/oberstet/cpy381_5/lib/python3.8/site-packages/twisted/internet/defer.py", line 911, in ensureDeferred
    return _cancellableInlineCallbacks(coro)
  File "/home/oberstet/cpy381_5/lib/python3.8/site-packages/twisted/internet/defer.py", line 1529, in _cancellableInlineCallbacks
    _inlineCallbacks(None, g, status)
--- <exception caught here> ---
  File "/home/oberstet/cpy381_5/lib/python3.8/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/home/oberstet/scm/crossbario/autobahn-python/autobahn/wamp/protocol.py", line 506, in _type_check
    if not isinstance(arguments[name], kind.__args__):
  File "/home/oberstet/cpy381/lib/python3.8/typing.py", line 766, in __instancecheck__
    return self.__subclasscheck__(type(obj))
  File "/home/oberstet/cpy381/lib/python3.8/typing.py", line 774, in __subclasscheck__
    raise TypeError("Subscripted generics cannot be used with"
[...]

PR #1405, opened 2020-07-19
Type check improve

om26er:
Fixes https://github.com/crossbario/autobahn-python/issues/1322


om26er:
I also have some ideas about https://github.com/crossbario/autobahn-python/issues/1332, will come up with a solution to that in a different PR. It may make sense to move this helper out into a separate module, so that it can be tested easily.

PR #1411, opened 2020-08-18
cryptosign authextra allow arbitrary keys

om26er:
Fixes https://github.com/crossbario/autobahn-python/issues/1412


oberstet:
yeah, should be good. merging. for background: I am generally following a conservative approach, that is, explicitly check and disallow anything _not_ strictly required. but I can't see much potential for harm or maintenance headaches down the line in this case ..

PR #1426, opened 2020-11-15
wamp idl

oberstet:


oberstet:
so for example, this wamp interface https://github.com/crossbario/xbr-app/blob/master/schema/climate.fbs can be compiled with flatc to a binary schema file, which in turn can be loaded using the helpers from this PR: https://gist.github.com/oberstet/905c140969c71fdfa8be84ac9994e40b

oberstet:
https://gist.github.com/oberstet/57723ff4bb07c2ea0b71a67ef3e5bd5d

PR #1427, opened 2020-11-16
Remove meaningless assertions

sergio-bershadsky:


oberstet:
what do you mean with "meaningless"?

sergio-bershadsky:
There is no any case when `*args` or `**kwargs`  will be equal to None or it will no be type of tuple or dict.

oberstet:
> There is no any case when *args or **kwargs will be equal to None or it will not be type of tuple or dict.

that is exactly what the assert test for

oberstet:
aah, you mean

```
 def progress(*args, **kwargs):
```

in this specific function: true

for the others: not true

sergio-bershadsky:
Please follow python manual. *[anything] are always tuple (empty to filled) so **[anything] are always dicts to need to check it

sergio-bershadsky:
This "meaningless" touches all editing I sent

PR #1429, opened 2020-11-24
Allow transport_factory default options modification in AppRunner

sestebanz:
Hi Autobahn,

This a new solution to https://github.com/crossbario/autobahn-python/pull/912 and https://github.com/crossbario/autobahn-python/issues/838:

Adding kwargs to run method allow to modify default transport options:

`runner = ApplicationRunner(url, realm=realm, extra=extra)`
`coro = runner.run(make, start_loop=False, openHandshakeTimeout=10, autoPingSize=5)`

Using a configuration method as suggested  in #912 is not possible because the transport factory is created inside the run method. This commit allows to pass parameters to the transport factory in a seamless way.

Thanks,

Santi

### DIFF
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -160,8 +160,8 @@ class ApplicationRunner(object):
             start a new asyncio loop.
         :type start_loop: bool
 
-        :param kwargs: Key value list of arguments to override default 
-            websocket transport factory options. Allowed parameters are 
+        :param kwargs: Key value list of arguments to override default
+            websocket transport factory options. Allowed parameters are
             defined in :func:`autobanh.websocket.interfaces.IWebSocketClientChannelFactory.setProtocolOptions`
             See the docs: https://autobahn.readthedocs.io/en/latest/websocket/programming.html#websocket-options
 
--- a/autobahn/asyncio/wamp.py
+++ b/autobahn/asyncio/wamp.py
@@ -223,7 +223,7 @@ class ApplicationRunner(object):
                                                  autoPingSize=4,
                                                  perMessageCompressionOffers=offers,
                                                  perMessageCompressionAccept=accept)
-            # update defaults 
+            # update defaults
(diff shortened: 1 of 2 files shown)


oberstet:
@meejah do you have an opinion? do we want new public API, and if so, "blanko" .. that is forwarding anything? I would argue we should probably use explicit options forwarding to transport .. maybe

meejah:
Oh, hrm .. yeah I think I meant something a little different in that #912 comment but didn't follow-up :/ 

So you still have to subclass ApplicationRunner to get custom options, which isn't great.

(What I think I meant originally was to have a `set_transport_options(self, **kwargs)` that would remember all the `kwargs` and then pass them to the factory whenever it creates that).

Functionally, passing on transport options in `run()` is pretty similar .. and maybe even makes the most sense, since there will be one transport-factory created per call to `run()` (that said, of course you're only really supposed to call `run()` once).

I'm pretty neutral on whether it's `**kwargs` or explicit .. the latter has the downside of possibly being out of sync.

This PR at least needs to adjust the docstring so a caller knows they can pass transport options like that (and ideally a link to the list of them .. basically everything https://github.com/crossbario/autobahn-python/blob/master/autobahn/websocket/protocol.py#L3236 takes so you should be able to link in the docs to https://github.com/crossbario/autobahn-python/blob/master/autobahn/websocket/interfaces.py#L129 )

sestebanz:
I have added the documentation as suggested. I hope it is OK.

oberstet:
> I'm pretty neutral on whether it's **kwargs or explicit .. the latter has the downside of possibly being out of sync.

fwiw, I am -1 on adding a "catch all", duck typed `**kwargs`. IMO a better way could be:

1. create a new `autobahn.wamp.types.ProtocolOptions` class
2. add all attributes from `setProtocolOptions` to that class
3. take an optional `protocol_options: Optional[ProtocolOptions]=None` to both async and twisted `ApplicationRunner.run`

oberstet:
as said, we don't want a generic `**kwargs` catch all

Issue #1430, opened 2020-11-27
nicely_exit doesn't exit nicely with python >=3.7


jacobanana:
From Python 3.7, `asyncio.Task.all_tasks()` has been moved to `asyncio.all_tasks()` and `asyncio.Task.current_task()` has moved to `asyncio.current_task()`

Therefore, `nicely_exit` of [autobahn/asyncio/component.py](https://github.com/crossbario/autobahn-python/blob/df834083bd8740eec7c7cb0cdfc57639d6f54145/autobahn/asyncio/component.py#L363
) doesn't exit so nicely anymore

```python
    @asyncio.coroutine
    def nicely_exit(signal):
        log.info("Shutting down due to {signal}", signal=signal)

        tasks = asyncio.Task.all_tasks()
        for task in tasks:
            # Do not cancel the current task.
            if task is not asyncio.Task.current_task():
                task.cancel()
```
and throws the following exception

`AttributeError: type object '_asyncio.Task' has no attribute 'all_tasks'`

Python >= 3.7
Autobahn 20.7.1

PR #1431, opened 2020-11-27
fix nicely_exit with python >= 3.7

jacobanana:
This proposes a fix for #1430 that should be backwards compatible with python < 3.7


oberstet:
LGTM, thanks for contributing!

PR #1433, opened 2020-12-11
fix event handler removal

0xfadeef:
Linked to 7f1bdacaf3df7460659dfcdbc1741a2ea7743c02

### DIFF
--- a/autobahn/util.py
+++ b/autobahn/util.py
@@ -797,7 +797,10 @@ class ObservableMixin(object):
                 if handler is None:
                     del self._listeners[event]
                 else:
-                    self._listeners[event].discard(handler)
+                    try:
+                        self._listeners[event].remove(handler)
+                    except ValueError:
+                        pass
 
     def fire(self, event, *args, **kwargs):
         """


oberstet:
@meejah any reason for the change https://github.com/crossbario/autobahn-python/commit/7f1bdacaf3df7460659dfcdbc1741a2ea7743c02 ? why not keep "set"? a handler should _not_ be added twice ..

meejah:
Hmmm, too bad the commit doesn't explain *why* ;) but I think it might be because you can't change a set while iterating it...? (but you can with a list). So, e.g., if you're notifying some listeners and the handler does something (like "remove me as listener") then it'll fail as a set

0xfadeef:
AFAIK it's never safe to mutate a dataset while iterating over it (be it a `set` or a `list`). The common approach is to iterate over a copy. Since we don't usually expect a _huge_ amount of handlers, wrapping handlers set to a `list` shouldn't be that expensive.

On Py 3.8: 
```
>>> ll = [1, 2, 3]
>>> for item in ll:
...     ll.remove(item)
... 
>>> ll
[2]
```

0xfadeef:
@oberstet @meejah Any chance we could return to discussion? The `ObservableMixin.off()` method is virtually broken in the current implementation and the service I'm working at right now relies heavily on this functionality.

Bottom line, this PR merely fixes the bug, stayed in the codebase for more than four years. The `list` vs `set` dilemma is the whole different topic.

oberstet:
I'm gonna merge this now: it fixes an error resulting from a wrong change. also, we don't use this anyways (eg in crossbar). The observer thing should use set, not list. eg currently, it allows to add the same callback multiple times (as it is a list). it should only allow one (which is why it was a set). when it does change again, we'll change the code here back.

PR #1438, opened 2020-12-15
deletes deleted example from README.md

benjaminbecker:
The example file `examples/ayncio/websocket/echo/client_coroutines_py2.py` has been deleted. I does not make sense anymore to refer to it in the README.


oberstet:
thanks for the correction!

Issue #1370, opened 2020-04-24
Component stop() may fail dependent on goodby message


ickby:
I'm using autobahn as client on a golang nexus router. I use the component API with asyncio. When I call Component.stop() it should quit the connection and not reconnect. That does work with crossbar router, however, not with nexus.  For nexus it results in an error and tries to reconnect.

The reason for that is, that nexus replies to goodby with "wamp.close.goodbye_and_out".  However, autobahn explicitly wants to see "wamp.close.normal" [here](https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/component.py#L770). If I change this code to also accept goodbye_and_out everthing works fine. 

I looked into the WAMP specification, and there is a[n example](https://wamp-proto.org/_static/gen/wamp_latest.html#session-closing) given explicitly using "goodbye_and_out" for session closing, hence I think this is a autobahn bug. (Why is the URI relevant anyway, isn't it just the message type that is important?)

ickby:
Any update on this issue? Seems to be a clear violation of specification and hence should be fixed.

ickby:
Thanks, this works very well now!

PR #1450, opened 2021-01-19
consider 'wamp.close.goodbye_and_out' a clean exit

meejah:
See #1370

### DIFF
--- a/autobahn/wamp/component.py
+++ b/autobahn/wamp/component.py
@@ -767,7 +767,7 @@ class Component(ObservableMixin):
                         details=details,
                     )
                     if not txaio.is_called(done):
-                        if details.reason in ["wamp.close.normal"]:
+                        if details.reason in ["wamp.close.normal", "wamp.close.goodbye_and_out"]:
                             txaio.resolve(done, None)
                         else:
                             f = txaio.create_failure(
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,7 @@ master
 ------
 
 * fix: add missing jinja2 dependency for XBR CLI (#1447)
+* fix: ``wamp.close.goodbye_and_out`` counts as a clean exit (#1370)
 
 20.12.3
 -------
--- a/tox.ini
+++ b/tox.ini
@@ -187,7 +187,7 @@ extras =
 commands =
     python -V
     sphinx-build --version
-    python -c "from autobahn import xbr; print('HAS_XBR={:}'.format(xbr.HAS_XBR))"
+    python -c "from autobahn import xbr; print('HAS_XBR=' + str(xbr.HAS_XBR))"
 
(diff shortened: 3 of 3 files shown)


oberstet:
looks good .. not sure why CI is sad .. some weird "substitution blabla" error?

meejah:
Weird .. I ran `tox -e flake8` successfully locally .. and just tried pushing "re-run all jobs" in case it was some transient / version issue but it seems to have failed again in the same way

meejah:
(...and it's hard to imagine a *simpler* change than this PR ...)

meejah:
ah, I had `tox 3.19.0` locally before; just upgraded (and got 3.21.2) and it fails similarly...

meejah:
WAT?! It appears to be this line in the config (but it fails even before building the tox environment, so it's not *running* the line that's confusing tox, but merely *parsing the config*): https://github.com/crossbario/autobahn-python/blob/master/tox.ini#L190

meejah:
(I'm pushing a change commenting-out that line just to see if the CI runs .. flake8 at least runs locally after that change)

oberstet:
ok, weird!

> maybe it chokes on the `:` in `python -c "from autobahn import xbr; print('HAS_XBR={:}'.format(xbr.HAS_XBR))"` ? not sure .. forgot what it does here anyways;)

meejah:
AAaaaaaaarrg, it seems that Tox figures it should look for `{}` anywhere at all and try to do .. substitutions? I guess .. so the "fix" in that last commit works, but also: WTF.

oberstet:
yes, agreed, wtf! so I guess the take away is: be careful with literal code in tox ini files

meejah:
I think it was probably this: https://github.com/tox-dev/tox/issues/1711 ... but also makes me wonder if that code was EVER doing what was intended, because I suspect Tox was replacing the `{}` with the os path-separator and so it never really did `HAS_XBR=True/False` correctly?

oberstet:
wau. the issue you link: yeah, that looks like it. well. doing some silent substitutions on the literal lines in tox files .. is a bad idea. and yes

> makes me wonder if that code was EVER doing what was intended

indeed. it might never have actually worked as intended .. even though it did not break. mmh, well.

oberstet:
@fijal rgd change https://github.com/crossbario/autobahn-python/commit/a468e5724b588ba51e3bbcad42317b906ac0606b#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449 was that intentional, related to above?

fijal:
So one part is because CI complains about {} vs {:} (and fails flake8)

The second one is printing full information about the error as opposed to just Exception.

oberstet:
thanks! merged it .. lets see https://github.com/crossbario/autobahn-python/actions/runs/510649942

PR #1469, opened 2021-03-02
fix workaround in AB for a bug in twisted that by being fixed in tx 2…

oberstet:
…1.21, blows up the workaround

### DIFF
--- a/autobahn/twisted/websocket.py
+++ b/autobahn/twisted/websocket.py
@@ -242,7 +242,7 @@ class WebSocketAdapterProtocol(twisted.internet.protocol.Protocol):
         # the peer we are connected to
         try:
             self.peer = peer2str(self.transport.getPeer())
-        except AttributeError:
+        except (AttributeError, NotImplementedError):
             # ProcessProtocols lack getPeer()
             self.peer = 'process:{}'.format(self.transport.pid)
         self.peer_transport = 'websocket'
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,11 @@
 Changelog
 =========
 
+21.3.1
+------
+
+* fix: Twisted v21.2.0 breaks Crossbar.io (see https://github.com/crossbario/crossbar/issues/1864)
+
 21.2.2
 ------
 


codegastudio:
Hello,
Do you think it's possible to have the fix in a 21.2.1 to be able to get it with a python version 3.6 please ?

codegastudio:
cause this mean more job on this project to level up to 3.7 or more and this is not yet on the roadmap :/

oberstet:
> cause this mean more job on this project to level up to 3.7 or more and this is not yet on the roadmap :/

I'm sorry, but offloading work doesn't work;) what roadmap? upgrade to cpython 3.9.2 and use the latest autobahn and twisted is my recommendation. it's easy shit.

codegastudio:
that mean for us to rebuild opencv to generate a new python lib for this version, update some steps of our CI to update preinst/postinst debian package script and rebuild/validate the 36 projects dependencies which uses it just for a missing exception.

codegastudio:
if there is no possibility to get the fix in 21.2.1, so we will need to do it but as i said it was not plan yet to take resources to update our python version while we are developing features.

oberstet:
possible: sure, it is possible, but neither a priority for us nor sth we want to do (maintain multiple branches). you can of course also fork autobahn and maintain the branch yourself. might be less work in your situation. or you can throw money at us to do it. maybe, depends. point is: why would we spend our resources for free on sth that only you need for - I assume - some commercial closed source product you work on? strange idea;)

codegastudio:
Their are nothing special to say for that.
You just provide a solution. 
And as many software you use a semver notation to increase feature/fix.
So basically i guess than you was able to deliver this fix for the impacted version.
Most of all when this implies for the community to evolve their python version while the 3.6 end of life is in few months.

I am a little disappointed cause this crash could have been detected with really basic config.json. 
```
{
    "version": 2,
    "workers": [
        {
            "type": "router"
        }
    ]
}
```
This have some consequences and i guess not only for our project.
Also this coming in the same time than this hot subject with [cryptography](https://github.com/pyca/cryptography/issues/5771) package which come in dependencies of crossbar package.
So after spend time to find a way to fix cryptography package witch require RUST and is not ready to run out of the box on ARMV8 arch, we get this new bottleneck.

But this is the way...

Get it just as a user feedback !

oberstet:
> And as many software you use a semver notation to increase feature/fix.

we are using calver, not semver

> So basically I guess you could have provided this fix for the affected version.

no, we could not, because we don't have the resources

> Also this coming in the same time than this hot subject with cryptography package which come in dependencies of crossbar package.

yeah, we ran into that a couple of weeks ago: this one is easy as well, just use upstream rust, not outdated distro packages. that's what we do for arm64 eg for our yocto builds, or for docker 

Bildschirmfoto von 2021-03-10 16-17-09

PR #1479, opened 2021-03-25
expand wamp auth scram and xbr argon2/hkdf

oberstet:


oberstet:
this PR is supposed to add all the helper functions we need for this:

Bildschirmfoto von 2021-03-26 09-48-28

autobahn/wamp/auth.py
```diff
@@ -599,3 +605,69 @@ def compute_wcs(key, challenge):
...
+    * time cost ``4096``
+    * memory cost ``512``
+    * parallelism ``1``
+
+    See `draft-irtf-cfrg-argon2 <https://datatracker.ietf.org/doc/draft-irtf-cfrg-argon2/>`__ and
+    `argon2-cffi <https://argon2-cffi.readthedocs.io/en/stable/>`__.
+
+    :param email: User email.
+    :param password: User password.
+    :param salt: Optional salt to use (must be 16 bytes long). If none is given, compute salt
```
meejah, line 625: Actually, re-reading some more SCRAM docs I think we can just get rid of this and always use `os.urandom(16)` or so .. the "hash_data" returned from the Argon2 hasher encodes the salt into it .. that is, you get strings back like `'$argon2id$v=19$m=102400,t=2,p=8$tSm+JOWigOgPZx/g44K5fQ$WDyus6py50bVFIPkjA28lQ'` which in this case has salt of `tSm+JOWigOgPZx/g44K5fQ`

That said, I don't know if it's "bad" to have a predictable salt based on email address .. maybe that still counts as "random" for the purposes of salting (after all, it's a public value), but also I'm not a cryptographer that could re-do the Argon security-analysis with that assumption instead so it certainly seems safer to use cryptographic-random salt (i.e. `os.urandom(16)`) and I don't think that has any downsides here.

(Posting as a bare comment, haven't reviewed the whole PR yet)

oberstet: > "hash_data" returned from the Argon2 hasher encodes the salt into it

oh, great! if that is the case, then of course we can use a random salt and ship it around ..

meejah: The salt is encoded / shipped around in the credential already https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/auth.py#L669 right? So I think all that's needed is to make the salt always be `os.urandom(16)` if not provided (instead of derived from email).

oberstet: one aspect I just refigured: the premise for this auth flavor is that I only retain email + password, so I don't have the salt. while the salt is shipped around technically, the question is if we want to have a salt that is actually not a function of the information we already have. there are only 2: password and email. hence ... not sure, what do you think?

meejah: Hmmmm, right. I think a "predictable" salt is fine as long as it is different for each hashed password .. I'm not *100% sure* that you can't still use the predictability here to speed up cracking the database though. (It'll still be better than fixed or no salt though).

meejah: A real cryptographer I spoke with says salts should be cryptographically-random (and not derived via hash from email or anything else).

Issue #1489, opened 2021-06-06
Mixed case is less compatible.


ioquatix:
[RFC6455](http://tools.ietf.org/html/rfc6455) always uses `Upgrade: websocket` as value, but [Section 11.2](http://tools.ietf.org/html/rfc6455#section-11.2) registers WebSocket as Upgrade keyword.

Unfortunately, no RFC specifies that the upgrade header should be case insensitive. However, there was some discussion here https://github.com/httpwg/http-core/pull/131 which is currently unpublished (unofficial guidance at best).

For maximum compatibility, following the examples given in the RFC makes sense to me. AFAIK, most browsers follow this spec (lower case token) and most clients also use the lower case form.

https://github.com/crossbario/autobahn-python/blob/fd5530cecff0be210b148bddf1e984ed31d2863c/autobahn/twisted/resource.py#L125

ioquatix:
I've created a PR to check if CI passes or not.

oberstet:
First of all, thanks for the report and your work (PR)!

Rgd the specific question: HTTP headers (the keys) are case insensitive (this is in some HTTP RFC .. not 6455), always.

Hence, an implementation that relies on HTTP header key case is broken. And we have a policy of "fail fast" instead of "butch, guess and work around around broken peers". IOW: won't fix, sorry:(

PR #1490, opened 2021-06-06
Prefer lower case `websocket` upgrade header.

ioquatix:
Potential fix for #1489.


oberstet:
thx, but pls see https://github.com/crossbario/autobahn-python/issues/1489#issuecomment-861391942

PR #1492, opened 2021-06-12
Fix duplicate xbr packages and xbr templates package_data.

jameshilliard:
Fix duplicate packages introduced in #1479.

Fix xbr package_data templates include path and de-duplicate include.


oberstet:
ah right, sth got mixed up .. thanks!

PR #1493, opened 2021-06-13
Build with nvx by default and don't publish universal wheel.

jameshilliard:
Per [docs](https://pythonwheels.com/):
> Warning: If your project has optional C extensions, it is recommended not to publish a universal wheel, because pip will prefer the wheel over a source installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,13 +29,13 @@ install:
 
 build:
 	-rm -f dist/*
-	# python setup.py sdist bdist_wheel --universal
-	AUTOBAHN_USE_NVX=1 python setup.py sdist bdist_wheel
+	# AUTOBAHN_USE_NVX=0 python setup.py sdist --universal
+	AUTOBAHN_USE_NVX=1 python setup.py sdist
 	ls -la dist
 
 # upload to our internal deployment system
 upload: clean
-	python setup.py sdist bdist_wheel --universal
+	AUTOBAHN_USE_NVX=0 python setup.py sdist --universal
 	aws s3 cp --acl public-read \
 		dist/autobahn-*.whl \
 		s3://fabric-deploy/autobahn/
@@ -69,7 +69,7 @@ clean:
 
 # publish to PyPI
 publish: clean
-	python setup.py sdist bdist_wheel --universal
+	AUTOBAHN_USE_NVX=0 python setup.py sdist --universal
 	twine upload dist/*
 
 clean_docs:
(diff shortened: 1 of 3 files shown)


oberstet:
ok, but what's the problem? did you run into any? if there is none, I wouldn't change it from an universal wheel -  pls see my comment https://github.com/crossbario/autobahn-python/pull/1493/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L114 - also: NVX is really still alpha anyways .. I didn't find time to finish that at the time.

jameshilliard:
> ok, but what's the problem? did you run into any?

Mostly just saw your comment and figured it should be a fairly straight forward to fix it.

> I wouldn't change it from an universal wheel

Well universal wheels at least only work correctly with pure python modules from my understanding and shouldn't be uploaded if anyone might want to build the extension module.

> NVX is really still alpha anyways .. I didn't find time to finish that at the time.

Is it buggy or just incompatible with some archs? I figured it was just incompatible with some but making the extension optional makes it fall back to using the pure python version if there is a build failure.

oberstet:
> Well universal wheels at least only work correctly with pure python modules from my understanding and shouldn't be uploaded if anyone might want to build the extension module.

yes, ideally the user should be able to choose (python only bare library or library with extensions like "compression" or "nvx"). the bare library (when not specifying any dist flavor like "nvx") _is_ universal (as it is python only etc). because of that, I thought it would make sense to mark the package "universal". it is universal however only when not specifying any of the dist flavors. this holds true even for other dist flavors (eg I think "compression" also requires native extension code).

so actually I don't know for sure if ^ is the "right" or encouraged way of doing things ... is it not?

should a python package be marked "universal" even when it is only when _not_ specifying any dist flavor?

> Is it buggy or just incompatible with some archs? I figured it was just incompatible with some but making the extension optional makes it fall back to using the pure python version if there is a build failure.

good questions! unfortunately, too long ago;) I mean, I forgot. I would have expected it to fall back in any case to a pure python implementation (or a "non-NVX" one for that matter ..).

jameshilliard:
> so actually I don't know for sure if ^ is the "right" or encouraged way of doing things ... is it not?

Well from my understanding binary wheels should only be provided with pre-built c extension binaries, so that the fallback to building extensions from source works properly if there is not a compatible pre-built wheel.

> should a python package be marked "universal" even when it is only when not specifying any dist flavor?

My reading is that we should not upload universal wheels here at all, the sdist is what would get used when there's not a pre-built c extension binary available which is what we want I think.

> good questions! unfortunately, too long ago;) I mean, I forgot. I would have expected it to fall back in any case to a pure python implementation (or a "non-NVX" one for that matter ..).

Yeah, so this should at least properly handle falling back to pure python automatically if there's a build failure of any sort...but would use the c extension opportunistically. If the issue is a runtime failure and not a build failure however...we might not want to enable opportunistically.

Issue #1499, opened 2021-11-14
Refactor SigningKey


om26er:
Create an interface class for the "signing keys", one of the implementation for that would be the existing SigningKey class.

The other implementation could be a Vendor specific hardware wallet.

om26er:
Fixed via https://github.com/crossbario/autobahn-python/pull/1500

PR #1500, opened 2021-11-14
refactor SigningKey class

om26er:
This is an attempt at fixing https://github.com/crossbario/autobahn-python/issues/1499, however I am putting it out to get feedback if this is the right approach.

This PR adds
* `SigningKeyBase` class, which abstracts out a few methods from the existing `SigningKey` implementation for reusability.
* `ISigningKey` interface, which only defines 4 methods. (missing doc string, will add once this looks "reasonable").

### DIFF
--- a/autobahn/wamp/cryptosign.py
+++ b/autobahn/wamp/cryptosign.py
@@ -463,7 +463,6 @@ if HAS_CRYPTOSIGN:
             # the signature
             return txaio.create_future_success(sig.signature)
 
-
     @util.public
     class SigningKey(SigningKeyBase):
         """
@@ -596,7 +595,6 @@ if HAS_CRYPTOSIGN:
                 key = signing.VerifyKey(keydata)
 
             return cls(key, comment)
-
 
     ISigningKey.register(SigningKey)
 
--- a/autobahn/wamp/cryptosign.py
+++ b/autobahn/wamp/cryptosign.py
@@ -30,6 +30,7 @@ import struct
 import txaio
 
 from autobahn import util
+from autobahn.wamp.interfaces import ISigningKey
 from autobahn.wamp.types import Challenge
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


oberstet:
yes, this approach looks good! couple of comments:

* rgd the interface: yes, keeping it minimal is great.
* the `sign` method obviously needs the `data` argument and Future(binary) return types.
* The `sign_challenge` method is not strictly about Ed25519 key signatures, but the specific way WAMP-cryptosign wants it signature formatted during the WAMP authentication handshake. One might argue therefor, the method should not be part of the ifc, and the code probably somewhere else (eg in a free standing function in the same file). 
* there will be a need for dynamically constructing objects of classes implementing above ifc given a CB node config portion. that code (the dyn factory) could be in CB then, and use the constructors of the new refactored classes (one for file-based, one for HW-based).

om26er:
> * the `sign` method obviously needs the `data` argument and Future(binary) return types.

done

> * The `sign_challenge` method is not strictly about Ed25519 key signatures,

we could do that, we'll need to pass the SigningKey to that method in parameters as it internally calls `self.sign`. Shall I do that ? 

> * there will be a need for dynamically constructing objects of classes implementing above ifc given a CB node config portion. that code (the dyn factory) could be in CB then, and use the constructors of the new refactored classes (one for file-based, one for HW-based).

Will get to that one, after this lands in some form. We may need to make further change in AB.



> * CI only complains about flake .. did the WAMP-cryptosign tests run successfully .. not sure, but I think there is some minimal testing?

RLinks are working for me with the new change, so this is mostly good. The CI passes now as well.

oberstet:
> we could do that, we'll need to pass the SigningKey to that method in parameters as it internally calls self.sign. Shall I do that ?

yes, right. In that case, better would be a new separate free standing function `format_challenge` or similar, which only takes the Challenge and returns bytes - which then must be provided to `SigningKey.sign`

> RLinks are working for me with the new change, so this is mostly good. The CI passes now as well.

=) great work!

pls merge this PR as you see fit .. eg directly after ^ format_challenge

oberstet:
rgd the ifc, just noticed:

`def sign(self, data: bytes) -> bytes:`

actually, this returns an asyncio/Twisted Future/Deferred .. either of those depending on which network framework is in place.

further, I am pretty sure that we will need to make the method itself async too

the reason is: for a HW-based impl., you need to talk over Serial / USB, and to make that non-blocking, the method needs to be async

om26er:
> def sign(self, data: bytes) -> bytes:

Ah, sorry about that, I had fixed that a few hours ago, forgot to push the commit. Done now.

Will take care of `format_challenge` now.

PR #1501, opened 2021-11-18
Make challenge signing reusable

om26er:
This is important for https://github.com/crossbario/crossbar/pull/1906 which I now have got working

### DIFF
--- a/autobahn/wamp/cryptosign.py
+++ b/autobahn/wamp/cryptosign.py
@@ -26,6 +26,7 @@
 
 import binascii
 import struct
+from typing import Callable
 
 import txaio
 
@@ -354,7 +355,7 @@ def _verify_signify_ed25519_signature(pubkey_file, signature_file, message):
 
 if HAS_CRYPTOSIGN:
 
-    def format_challenge(challenge, channel_id_raw, channel_id_type) -> bytes:
+    def format_challenge(challenge: Challenge, channel_id_raw: bytes, channel_id_type: str) -> bytes:
         if not isinstance(challenge, Challenge):
             raise Exception(
                 "challenge must be instance of autobahn.wamp.types.Challenge, not {}".format(type(challenge)))
@@ -390,7 +391,7 @@ if HAS_CRYPTOSIGN:
 
         return data
 
-    def sign_challenge(data, signer_func):
+    def sign_challenge(data: bytes, signer_func: Callable):
         # a raw byte string is signed, and the signature is also a raw byte string
         d1 = signer_func(data)
 
--- a/autobahn/wamp/cryptosign.py
+++ b/autobahn/wamp/cryptosign.py
@@ -354,7 +354,7 @@ def _verify_signify_ed25519_signature(pubkey_file, signature_file, message):
 
 if HAS_CRYPTOSIGN:
 
-    def format_challenge(session, challenge, channel_id_type) -> bytes:
+    def format_challenge(challenge, channel_id_raw, channel_id_type) -> bytes:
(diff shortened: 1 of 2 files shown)


autobahn/wamp/cryptosign.py
```diff
@@ -354,7 +354,7 @@ def _verify_signify_ed25519_signature(pubkey_file, signature_file, message):
 
 if HAS_CRYPTOSIGN:
 
-    def format_challenge(session, challenge, channel_id_type) -> bytes:
+    def format_challenge(challenge, channel_id_raw, channel_id_type) -> bytes:
```
om26er, line 357: @oberstet it might make sense to change this method to support `challenge` as a hex string as well.

One reason for that is that the worker would need to cbor serialize the challenge object manually and then the controller would need to deserialize that.

oberstet: my view: this is JS style;) we should add py3 type annotations (I did that for parts that I changed since this feature is available in functions that I touched .. partially at least), and also avoid `Optional`. in general, I would like to move the AB&CB codebases to type hinting inside. this has just so many advantages. now, of course, you could argue: why not `Optional` with str and bytes possible? well yes .. possible ... so I'd be fine with both these options. with type hinting

oberstet:
yes, cool! with added function docs and type hinting, I'd be completely happy=)

autobahn/wamp/cryptosign.py
om26er: Didn't need to change the above signature, as I went with a different solution in Crossbar. https://github.com/crossbario/crossbar/pull/1906/files#r752120365

om26er: Docstring and annotations are added now

om26er:
This PR also attempts to fix S3 deploy. Next need to fix docker container upload

oberstet:
Awesome! I am happy=)
Bildschirmfoto von 2021-11-18 21-57-50

autobahn/wamp/cryptosign.py
```diff
@@ -392,6 +399,36 @@ def format_challenge(session, challenge, channel_id_type) -> bytes:
 
         return data
 
+    def sign_challenge(data: bytes, signer_func: Callable):
```
oberstet, line 402: on a side note, I am wondering if there is an async callable type for the hinting .. as the callable must return a future (since the code doesn't use "maybe future" wrapper)

om26er: There is no neat solution there, especially since our functions return `Deferred` OR `Future` depending on the txaio backend.

The general suggestion on the internet seems to be `Callable[[bytes], Awaitable[bytes]]`, which means that the parameter takes `bytes` as a first argument and returns an "awaitable" that resolves to `bytes`

om26er:
Landing this now, will make further changes, if needed in a different PR.

autobahn/wamp/cryptosign.py
oberstet: ok, I see, yes. without a proper advanced type system, it's hard to express the choice of network framework determines exactly which of Future or Deferred must be returned. A run-time cross-cutting concern. Anyways, it's cool for me! type hint and docs should be clear enough.

oberstet: > Callable[[bytes], Awaitable[bytes]]

yeah, I can see how that makes sense. however, it points to an even bigger elephant: `Callable`? great, `lambda: None` is certainly good then, right? ;) IOW: `Callable` is an incomplete type.

PR #1512, opened 2022-01-05
Reset transport retry status when connection succeeds

alan-ls:
`max_retries` is intuitively the number of consecutive failures allowed before aborting.

Use case: The WAMP router should be allowed to restart or connections temporarily lost.


meejah (approved):
LGTM!

PR #1513, opened 2022-01-06
Allow optional keys in the transport endpoint configuration

alan-ls:
This came from an attempt to give more time to the router to start when both the component and the router services are started together.

I grepped for `endpoint.*get` and found 2 valid keys that cannot be set from the transport config:
* `timeout`
* `version`


oberstet:
yeah, that looks right! thanks=)

alan-ls:
Thanks for reviewing and merging so quickly!

PR #1544, opened 2022-04-10
Key modules

oberstet:


oberstet:
so this PR is supposed to add/expand interfaces in AB for i) consolidate key types and signing/verification, and b) allow both SW and HW based implementations ...

om26er (approved):
LGTM. And thanks, this lays good foundation for a future hardware based cryptosign implementation

oberstet:
yeah, it should be flexible enough to allow different integrations, incl. eg NXP SE050 talking the APDUs over I2C this chips implements in the Java IoT applet it runs on its JCOP on-chip Java.

it would also allow to integrate smart HW, like eg a signing HW with display allowing to present the signed data to the user (this is what EIP712 is for).

eg the HSM could talk WAMP-RawSocket-CBOR over Serial-USB and expose `IEthereumKey.sign_typed_data` to display and sign typed application data.

anyways, I'll add a few more doc comments and then merge, thanks for reviewing!

oberstet:
also, CB we should then obviously adjust CB to only use `IEd25519Key` and `IEthereumKey`. and AB cryptosign and xbr stuff.

once we have that, using a HSM should just be a plug away (well, we need the HSM impl as well;)

oberstet:
two Ifc questions we might still do:

* expose reference to containing ISecurityModule in ISigningKey
* move can_lock/lock/unlock from the module to the key level

I'm pretty sure the former is good, not sure about the latter .. what do you think?

PR #1573, opened 2022-05-20
Replace unmaintained pyqrcode dependency with qrcode

mgorny:
pyqrcode has not seem any activity since 2015.  Replace it with qrcode
library that is still maintained.  The results are roughly the same;
it seems that there is a tiny difference in SVG scale, and the terminal
version is half the size as qrcode is using more box drawing characters.


oberstet:
awesome, thanks a lot! looks very good, though sadly we lack test coverage - I will test it with crossbar (where this stuff is used), but would expect it to just work.

mgorny:
Thanks. I've tested them by calling the two functions directly, and comparing the results visually. Feel free to ping me if you run into any problems.

PR #1576, opened 2022-06-07
Flatbuffers IDL based payload validation

oberstet:


oberstet:
Bildschirmfoto von 2022-06-09 02-30-21

Issue #1592, opened 2022-08-18
asyncio rawsocket clients crash on WELCOME message


alan-ls:
Asyncio rawsocket clients crash when receiving a WELCOME message.
```
WARNING  root:rawsocket.py:87 Instantiate autobahn component
WARNING  root:rawsocket.py:92 Start component
WARNING  root:rawsocket.py:96 Wait for the compoenent to be connected
WARNING  autobahn.asyncio.component.Component:aio.py:162 trying transport 0 ("None") using connect delay 0
ERROR    autobahn.asyncio.wamp.Session:aio.py:162 Session.onUserError(): "While firing onWelcome"
Traceback (most recent call last):
  File "/.../txaio/aio.py", line 489, in done
    callback(res)
  File "/.../autobahn/wamp/protocol.py", line 605, in success
    transport=self._transport.transport_details,
AttributeError: 'WampRawSocketClientProtocol' object has no attribute 'transport_details'
```

Indeed compared to `autobahn.twisted.rawsocket.WampRawSocketProtocol`, the `transport_details` property is missing from `autobahn.asyncio.rawsocket.PrefixProtocol` and the framing type is not updated once the transport details are created.

oberstet:
fixed via https://github.com/crossbario/autobahn-python/pull/1593

PR #1593, opened 2022-08-18
Fix asyncio rawsocket protocol transport details (#1592)

alan-ls:


oberstet:
from a quick review, looks good to me! thanks for contributing

PR #1604, opened 2022-12-11
fix twisted 22.10.0 incompability

Mic92:


lopsided98 (commented):
I believe this causes a slight, and probably inconsequential, change in the error page content:

Previously, [`NoResource("No such child resource.")`](https://github.com/twisted/twisted/blob/8e7857d08355273179ca330c55f41713c65254c2/src/twisted/web/resource.py#L356) called [`_UnsafeErrorPage(brief="No Such Resource", detail="No such child resource.")`](https://github.com/twisted/twisted/blob/8e7857d08355273179ca330c55f41713c65254c2/src/twisted/web/resource.py#L295).

Now, [`notFound("No such child resource.")`](https://github.com/twisted/twisted/blob/f2f5e81c03f14e253e85fe457e646130780db40b/src/twisted/web/pages.py#L102) calls [`_ErrorPage(brief="No such child resource.", detail="Sorry. No luck finding that resource.")`](https://github.com/twisted/twisted/blob/f2f5e81c03f14e253e85fe457e646130780db40b/src/twisted/web/pages.py#L23).

autobahn/twisted/resource.py
```diff
@@ -100,7 +105,7 @@ def getChildWithDefault(self, name, request):
         """
         This resource cannot have children, hence this will always fail.
         """
-        return NoResource("No such child resource.")
+        return notFound(message="No such child resource.")
```
Mic92, line 108: Now both and old code should be better aligned...

oberstet:
thanks for contributing! fwiw, I _think_ the check CI step is failing due no a regression with CPy 311 - which the check step seems to use - and this is unrelated to the PR here (which looks good btw)

seems to be a thing a couple of project run into, google for "cannot import name 'getargspec' from 'inspect'" ..

I haven't dug further, but likely some import needs a fix, or the check CI step should use a fixed CPy, not CPy 3.x ...

oberstet:
the failing 2 steps (check and docs) in

https://github.com/crossbario/autobahn-python/blob/master/.github/workflows/main.yml

are failing since the steps use "open ended" python version

```
python-version: '3.x'
```

which runs into issues unrelated to the PR.

we might change those 2 location from "3.x" to "3.10".

and we should also expand

https://github.com/crossbario/autobahn-python/blob/a99ff5dd4e54595bd13ffdcf92d72e0737246e38/.github/workflows/main.yml#L75

to include 3.11 ...

---

with those changes to main.yaml, this PR should run fully green and could be merged

> note: I just merged another PR already a couple of weeks old - which now runs red on check/docs after merging due to the same unrelated issue. so after above, everything should be green again;)

Mic92:
done.

oberstet:
alright, thanks, that "fixed" (circumvents) the problem on check/docs step. unfortunately, the enabling of 3.11 for the main tests uncovers another issue:

```
/internet/test/fake_CAs
        running build_ext
        building 'twisted.test.raiser' extension
        creating build/temp.linux-x86_64-cpython-311
        creating build/temp.linux-x86_64-cpython-311/src
        creating build/temp.linux-x86_64-cpython-311/src/twisted
        creating build/temp.linux-x86_64-cpython-311/src/twisted/test
        gcc -pthread -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/home/runner/work/autobahn-python/autobahn-python/.tox/tw1910/include -I/opt/hostedtoolcache/Python/3.11.0/x64/include/python3.11 -c src/twisted/test/raiser.c -o build/temp.linux-x86_64-cpython-311/src/twisted/test/raiser.o
        src/twisted/test/raiser.c:198:12: fatal error: longintrepr.h: No such file or directory
          198 |   #include "longintrepr.h"
              |            ^~~~~~~~~~~~~~~
        compilation terminated.
        error: command '/usr/bin/gcc' failed with exit code 1
        [end of output]
    
    note: This error originates from a subprocess, and is likely not a problem with pip.
  error: legacy-install-failure
  
  × Encountered error while trying to install package.
  ╰─> twisted
```

other project are running into this as well https://github.com/cython/cython/pull/4428

googling for "twisted python 3.11 longintrepr.h" doesn't seem to return anything specific twisted related, so this might be a new twisted issue (on 3.11)

----

obviously, this is unrelated to this PR, and hence, you could change this PR again back to `['3.7', '3.10', 'pypy-3.8']` - well, unless you have a fix for twisted that is;)

as a consequence: we can't bump autobahn to fully support py 3.11, at least because of the remaining twisted issue .. sorry for the hassles! OSS maintenance real-world;)

PR #1634, opened 2024-04-12
unit test fixes

dimbleby:


oberstet:
oh cool, thanks for your work and for contributing!

it seems like this is really an improvement, and the only test failure remaining is only of pypy on an _older_ twisted (22.10), and the failure is _not_ present_ anymore on that same pypy (3.9) with twisted trunk! interesting.

*I'd be fine with bumping the minimum required twisted for the whole of autobahn - moving radically, that is moving to minimum twisted v24.3.0 !*

but we can do that in another PR, I am going to merge this one now, it is really an improvement, thanks again!

----

on a totally unrelated topic, I noticed (as I am usually having a quick look at what new contributors actually do / have on their plate), I noticed your work on

https://github.com/blue-llama/e-treasure-hunt

what is this thing? I am curious;) is it related to http://e-treasurehunt.net/ ? and what is that? the tag line crosses 2 areas (outdoor/adventure and web/online-gaming?) which I find interesting, and actually I always wondered about the cross-over section between both ... well .. if that is what this thing is about anyways;)

oberstet:
> moving to minimum twisted v24.3.0

let's see: https://github.com/crossbario/autobahn-python/pull/1635

dimbleby:
The treasure hunt thing is not the one you found, it's a puzzle/game sort of thing in which images are gradually revealed and solvers are supposed to find a location.

oberstet:
> it's a puzzle/game sort of thing in which images are gradually revealed and solvers are supposed to find a location.

so "it" is https://github.com/blue-llama/e-treasure-hunt here?

and "location" (in that puzzle/game you had been contributing to) is a real / physical world location, and so "images" are photos of the real / physical world?

sorry for stupid questions;) but now that I started, I actually want to understand .. "finding real world" locations from / via an electronic puzzle/game .. sounds interesting.

dimbleby:
it is a game I play with colleagues.  Each level consists of 5 images, revealed one at a time as needed (players can choose to ask for hints, but not before time has passed - they are made to suffer first).  Each solution is a location.

The only constraints on the levels are: five images, solution is a precise location.  The expectation is that the puzzle should be difficult with only one image available, obvious with five images available - and graded appropriately in between.

We crowd-source the levels, they vary enormously.  Some are straight-up pictures of places.  Some hint at people, with the solution being a place associated with the person.  Some are entirely cryptic puzzles.  Some get meta, linking together previous levels.  And so on.

If you are actually interested in playing with this thing you will want to use my fork, that is the one that we actually run the game from these days.

oberstet:
> The only constraints on the levels are: five images, solution is a precise location.

Sounds interesting! This connecting physical world images and make that into an online game searching for the right physical location (the photos had been taken from).

How do players reveal or propose precise locations?

GPS coordinates?

It sounds like to win, I only need to enter (online in the game) the right coordinates?

So to "reveal" (or propose a solution), I do _not_ have to be physically present (my body) _at that location_ in the real world (entering is enough), right?

If so, have you thought about that latter thing?

Because to _take_ the real world photos, I need to be physically present?

Also: can I "cheat" and just download some pics made by others of some physical location and make up a new game?

What I am trying to get at: how does the "connect" between virtual world (the online portions) and the physical world look/work exactly? Curious;) Sorry for the many questions.

dimbleby:
answers are submitted either by clicking on a map, or as coordinates.  (Quite often the cryptic clues do not refer to a place as such, but give a numerical solution - then coordinates is the way to give the answer),

We have had one or two levels that interacted with the real world but the players are typically distributed across different geographies so our version is almost entirely an online game.

oberstet:
thanks a lot! I get it.

> We have had one or two levels that interacted with the real world but the players are typically distributed across different geographies so our version is almost entirely an online game.

ok, I see.

yeah, "player are distributed across different geographies" - this "limits" the ways one could create the physical-virtual world links.

In fact, I was originally more thinking like:

* player A hides a treasure in the physical world, takes a series of photos giving clues, and then submit the photos
* players B<n> then - given the hinting photos (sequentially revealed in gaming rounds) - try to figure out the physical location BUT THEN: have to physically go there .. to that supposedly correct real world location ..
* and IF they are at the hidden treasure, somehow proof the fact that they really have found that treasure and are physically present there

That could be done using NFC crypto tags with 30 cent price which can be securely scanned and submitted (the presence at that tag) using any smartphone without an app

But yes, this presumes physical presence, by both the original player hiding the treasure and players which want to find / reveal / gain that treasure!

It could also be be made to allow multiple players to find that one treasure (when the tag is not physically removed once found) .. securely

Anyways, interesting topic!!

dimbleby:
We have had the occasional metadata accident, but setters are encouraged to be careful.  Some have even used the metadata as part of the puzzle. 

Required precision is chosen by the setter, experience says that requiring a click to be more accurate than 50m or so is pretty annoying.

oberstet:
> Some have even used the metadata as part of the puzzle.

ah, right! I forgot: "revealing" unintentional correct metadata also always means ability to intentionally "reveal" incorrect metadata;) nice! so a "metadata analyzer" (for players) would need to check both for presence of metadata at all, PLUS estimate a likelihood of "metadata is fake" ... e.g. "forward image search": given a presumed location (eg from metadata), how likely is it that the revealed picture had actually been taken at that location?

> Required precision

I see, makes sense. 50m precision: will likely make my cheat c) non-working.

PR #1636, opened 2024-04-13
Check first and then do increment

SunFulong:
Wrapping around at the boundary should do check first.

### DIFF
--- a/autobahn/util.py
+++ b/autobahn/util.py
@@ -235,8 +235,9 @@ class IdGenerator(object):
         :returns: The next ID.
         :rtype: int
         """
-        self._next += 1
-        if self._next > 9007199254740992:
+        if self._next < 9007199254740992:
+            self._next += 1
+        else:
             self._next = 1
         return self._next
 


oberstet:
This is wrong. It won't create an ID 9007199254740992. It is also unnecessary, the code as it stands is correct.

SunFulong:
> It won't create an ID 9007199254740992.

Although it won't created, but it can be reached at previous round.

oberstet:
> Although it won't created, but it can be reached at previous round.

yes, you are right actually, it would spit out 9007199254740992. however, I stand by my other comment: it is unnecessary (IOW: the PR code does not change the functional behavior in any way. and non-functional behavior .. performance .. I am not convinced it would be an improvement ... or even necessary anyways.

if you file a PR that changes the `0` to `1` to fix the bug I mentioned above, that would be highly welcome! I'd merged it right away ...

SunFulong:
> if you file a PR that changes the 0 to 1 to fix the bug

See #1637

PR #1645, opened 2024-07-29
This branch pins setuptools to 70.0.0

manand881:
which is a wellknown stable version of setuptools
and has been done to avoid issues such as the one
in the link given below

https://github.com/pypa/setuptools/issues/4519


dimbleby:
setuptools wants you not to do this https://github.com/pypa/setuptools/pull/4524

dimbleby:
Also you've put the bound in the wrong place to do anything about build issues anyway

oberstet:
sorry, this is just some hack in my eyes .. whatever the actual issue is .. if there is one

PR #1648, opened 2024-08-22
Switch to log.warning instead of log.warn

### LINKED ISSUE #1098: txaio/aio.py:202: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
I know the warning comes from txaio, but, in fact, this is autobahn-python that uses the old deprecated log calls.

~~(Will try to do some PR later)~~

edit: changing warn by warning breaks all twisted tests, Logger has no attribute warning ...

Jenselme:
log.warn is removed in Python 3.13

Close #1098

### DIFF
--- a/autobahn/asyncio/component.py
+++ b/autobahn/asyncio/component.py
@@ -304,7 +304,7 @@ class Component(component.Component):
         """
 
         if loop is None:
-            self.log.warn("Using default loop")
+            self.log.warning("Using default loop")
             loop = asyncio.get_event_loop()
 
         return self._start(loop=loop)
--- a/autobahn/asyncio/rawsocket.py
+++ b/autobahn/asyncio/rawsocket.py
@@ -303,7 +303,7 @@ class WampRawSocketMixinGeneral(object):
             self._session.onOpen(self)
         except Exception as e:
             # Exceptions raised in onOpen are fatal ..
-            self.log.warn("WampRawSocketProtocol: ApplicationSession constructor / onOpen raised ({err})", err=e)
+            self.log.warning("WampRawSocketProtocol: ApplicationSession constructor / onOpen raised ({err})", err=e)
             self.abort()
         else:
             self.log.info("ApplicationSession started.")
@@ -316,11 +316,11 @@ class WampRawSocketMixinGeneral(object):
                 self._session.onMessage(msg)
 
         except ProtocolError as e:
-            self.log.warn("WampRawSocketProtocol: WAMP Protocol Error ({err}) - aborting connection", err=e)
+            self.log.warning("WampRawSocketProtocol: WAMP Protocol Error ({err}) - aborting connection", err=e)
             self.abort()
 
         except Exception as e:
-            self.log.warn("WampRawSocketProtocol: WAMP Internal Error ({err}) - aborting connection", err=e)
+            self.log.warning("WampRawSocketProtocol: WAMP Internal Error ({err}) - aborting connection", err=e)
             self.abort()
 
     def send(self, msg):
@@ -360,7 +360,7 @@ class WampRawSocketMixinAsyncio(object):
             self._session.onClose(wasClean)
         except Exception as e:
             # silently ignore exceptions raised here ..
-            self.log.warn("WampRawSocketProtocol: ApplicationSession.onClose raised ({err})", err=e)
+            self.log.warning("WampRawSocketProtocol: ApplicationSession.onClose raised ({err})", err=e)
         self._session = None
 
     def close(self):
--- a/autobahn/twisted/choosereactor.py
+++ b/autobahn/twisted/choosereactor.py
@@ -94,11 +94,11 @@ def install_optimal_reactor(require_optimal_reactor=True):
                     from twisted.internet import kqreactor
                     kqreactor.install()
                 except:
-                    log.warn('Running on *BSD or MacOSX, but cannot install kqueue Twisted reactor: {tb}', tb=traceback.format_exc())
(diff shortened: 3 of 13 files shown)


Jenselme:
From what I understand of the error, I may be completely wrong on how to solve this: I seem to have modified calls to twited `Logger` class which does use `warn`. So the issue about `warn` being deprecated may come from a dependency and not this packages directly (nor txaio).

Testing this properly is made harder by the fact that [pyunormalize cannot be installed on Python 3.13 yet](https://github.com/mlodewijck/pyunormalize/issues/4)

Any insights no this?

oberstet:
thanks for taking care of such things .. pls lemme comment ..

> log.warn is removed in Python 3.13

if so, then

* the CI in Autobahn should _first_ be upgraded to 3.13 in an _independent_ PR
* this should then run red (fail), because the function is removed

once that is done, _this_ PR can continue ..

-----

right now, 3.13 is not yet released

grafik

----

currently, the CI uses CPython 3.11 and PyPy 3.9

https://github.com/crossbario/autobahn-python/blob/7bc85b34e200640ab98a41cfddb38267f39bc92e/.github/workflows/main.yml#L60

and PyPy is now at 3.10

https://pypy.org/download.html

for the CI upgrade PR, we should first define the exact version goals for CPython and for PyPy ..

Jenselme:
Sounds good to me. I started looking into it because fedora is starting to switch to 3.13 in its dev version and it seemed easy to do (sadly I was proven wrong).

But, yes, we can wait for Python 3.13 to be officially release and migrate the CI first.

oberstet:
> But, yes, we can wait for Python 3.13 to be officially release and migrate the CI first.

I'm confident this ordering will support Autobahn's goal of improving robustness & production stability.

if there is a regression in Autobahn triggered by a dependency upgrade, the regression should first be _proven_ to actually exist and be valid

this requires the CI to first not touch any code, but only upgrade the dependency

once the regression is confirmed, the issue can be fixed in another PR

this allows to exactly follow the sequence of events with proof at each step, and the option for relatively easy rollback should something goes wrong. as there will more often than not in reality;) like ripples in a subsequent, dependent project like Crossbar.io (which obviously uses Autobahn a lot)

yan12125:
Python 3.13 is out. Time to move on?

oberstet:
@yan12125 wau. you are quick;) I also saw that flying by. so yes: +1!

maybe worth: I think it would be good to first (and only that) switch from 3.11 to 3.13 everywhere in CI etc, e.g.

https://github.com/crossbario/autobahn-python/blob/7bc85b34e200640ab98a41cfddb38267f39bc92e/.github/workflows/main.yml#L28

and once that runs, only then move on to the log issue (the original itch of this issue), and this PR

in other words, what I am saying is: it might be worth doing the 3.11 to 3.13 switch (and only that) in a (new) separate PR

this latter (new) PR _should_ be all green immediately. in theory. I don't believe it. sadly.

then _this_ PR can simply catch up with / merge main to have 3.13

it will then run as is right away, and we'll see what is yet missing then ...

sounds like a good / workable plan?

yan12125:
Sounds good. I can work on 3.11 -> 3.13 switch if you wish.

yan12125:
The linked issue https://github.com/crossbario/autobahn-python/issues/1098 mentions:

> FYI, `logger.warn` is removed in Python 3.13: https://docs.python.org/3.13/whatsnew/3.13.html#logging

That page on Python official website actually places `logging.warn` in the section "Pending Removal in Future Versions", so that function is still in Python 3.13. Also, Python 3.13 unit tests passed without changes in this pull request: https://github.com/yan12125/autobahn-python/actions/runs/11666669688.

@Jenselme could you check how to make deprecation warnings visible in unit tests?

Jenselme:
@yan12125 I’m already seeing the warnings:

```
.tox/py3.11-asyncio/lib64/python3.11/site-packages/websockets/legacy/__init__.py:6
  /home/jenselme/Projects/_Clones/autobahn-python/.tox/py3.11-asyncio/lib64/python3.11/site-packages/websockets/legacy/__init__.py:6: DeprecationWarning: websockets.legacy is deprecated; see https://websockets.readthedocs.io/en/stable/howto/upgrade.html for upgrade instructions
    warnings.warn(  # deprecated in 14.0
```

It’s still not easy to spot where the issue comes from though.

yan12125:
> @yan12125 I’m already seeing the warnings:
> 
> ```
> .tox/py3.11-asyncio/lib64/python3.11/site-packages/websockets/legacy/__init__.py:6
>   /home/jenselme/Projects/_Clones/autobahn-python/.tox/py3.11-asyncio/lib64/python3.11/site-packages/websockets/legacy/__init__.py:6: DeprecationWarning: websockets.legacy is deprecated; see https://websockets.readthedocs.io/en/stable/howto/upgrade.html for upgrade instructions
>     warnings.warn(  # deprecated in 14.0
> ```
> 

Thanks, but the warning is about `warnings.warn`, not `log.warning` .

> It’s still not easy to spot where the issue comes from though.

You may check it by running `export PYTHONWARNINGS=error` first.

Jenselme:
Currently, the test suite fails to run with Python 3.13 because of https://github.com/jnwatson/py-lmdb/issues/362 Will try to monitor this issue and come back to this once it’s solved.

oberstet:
ah, right. thanks for digging! the PR https://github.com/jnwatson/py-lmdb/pull/368 claims to solve that, so let's wait and see, yeah.

Jenselme:
https://github.com/jnwatson/py-lmdb/pull/368 was merged and I know of no issue with log.warn. Closing. Sorry for the noise.

PR #1651, opened 2024-12-10
Stop using twisted.internet.defer.returnValue

cjwatson:
`defer.returnValue` was only needed in Python 2; in Python 3, a simple `return` is fine.

`twisted.internet.defer.returnValue` is deprecated as of Twisted 24.7.0.

### DIFF
--- a/autobahn/twisted/cryptosign.py
+++ b/autobahn/twisted/cryptosign.py
@@ -27,7 +27,7 @@
 
 from autobahn.wamp.cryptosign import HAS_CRYPTOSIGN, CryptosignKey
 
-from twisted.internet.defer import inlineCallbacks, returnValue
+from twisted.internet.defer import inlineCallbacks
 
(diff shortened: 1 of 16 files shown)


oberstet:
hi there, thanks for contributing!

unfortunately, another case of unrelated issues .. sorry, it is kinda never ending story - due to the complexity of this project, and even though we've tried hard over and over to pin dependencies and what not, well, still not

the docs CI failure .. seems like sphinx has a new / more strict check - unrelated to this PR

https://github.com/crossbario/autobahn-python/actions/runs/12254786525/job/34188986727?pr=1651#step:6:452

the pypy CI failure -- also unrelated .. nasm too old, wtf, haven't analyzed what change upstream is the origin

https://github.com/crossbario/autobahn-python/actions/runs/12254786525/job/34188989569?pr=1651#step:6:512

PR #1658, opened 2025-05-13
attempt to fix CI

deronnax:
CI on main is [currently failing](https://github.com/crossbario/autobahn-python/actions/runs/14976696211/job/42070723486) with this error:

```
    ../meson.build:31:9: ERROR: Dependency 'girepository-2.0' is required but not found.
```
Probably just a missing dependancies that was previously present by default on ubuntu images. Let's try to fix it in this PR. I think we are experiencing the same problem than https://github.com/beeware/toga/issues/3143 (TLDR: pygobject packaging changed)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,7 +18,7 @@ jobs:
 
   build_wheels:
     name: Build & Upload
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v3
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
       # WEB3_INFURA_PROJECT_ID: ${{ secrets.WEB3_INFURA_PROJECT_ID }}
       # WEB3_INFURA_API_SECRET: ${{ secrets.WEB3_INFURA_API_SECRET }}
 
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v3
 
@@ -52,8 +52,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-22.04]
-        # os: [ubuntu-22.04, macos-latest, windows-latest]
+        os: [ubuntu-24.04]
+        # os: [ubuntu-24.04, macos-latest, windows-latest]
 
         # python 3.11 fails with "src/twisted/test/raiser.c:198:12: fatal error: longintrepr.h: No such file or directory"
         # twisted doesn't yet support 3.11 formally: https://github.com/twisted/twisted/blob/trunk/pyproject.toml#L24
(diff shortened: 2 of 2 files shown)


deronnax:
@oberstet can I get a workflow run please ?

oberstet:
thanks for working on this! let's see what happens. oh, did you retract the PR again? anyways, I'll start again should you submit a PR. rgd

>missing dependancies that was previously present by default on ubuntu images

we are using virtualenv's, and not distro Python. everywhere. as it should be. do not use distro Python, the repackaging of Python packages into distro packages is unneeded/unhelpful/not-recommended. well, in general, and yes, I have strong opinions;)

oberstet:
`girepository` isn't on PyPI - as a separate package. what Python package does it come with?

maybe just gobject? as in: https://github.com/beeware/toga/issues/3143

deronnax:
I will explain you a bit later. I just pushed a commit but I need an approval for the CI run, again. Can you do make it so that I don't need an approval on this PR for every commit?

oberstet:
thanks for trying! makes sense, but ... 

```
oberstet@intel-nuci7:~$ uname -a
Linux intel-nuci7 5.15.0-139-generic #149-Ubuntu SMP Fri Apr 11 22:06:13 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
oberstet@intel-nuci7:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.5 LTS
Release:	22.04
Codename:	jammy
oberstet@intel-nuci7:~$ apt show libgirepository-2.0-dev
N: Paket libgirepository-2.0-dev kann nicht gefunden werden.
N: Mittels des Musters »libgirepository-2.0-dev« konnte kein Paket gefunden werden.
N: Paket libgirepository-2.0-dev kann nicht gefunden werden.
N: Mittels des Musters »libgirepository-2.0-dev« konnte kein Paket gefunden werden.
E: Keine Pakete gefunden
oberstet@intel-nuci7:~$ 
```

which is the same on the GH build env

https://github.com/crossbario/autobahn-python/blob/edc3c94c3cff717c9128217115dc6b980e2784ea/.github/workflows/main.yml#L19

> Can you do make it so that I don't need an approval on this PR for every commit?

I _think_ this will happen automatically once a PR by you is merged to this repo ..

oberstet:
what you suggest (the 2 bullet points) makes sense IMO, I agree!

so bumping `runs-on: ubuntu-22.04` to 24.04 LTS - in all GH workflow files (not only main) in this repo

I would leave the CPython which we currently use at 3.11 as it is quite current (the latest upstream release is 3.13.3)

but upgrading the PyPy also to 3.11 would make sense IMO - that version is now available upstream, so we could bump `pypy-3.10` to `pypy-3.11` - and then have both CPython and PyPy in sync

if you have time, I'd be +1 and merge, in a *separate PR*, and once that is merged, rerun this PR - which should then be green - and merge as well ...

oberstet:
>  sudo: command not found

not sure why ... _maybe_ because you changed multi-line `run: |` to single-line `run ... && ...` .. maybe GH does some magic with `|` .. why did you change?

deronnax:
Yep. I am trying a few things on a PR of my own, don't mind me. I close this PR, I will reopen it when I have the final version.

PR #1661, opened 2025-06-08
Plain twisted utilities are sufficient

meejah:
Twisted 25.5.0 causes Autobahn's test reactor to fail -- but it turns out it's not really required and a fix for Twisted is to just remove this.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
       # WEB3_INFURA_PROJECT_ID: ${{ secrets.WEB3_INFURA_PROJECT_ID }}
       # WEB3_INFURA_API_SECRET: ${{ secrets.WEB3_INFURA_API_SECRET }}
 
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
 
@@ -52,12 +52,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-22.04]
-        # os: [ubuntu-22.04, macos-latest, windows-latest]
+        os: [ubuntu-latest]
+        # os: [ubuntu-latest, macos-latest, windows-latest]
 
         # python 3.11 fails with "src/twisted/test/raiser.c:198:12: fatal error: longintrepr.h: No such file or directory"
         # twisted doesn't yet support 3.11 formally: https://github.com/twisted/twisted/blob/trunk/pyproject.toml#L24
-        python-version: ['3.9', '3.11', 'pypy-3.10']
+        python-version: ['3.9', '3.11', 'pypy-3.9']
         framework: ['asyncio', 'tw2403', 'twtrunk']
 
     # https://github.blog/changelog/2020-04-15-github-actions-new-workflow-features/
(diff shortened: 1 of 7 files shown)


oberstet:
I _think_ this `Run-time dependency girepository-2.0 found: NO (tried pkgconfig and cmake)`

https://github.com/crossbario/autobahn-python/actions/runs/15514087624/job/43678656324?pr=1661#step:6:113

and similar is due to

https://github.com/crossbario/autobahn-python/blob/edc3c94c3cff717c9128217115dc6b980e2784ea/.github/workflows/main.yml#L19

which has been bumped on Crossbar.io to

https://github.com/crossbario/crossbar/blob/d76f83c4eb5bb6f584b485c5160cf634fc44f23f/.github/workflows/main.yml#L13

sadly, I can't find the issue on Crossbar.io where this was discussed - that has more details ..

oberstet:
@meejah OT, but a little bit of fun;) want a "Verified Tick" on LinkedIn? As in:

grafik

Here is the badge generator:

https://gist.github.com/oberstet/893eae7a6bc6c152aafca71d13b8b146

AI generated in 5min:

grafik

meejah:
hey ... I haven't looked at it very closely. I like the general goal ("more distributed") but also know that's hard -- and Signal works well (as in: my non-tech friends have it, and use it). And Signal has seen a lot of security attention (and, nearly as important these days, actual subpoenas to respond to giving good information on what exactly _is_ stored).

meejah:
Still not sure what's up with PyPy vs. LMDB -- works fine locally. Only tried on Debian, but ...

oberstet:
> Still not sure what's up with PyPy vs. LMDB -- works fine locally. Only tried on Debian, but ...

well:( yes. kinda 2 techs (PyPy and LMDB) that _are_ awesome, and are underrated or better: would _deserve_ lots more. support. money. all. seems like not happening on this planet. fact.

meejah:
A big reason Signal used phone numbers was so they *didnt* learn the social graph. They've since launched usernames etc (PIR-based lookup I think?) so they probably *could* ditch them now. You no longer have to use phone numbers w Signal contacts -- and there's a PIN to prevent SS7 attacks.

I assume they keep them now for spam prevention / familiarity?

meejah:
p.s. I've had no problem making Signal accounts w landline or temp numbers from e.g. tossabledigits dot com or similar

Skully17:
I've noticed that the CI runs green if pypy is changed back to 3.9.
I don't suppose it's great to go backwards but I think it's better to have a passing CI and have a separate PR for upgrading pypy.

meejah:
(p.s. rebased all the fixup commits into just "fix CI")

oberstet:
> like "PyPy version running Python version 3.X equivalent"

fwiw, my understanding is that there is a PyPy version V1 that is for the PyPy software release *itself*, and then there is a version of CPython V2 which V1 is supposed to correspond to and replicate (behave like it)

I think - and still do so - that AB and CB workflows aim to _name_ the workflow steps for _both_ CPython _and_ PyPy using one versioning scheme, namely the CPython / V2 one

that idea is: if there is `'3.9'` and `pypy-3.9`, then the former is for CPython at **V2** == v3.9, and the latter is PyPy at V1 = ? and **V2** = v3.9

looking up the PyPy version V1 _by_ V2 would be incredibly useful to have documented in PyPy upstream or even programmatically available for all released PyPy versions

further, a statement like "PyPy version vX.Y" is slightly ambiguous, because it is unclear whether vX.Y refers to the V1 or V2 versioning scheme

personally, I would say "PyPy version vX.Y - unqualified" should mean V1, thus the version of PyPy itself, thus v7.?.? 

one could argue - overly cleverly IMP - that if the version starts with **7**, then it MUST be V1. which is true, from looking the numbers 

https://downloads.python.org/pypy/

image

Issue #1834, opened 2026-01-01
[BUG] Incorrect machine is detected while cross compiling


jameshilliard:
Host compiler args are used instead of target args for target build when cross compiling:
```python
>>> python-autobahn 25.12.2 Building
[...]

PR #1835, opened 2026-01-01
Use sysconfig.get_platform for machine detection

jameshilliard:
This should function correctly when cross compiling unlike platform.machine.

## Description

Uses sysconfig to detect target architecture correctly.

---

## Related Issue(s)

Closes or relates to #1834

---

## Checklist

      updated the changelog
      in this PR

### DIFF
--- a/.audit/jameshilliard_sysconfig-machine.md
+++ b/.audit/jameshilliard_sysconfig-machine.md
@@ -1,0 +1,10 @@
+## AI Assistance Disclosure
+
+- [x] I did **not** use any AI-assistance tools to help create this pull request.
+- [ ] I **did** use AI-assistance tools to *help* create this pull request.
+- [x] I have read, understood and followed the projects' [AI Policy](https://github.com/crossbario/autobahn-python/blob/main/AI_POLICY.md) when creating code, documentation etc. for this pull request.
+
+Submitted by: @jameshilliard
+Date: 2026-01-01
+Related issue(s): #1834
+Branch: sysconfig-machine
--- a/src/autobahn/nvx/_compile_args.py
+++ b/src/autobahn/nvx/_compile_args.py
@@ -124,7 +124,7 @@ See Also
 
 import os
 import sys
-import platform
+import sysconfig
 
 
 def is_building_wheel():
@@ -194,7 +194,7 @@ def get_compile_args():
         return ["/O2", "/W3"]
 
     # GCC/Clang on POSIX (Linux, macOS, *BSD)
-    machine = platform.machine().lower()
+    machine = sysconfig.get_platform().lower().split("-")[-1]
 
     # Base flags for all POSIX platforms
     base_args = [
@@ -245,7 +245,7 @@ def _get_safe_march_flag(machine):
     Parameters
     ----------
     machine : str
-        Machine architecture from platform.machine().lower()
+        Machine architecture from sysconfig.get_platform().lower().split("-")[-1]
 
     Returns
     -------


oberstet:
# Review of PR #1835

Thanks for the contribution! Your change to use `sysconfig.get_platform()` is on the right track, but the fix is incomplete.

## The Root Cause

The problem has two parts:

1. **Architecture detection** uses `platform.machine()` which returns the HOST architecture (x86_64), not the TARGET (aarch64)
2. **`-march=native`** is used for "local source builds" but this is invalid for cross-compilation

Your PR fixes part 1, but not part 2.

## Why the Current PR is Incomplete

After your change, the code path is:

```python
machine = sysconfig.get_platform().lower().split("-")[-1]  # Now correctly: "aarch64"
# ...
else:
    # Building from source locally: use -march=native for maximum performance
    return base_args + ["-march=native"]  # ❌ Still fails!
```

In buildroot's cross-compilation:
- No CI/wheel environment variables are set (`CI`, `CIBUILDWHEEL`, etc.)
- `is_building_wheel()` returns `False`
- Code falls through to `else` branch → uses `-march=native`
- **The aarch64 cross-compiler still gets `-march=native` and fails**

The error happens because `-march=native` tells GCC to query the HOST CPU's capabilities, but the aarch64 cross-compiler doesn't understand x86_64 architecture features.

## Complete Fix Needed

The fix needs to also **detect cross-compilation** and avoid `-march=native`:

```python
import platform
import sysconfig

def is_cross_compiling():
    """Detect if we're cross-compiling (host != target)."""
    host_machine = platform.machine().lower()
    target_platform = sysconfig.get_platform().lower()
    target_machine = target_platform.split("-")[-1]

    # Normalize architecture names
    host_normalized = _normalize_arch(host_machine)
    target_normalized = _normalize_arch(target_machine)

    return host_normalized != target_normalized

def _normalize_arch(arch):
    """Normalize architecture names for comparison."""
    if arch in ("x86_64", "amd64", "x64"):
        return "x86_64"
    elif arch in ("aarch64", "arm64"):
        return "aarch64"
    return arch
```

Then in `get_compile_args()`, add a check before the `else` branch:

```python
elif is_cross_compiling():
    # Cross-compiling: -march=native is invalid, use safe baseline for target
    arch_flag = _get_safe_march_flag(machine)
    if arch_flag:
        return base_args + [arch_flag]
    else:
        return base_args

else:
    # Native build: -march=native is safe
    return base_args + ["-march=native"]
```

## CI Failure Note

The type check (`ty`) failure you're seeing is **not related to your PR**. It's a pre-existing typo in the justfile:

```
--ignore non-subscriptable \   # WRONG - should be "not-subscriptable"
```

Would be cool if you could add that as well, because then the PR should run fully green!

---

Would you like to update your PR with the complete cross-compilation detection, or would you prefer I handle this in a separate PR?

jameshilliard:
There's a rather subtle bug in this code that I've seen in cargo as well, you can't actually reliably detect cross compilation with this approach at all. It fails to cover scenarios like cross compiling on a x86_64 host for a x86_64 target(say with a different userspace and/or cpu variant) which can result in a number of problems: Best option IMO is to avoid `-march=native` and purely use sysconfig for platform detection.
```python
def is_cross_compiling():
    """Detect if we're cross-compiling (host != target)."""
    host_machine = platform.machine().lower()
    target_platform = sysconfig.get_platform().lower()
    target_machine = target_platform.split("-")[-1]

    # Normalize architecture names
    host_normalized = _normalize_arch(host_machine)
    target_normalized = _normalize_arch(target_machine)

    return host_normalized != target_normalized
```


> The error happens because `-march=native` tells GCC to query the HOST CPU's capabilities, but the aarch64 cross-compiler doesn't understand x86_64 architecture features.

Yeah, I was testing with AUTOBAHN_ARCH_TARGET=safe in the env to avoid `-march=native` from being set, I figured that was sort of the expected way of handling non-native builds.

jameshilliard:
Technically buildroot can pass explicit [optimization flags](https://github.com/buildroot/buildroot/blob/2025.11/arch/Config.in.x86#L2-L21) kinda like [this](https://github.com/buildroot/buildroot/blob/2025.11/package/python-argon2-cffi-bindings/python-argon2-cffi-bindings.mk#L19-L23) as well via the env to a python package, this should be fairly reliable but adds a little bit of complexity on the packaging side of things.

oberstet:
so in summary:

## Complete Fix Needed

The fix should **change the default** from `-march=native` to safe baseline:

```python
else:
    # Building from source: use safe baseline by default
    # This ensures cross-compilation works (buildroot, etc.)
    # Users who want -march=native can set AUTOBAHN_ARCH_TARGET=native
    arch_flag = _get_safe_march_flag(machine)
    if arch_flag:
        return base_args + [arch_flag]
    else:
        return base_args
```

This way:
- **Cross-compilation works** out of the box (buildroot, Yocto, etc.)
- **Source distro users** (Gentoo, Arch) can opt-in with `AUTOBAHN_ARCH_TARGET=native`
- **Our published wheels** continue using safe baseline (existing behavior via `is_building_wheel()`)

## Summary of Required Changes

1. ✅ Use `sysconfig.get_platform()` for architecture detection (your current PR)
2. ❌ Change default from `-march=native` to safe baseline (missing)

## CI Failure Note

The type check (`ty`) failure you're seeing is **not related to your PR**. It's a pre-existing typo in the justfile:

```
--ignore non-subscriptable \   # WRONG - should be "not-subscriptable"
```

jameshilliard:
> * **Source distro users** (Gentoo, Arch) can opt-in with `AUTOBAHN_ARCH_TARGET=native`

Do you prefer opt-in like this for `-march=native` or opt-out using `AUTOBAHN_ARCH_TARGET=safe` for cross compilation?

jameshilliard:
Anyways I think the default is probably a separate issue than the machine detection issue here to some degree so maybe leave that for a follow-up PR?

oberstet:
> Do you prefer opt-in like this for -march=native or opt-out using AUTOBAHN_ARCH_TARGET=safe for cross compilation?

good question;) @mgorny : any opinions? I'd guess Gentoo would want "native"  - even though practical difference vs safe options is likely small (if measurable at all). but Gentoo isn't doing cross compile. Buildroot and Yocto do. personally, I mainly care about our own published native binary wheels for Debian ARM64, and for PyPy. and we are using Qemu with PyPA Docker images, not cross-compile, to build those wheels.

----

**to move forward, I prefer to "default on safe", and "opt-in for native"**

oberstet:
Superseeded by https://github.com/crossbario/autobahn-python/pull/1858

Thanks again @jameshilliard for reporting #1834 and for this PR — your
`sysconfig.get_platform()` approach for detecting the *target* architecture
under cross-compilation is exactly right, and it's carried over into the fix.

We've completed the change along the lines we discussed in this thread — making
the **portable baseline the default for every build context** (wheels, local
source installs, and cross-compilation), with **`-march=native` now opt-in** via
`AUTOBAHN_ARCH_TARGET=native`. That's what actually unblocks Buildroot/Yocto:
the cross toolchain is never handed the host-only `-march=native` flag
(`cc1: error: unknown value 'native' for '-march'`), and an unknown/cross target
arch emits no `-march` flag at all, letting the toolchain defaults / distro
`CFLAGS` decide.

This shipped in #1834 (PR #1858), which also adds tests for the
default-safe / opt-in-native behavior and the cross-compile path, and credits
you in the changelog and commit. Closing this PR in favor of that one — thank
you for the report and the groundwork!

PR #1838, opened 2026-01-07
Autobahn WebSocket Protocol - Improve typing

bblommers:
## Description

Adds typehints to `autobahn/twisted/websocket.py` and `autobahn/websocket/protocol`. I realize that's only a small section of the codebase, those just contain the classes that I use the most :slightly_smiling_face: 

Some style choices that I made:
 - `X | Y` notation (instead of the `Optional[X]` that is used as well in this codebase)
- Lowercase `dict[..]` and `tuple[..]`, as that is the preferred way for Python >= 3.10 (instead of `typing.Dict[..]` and `typing.Tuple[..]`)
- Removed the `:style..` from the docstrings, if the type is now available as a typehint, just because it seems superfluous to specify the type twice

## Checklist

      updated the changelog
      in this PR

### DIFF
--- a/justfile
+++ b/justfile
@@ -702,17 +702,15 @@ check-typing venv="": (install venv)
         --ignore unresolved-attribute \
         --ignore unresolved-reference \
         --ignore possibly-missing-attribute \
-        --ignore possibly-missing-import \
         --ignore call-non-callable \
         --ignore invalid-assignment \
         --ignore invalid-argument-type \
-        --ignore invalid-return-type \
         --ignore invalid-method-override \
         --ignore invalid-type-form \
         --ignore unsupported-operator \
         --ignore too-many-positional-arguments \
         --ignore unknown-argument \
-        --ignore non-subscriptable \
+        --ignore not-subscriptable \
         --ignore not-iterable \
         --ignore no-matching-overload \
         --ignore conflicting-declarations \
--- a/.audit/bblommers_websocket-server-typing.md
+++ b/.audit/bblommers_websocket-server-typing.md
@@ -4,5 +4,5 @@
 
 Submitted by: @bblommers
 Date: 2026-01-07
-Related issue(s): 
+Related issue(s): #1839
 Branch: bblommers:websocket-server-typing
--- a/.audit/bblommers_websocket-server-typing.md
+++ b/.audit/bblommers_websocket-server-typing.md
@@ -1,0 +1,8 @@
+- [x] I did **not** use any AI-assistance tools to help create this pull request.
+- [ ] I **did** use AI-assistance tools to *help* create this pull request.
+- [x] I have read, understood and followed the projects' [AI Policy](https://github.com/crossbario/autobahn-python/blob/main/AI_POLICY.md) when creating code, documentation etc. for this pull request.
+
+Submitted by: @bblommers
+Date: 2026-01-07
+Related issue(s): 
+Branch: bblommers:websocket-server-typing
--- a/src/autobahn/twisted/websocket.py
+++ b/src/autobahn/twisted/websocket.py
@@ -24,8 +24,10 @@
 #
 ###############################################################################
 
+from __future__ import annotations
+
 from base64 import b64decode, b64encode
-from typing import Any, Optional
+from typing import Any
 
 import txaio
 from zope.interface import implementer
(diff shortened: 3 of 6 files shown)


justfile
```diff
@@ -702,17 +702,15 @@ check-typing venv="": (install venv)
         --ignore unresolved-attribute \
         --ignore unresolved-reference \
         --ignore possibly-missing-attribute \
-        --ignore possibly-missing-import \
         --ignore call-non-callable \
         --ignore invalid-assignment \
         --ignore invalid-argument-type \
-        --ignore invalid-return-type \
```
bblommers, line 709: With the changes in `autobahn/wamp`, this check now passes

justfile
```diff
@@ -702,17 +702,15 @@ check-typing venv="": (install venv)
...
         --ignore invalid-method-override \
         --ignore invalid-type-form \
         --ignore unsupported-operator \
         --ignore too-many-positional-arguments \
         --ignore unknown-argument \
-        --ignore non-subscriptable \
+        --ignore not-subscriptable \
```
bblommers, line 713: Minor typo to get the CI to pass

oberstet:
thanks a lot for contributing!

in fact, this is funny/surprising, since adding/improving type hints is something I became interested in myself just recently, coming from a different angle (SLSA L4 / WASM).

this broader context and resulting "style guide" as related to type hints is now documented in https://github.com/crossbario/autobahn-python/issues/1839

I've added specific references to your PR, and I'd be very curious about your comments, remarks and general feedback!

please don't get this wrong, I am _not_ asking for "full type hints everywhere" for _this_ PR. 

the GitHub issue is just for giving the "outer scope / aggregate issue" as an anchor - eg I noticed you did add an audit file to this PR already (very good, thanks a lot!), but that does not yet refer a GitHub issue - because there wasn't one (which is important for the audit file as this properly relates "issue to PR to CI/CD to merge-to-master" and allows developers and users to figure out how these related also retrospectively)

bblommers:
Great minds think alike @oberstet!

I've pushed a new commit with some improvements, as pointed out in the guide:
 1. Add from __future__ import annotations to all modified files
 2. Replace all Optional[X] with X | None
 3. Remove unused typing imports (Optional, Union, Dict, Tuple)

Missing:
 - Add missing parameter types (especially payload: bytes)
 - Run ruff check --select ANN,UP and fix violations

There are still a ton of types missing, even in this small subsection - and I'm not familiar enough with the codebase/functionality to determine the correct types for them. That's why I left some as-is.

### General remarks / feedback

As someone who went through the same process of 'cleaning' a codebase both in terms of style and typehints, I have some advice, if I can:

1. Tackle it piece by piece. 
    a) `ruff` can be very useful, both to find improvements (`ruff check`) and sometimes to fix them automatically (`ruff check --fix`).
    b) At the same time - I found it a very useful to also apply these changes to one file/module at the time. Tackling everything is a great way to introduce new bugs/typos (speaking from experience :slightly_smiling_face: )
2. The style guide looks great! Very comprehensive and clear. However: automated checks will always be easier to follow for developers, so adding any `ruff` commands to `just check-typing` would  be very helpful for contributors


Speaking of: running `ruff check --select UP` seems to be mostly f-strings violations, which could be fixed automatically using `ruff check --fix`. I would be happy to add that as part of this PR - but I would recommend running that against the entire codebase in a separate PR for that. It makes reviewing a lot easier!

oberstet:
Thanks so much @bblommers — merging this now! 🎉

Really appreciated how quickly you aligned with the style guide after I filed #1839. The `from __future__ import annotations` addition and the `Optional → X | None` migration are exactly what we needed. Clean work.

On your feedback:

1. Piece by piece — 100% agree. That's the plan for #1839. Your PR sets a great pattern for what "typed subset" looks like in practice.
2. Adding ruff to just check-typing — great suggestion, I'll add ruff check --select ANN,UP to the recipe. Makes it much easier for contributors to validate locally before pushing.
3. f-strings cleanup in separate PR — yes please! If you're up for it, a focused ruff check --fix --select UP across the codebase would be very welcome. Keeping it separate makes review much easier, as you noted.

Remaining typing work (untyped payload params, more modules) is now tracked in #1839 — no pressure, but if you're interested in picking up more files, you'd be very welcome to. The high-priority ones are `wamp/protocol.py` and `wamp/component.py`.

Thanks again for the contribution and the thoughtful process feedback!

Issue #1841, opened 2026-01-09
[CHORE] Modernize Python syntax with pyupgrade (ruff UP rules)


oberstet:
## Summary

Run `ruff check --fix --select UP` across the codebase to modernize Python syntax automatically. This is a low-risk, high-value cleanup that improves code consistency and readability.

Suggested by @bblommers in #1838:
> "Speaking of: running `ruff check --select UP` seems to be mostly f-strings violations, which could be fixed automatically using `ruff check --fix`. I would be happy to add that as part of this PR - but I would recommend running that against the entire codebase in a separate PR for that. It makes reviewing a lot easier!"

## What pyupgrade (UP) Does

The `UP` rule set modernizes Python syntax to use newer, cleaner constructs. Key transformations include:

### String Formatting → f-strings

```python
# Before (UP031, UP032)
"Hello, %s" % name
"Hello, {}".format(name)

# After
f"Hello, {name}"
```

### Type Annotations → Modern Syntax

```python
# Before (UP006, UP007)
from typing import Dict, List, Optional, Union
x: Optional[str]
y: Union[int, str]
z: Dict[str, List[int]]

# After
x: str | None
y: int | str
z: dict[str, list[int]]
```

### Other Modernizations

| Rule | Before | After |
|------|--------|-------|
| UP005 | `assertEquals()` | `assertEqual()` |
| UP008 | `super(Foo, self)` | `super()` |
| UP009 | `# -*- coding: utf-8 -*-` | (removed, UTF-8 is default) |
| UP010 | `from __future__ import print_function` | (removed) |
| UP015 | `open("f", "r")` | `open("f")` |
| UP018 | `str("literal")` | `"literal"` |
| UP025 | `"unicode"` encoding | (simplified) |
| UP028 | `yield` in `for` loop | `yield from` |
| UP034 | `(x,)` extraneous parens | `x,` |
| UP035 | `typing.Dict` | `dict` |
| UP036 | Version checks for old Python | (removed) |
| UP038 | `isinstance(x, (A, B))` | `isinstance(x, A \| B)` |

### Step 1: Preview Changes

```bash
# See what would change (dry run)
ruff check --select UP src/autobahn/
```

### Step 2: Auto-fix

```bash
# Apply fixes automatically
ruff check --fix --select UP src/autobahn/
```

### Step 3: Manual Review

Some changes may need manual attention:
- Complex string formatting with many arguments
- Edge cases in type annotations
[...]

PR #1843, opened 2026-01-09
Chore: Conform src/autobahn to ruff's UP rules

bblommers:
## Description

Refactors the `src/autobahn` folder to conform to ruff's UP rules.

Most changes were done automatically using `ruff check --fix --unsafe-fixes --select UP src/autobahn/`, with some manual intervention to adjust the remaining ones.

---

## Related Issue(s)

Closes or relates to #1841 

---

## Checklist

      updated the changelog
      in this PR

### DIFF
--- a/src/autobahn/wamp/interfaces.py
+++ b/src/autobahn/wamp/interfaces.py
@@ -24,11 +24,12 @@
 #
 ###############################################################################
 
-import abc
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from __future__ import annotations
 
-# FIXME: see ISecurityModule.__iter__
-# from collections.abc import Iterator
+import abc
+from typing import Any
+from collections.abc import Callable
+
 from autobahn.util import public
 from autobahn.wamp.message import Message, Welcome
 from autobahn.wamp.types import (
@@ -99,7 +100,7 @@ class IObjectSerializer(abc.ABC):
 
     @public
     @abc.abstractmethod
-    def unserialize(self, payload: bytes) -> List[Any]:
+    def unserialize(self, payload: bytes) -> list[Any]:
         """
         Deserialize objects from a byte string.
 
@@ -118,7 +119,7 @@ class ISerializer(abc.ABC):
     @public
     @property
     @abc.abstractmethod
-    def MESSAGE_TYPE_MAP(self) -> Dict[int, "IMessage"]:
+    def MESSAGE_TYPE_MAP(self) -> dict[int, IMessage]:
         """
         Mapping of WAMP message type codes to WAMP message classes.
         """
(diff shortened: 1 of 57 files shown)


src/autobahn/wamp/interfaces.py
```diff
@@ -25,10 +25,9 @@
 ###############################################################################
 
 import abc
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any
+from collections.abc import Callable
 
-# FIXME: see ISecurityModule.__iter__
```
bblommers, line 30: There already was a FIXME in `ISecurityModule.__iter__`, so it seemed unnecessary to have it here as well

oberstet:
as a "note to self", cleaning up my mind, Claude to the rescue (bonus: I learned a new english word, "vestigial";)

> before I merge PR #1843, just one more question rgd "3. Class inheritance — class Foo(object): → class Foo:" => can you explain please? why was that needed in the _past_ anyways? was it related to twisted? I simply forgot "why" and just stuck to it (always derive from object for a final base class), and I want to "clean up my mind properly" so I'd like to understand the whole train of thoughts ;) 

● Great question! This is Python 2 vs Python 3 history.

  Python 2: Two Types of Classes

  In Python 2, there were two fundamentally different class types:

```python
  # Old-style class (classic class) - Python 2
  class Foo:
      pass

  # New-style class - Python 2.2+
  class Bar(object):
      pass
```

  Why New-Style Classes Mattered

  | Feature                                   | Old-style                  | New-style                                          |
  |-------------------------------------------|----------------------------|----------------------------------------------------|
  | Method Resolution Order (MRO)             | Depth-first, left-to-right | C3 linearization (correct for diamond inheritance) |
  | super()                                   | Broken/unavailable         | Works correctly                                    |
  | Descriptors (@property, __get__, __set__) | No                         | Yes                                                |
  | __slots__                                 | No                         | Yes                                                |
  | __getattribute__                          | No                         | Yes                                                |
  | type(instance)                            | Returns <type 'instance'>  | Returns actual class                               |

  The Diamond Problem

```python
  class A: pass        # old-style
  class B(A): pass
  class C(A): pass
  class D(B, C): pass  # MRO: D → B → A → C → A (A visited twice, wrong!)

  class A(object): pass  # new-style
  class B(A): pass
  class C(A): pass
  class D(B, C): pass    # MRO: D → B → C → A (correct C3 linearization)
```

  Twisted Connection

  Yes, Twisted specifically required new-style classes because:

  1. zope.interface (used by Twisted for implements()) required new-style classes
  2. Twisted's Deferred and protocol classes used descriptors
  3. Proper super() was needed for cooperative multiple inheritance

  So for any Twisted-based codebase (like Autobahn), class Foo(object): was mandatory for Python 2 compatibility.

  Python 3: All Classes Are New-Style

  In Python 3, the distinction was removed. All classes are new-style by default:

```python
  # Python 3 - these are 100% identical:
  class Foo:
      pass

  class Foo(object):
      pass
```

  The (object) is now purely vestigial noise. The ruff rule UP004 removes it.

  TL;DR

  - Python 2: (object) was required for correct MRO, super(), descriptors, and Twisted/zope.interface compatibility
  - Python 3: (object) does nothing — safe to remove

PR #1916, opened 2026-07-14
Fix 1909

### LINKED ISSUE #1909: [BUG] WebSocket `maxMessagePayloadSize` is enforced against compressed size, bypassed after permessage-deflate inflation
## Problem (GHSA-hxp9-w8x3-p566, reported by Team Atlanta)

On the WebSocket Hybi (v13) receive path, `maxMessagePayloadSize` is checked
against the **declared, compressed** frame length, *before* inflation, and the
inflated bytes are never re-counted. A client frame of ~22 compressed bytes can
inflate to 4096 bytes and reach `onMessage` even when the application configured
a 128-byte message limit.

Trace (`src/autobahn/websocket/protocol.py`):
- `onFrameBegin` → `_onMessageFrameBegin(self.current_frame.length)` — the
  **compressed** wire frame length (`:1841`).
- `onMessageFrameBegin` → `message_data_total_length += length` and checks it
  against `maxMessagePayloadSize` (`:634`, `:636`) — i.e. against compressed bytes.
- `onFrameData` inflates at `:1861`; for v13, `onMessageFrameData` only appends
  to `frame_data` (`:667`) and **never adds the inflated length** to
  `message_data_total_length`. The reassembled message is joined (`:690`) and
  delivered via `_onMessage` (`:693`) with no uncompressed-size ceiling.

This contradicts the documented contract ("re-assembled payloads",
`docs/websocket/programming.rst:512`), the **send** path (which checks
`maxMessagePayloadSize` against the *uncompressed* message, `:2554`), and the
**RawSocket** transport (which checks the uncompressed serialized message). Hixie
(v0) already accounts uncompressed bytes correctly (`:657`). Same boundary class
as CVE-2016-10544.

## Severity

Config-gated denial-of-service (memory/CPU amplification, availability only; no
confidentiality/integrity impact). Requires **two explicit opt-ins**:
permessage-deflate enabled *and* `maxMessagePayloadSize` set (default `0` =
unlimited). Not exploitable in a default configuration.

## Fix (approach)

Per META decision D1/D2: `maxMessagePayloadSize` bounds the **uncompressed,
reassembled per-message** size. Enforce it at the **inflation site**
(`onFrameData`), backend-agnostic and working in both message and streaming mode:

- Account the **uncompressed** length per frame (`uncompressedLen`, already
  computed at `:1862`) into the running message total, and check it against
  `maxMessagePayloadSize` there — replacing reliance on the pre-inflation
  compressed-length accounting for the message cap.
- For deflate, use the bounded decompress primitive from child 01 so a single
  frame's inflation stops at the remaining budget (memory bounded *during*
  inflation, not merely detected after). For codecs without a native output cap,
  post-decompress accounting still enforces the cap; per-frame wire size stays
  bounded by `maxFramePayloadSize`. (Full multi-codec bounded-during-inflation =
  child 03.)
- Keep `maxFramePayloadSize` semantics (wire/per-frame) unchanged.
- On exceed, use the existing `_max_message_size_exceeded` / `MESSAGE_TOO_BIG`
  path (`:998`), setting `wasMaxMessagePayloadSizeExceeded`.
- Preserve exact current behavior when `maxMessagePayloadSize == 0`.

Care: must not double-count (compressed accounting at `:634` vs new uncompressed
accounting), must keep Hixie (v0) path correct, and must not regress the
streaming/frame API.

## Red/green test plan (TDD, on this PR)

- **Workhorse unit test** modeled on the advisory PoC: drive `_dataReceived`
  directly (no network, deterministic), configure `maxMessagePayloadSize = N`,
  send a compressed frame that inflates to `> N`, assert the message is
  **rejected** (connection failed with `MESSAGE_TOO_BIG`,
  `wasMaxMessagePayloadSizeExceeded` true) and **not delivered** to `_onMessage`.
  Prove RED in CI (today it is delivered), then GREEN after the fix.
- **Matrix (parametrized):**
  - backends: Twisted **and** asyncio;
  - codecs: deflate (primary), plus snappy/bzip2/brotli where available
    (post-decompress enforcement must hold for all);
  - modes: whole-message **and** streaming (frame-based) consumer;
  - controls: uncompressed message just **under** N is delivered intact (no false
    positive); multi-frame fragmented message enforced on cumulative uncompressed
    total.
- **RawSocket regression:** a companion assertion that RawSocket continues to
  enforce `maxMessagePayloadSize` on the (uncompressed) serialized message, so the
  cross-transport contract is proven consistent.

## Acceptance criteria

- `maxMessagePayloadSize` bounds the uncompressed reassembled message on WebSocket
  receive, in both backends and both processing modes, for every compression
  backend.
- Advisory PoC no longer bypasses the limit; oversize → clean `MESSAGE_TOO_BIG`.
- Under-limit traffic unaffected; `maxMessagePayloadSize == 0` behavior identical
  to today.
- New parametrized test proven red-before / green-after in CI on this PR.

## References

- Advisory GHSA-hxp9-w8x3-p566. part of https://github.com/crossbario/autobahn-python/issues/1907. Depends on #TBD (draft 01).

_This work is being completed with AI assistance (Claude Code)._

oberstet:
fixes https://github.com/crossbario/autobahn-python/issues/1909 - this will be two-phase, proper "red green" TDD

### DIFF
--- a/.audit/oberstet_fix_1909.md
+++ b/.audit/oberstet_fix_1909.md
@@ -1,0 +1,8 @@
+- [ ] I did **not** use any AI-assistance tools to help create this pull request.
+- [x] I **did** use AI-assistance tools to *help* create this pull request.
+- [x] I have read, understood and followed the projects' [AI Policy](https://github.com/crossbario/autobahn-python/blob/main/AI_POLICY.md) when creating code, documentation etc. for this pull request.
+
+Submitted by: @oberstet
+Date: 2026-07-14
+Related issue(s): #1909
+Branch: oberstet:fix_1909
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,11 @@ Changelog
 26.7.1
 ------
 
+**Security**
+
+* Fix WebSocket ``maxMessagePayloadSize`` being enforced against the compressed on-the-wire frame length instead of the uncompressed reassembled message size when permessage-compress (deflate/bzip2/snappy/brotli) is negotiated. A small compressed frame could inflate far beyond the configured limit and be delivered to the application (a decompression-bomb style denial-of-service; security advisory GHSA-hxp9-w8x3-p566, same class as CVE-2016-10544). The limit is now re-checked at the inflation site against the running uncompressed message size, and the connection is failed with close code 1009 (message too big) before delivery — for both the whole-message and streaming receive APIs and every compression backend. Behaviour change: a compressed message that inflates past ``maxMessagePayloadSize`` is now rejected where it previously passed; uncompressed traffic and the per-frame ``maxFramePayloadSize`` wire guard are unaffected (#1909)
+* Fix the permessage-deflate ``max_message_size`` receive cap silently truncating an over-limit message and raising a zlib error instead of cleanly rejecting it: the bounded ``decompress(…, max_length)`` left the remaining input in ``unconsumed_tail`` undrained, so the message was corrupted rather than reported. Decompression is now bounded cumulatively across frames and raises ``PayloadExceededError`` as soon as the uncompressed size would exceed the cap (#1908)
+
 **FlatBuffers**
 
 * Fix ``check_zlmdb_flatbuffers_version_in_sync()`` comparing the build-time ``version()`` (which is ``(0, 0, 0, None, None)`` on installed wheels, where the vendored FlatBuffers ``__git_version__`` is unstamped) — it now compares the reliably-stamped ``__version__`` and returns a version string. Added regression tests (#1891)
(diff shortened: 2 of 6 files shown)


oberstet:
reviewed (by me), tested locally, and now TDD phase 1 is running, let's see it fail (what we do want):

image

oberstet:
there was an issue with phase 1 - it did fail, but for the wrong reason;) after pointing that out, an ack, and another iteration:

image

and after that 2nd commit, it now works - as in, it fails, but for the right reason:

image

---

however, there is still a thing missing / to fix for a perfect TDD red-green cycle:

image

---

so that's yet one more commit and run for the _right_ red! and then we can move on for the right green (the actual fix). I want to do this TDD properly;)

oberstet:
very good! test results are now exactly as intended:

https://github.com/crossbario/autobahn-python/actions/runs/29343192492/job/87121506560?pr=1916

here is in-depth results archive review:

image

Issue #1930, opened 2026-07-15
[FEATURE] Release Autobahn|Python 26.7.1


oberstet:
Part of the coordinated 26.7.1 release train (autobahn -> zlmdb -> crossbar)

---

# Release Autobahn|Python 26.7.1

## Why

Cut the 26.7.1 stable release. Master carries the completed WebSocket/RawSocket
payload size-limit **security epic** (#1907) plus the musllinux wheel track, none
of which has reached users yet — the tree is at `26.7.1.dev1`.

The musllinux work in particular is **merged but unpublished**: the user who
asked for Alpine wheels (#1877) gets nothing until this release ships.

## What's in it

**Security** (epic #1907, advisory GHSA-hxp9-w8x3-p566):

- **#1908** — permessage-deflate `max_message_size` cap silently truncated an
  over-limit message and raised a zlib error instead of rejecting it cleanly.
- **#1909** — `maxMessagePayloadSize` was enforced against the *compressed* wire
  length, so a small compressed frame could inflate past the limit and reach the
  application (decompression-bomb DoS; same class as CVE-2016-10544). Now enforced
  against the **uncompressed reassembled** size, failing with close code 1009.
  **Behaviour change**: such messages are now rejected where they previously passed.
- **#1910** — bounded decompression made backend-agnostic (`max_output_len`), so
  snappy/bzip2/brotli reach parity with deflate and a frame no longer inflates
  unbounded into memory before the size check.
- **#1911** — asyncio RawSocket receive limit made configurable
  (`setProtocolOptions(maxMessagePayloadSize=...)`), at parity with Twisted.

**Wheels / build**: musllinux (Alpine/musl) CPython wheels (#1877), aarch64 cp314
GIL-ABI fix (#1875), `.cicd` ABI-tag matching (wamp-cicd #11), ruff ANN/UP/TCH
gate (#1840), CalVer version recipes (#1894).

**FlatBuffers**: `version()` reliable on installed wheels (#1891).

## Steps

1. Branch `fix_<issue-number>` off master.
2. `just prep-release` — finalises `26.7.1.dev1` -> `26.7.1` (pyproject + `_version.py`).
   Do **not** hand-edit the version. (autobahn does **not** track `uv.lock`, so no relock.)
3. Review `docs/changelog.rst` 26.7.1 section is complete and accurate (it already
   carries the Security section for #1908/#1909/#1910/#1911).
4. PR -> normal CI cycle -> merge to master.
[...]

PR #1922, opened 2026-07-14
Fix 1911

oberstet:
fixes https://github.com/crossbario/autobahn-python/issues/1911 - will be done with proper red-green two-phase TDD

PR #1934, opened 2026-07-20
Skip -march guessing for external cross toolchains

leon-anavi:
## Description

Don't add -march flag when CC/CROSS_COMPILE names an external cross toolchain (Yocto/OpenEmbedded, Buildroot, ...). These already inject the correct target. Avoid issues such as an unknown value 'x86-64-v2' for '-march' when cross compiling for ARM machine.

---

## Related Issue(s)

Closes or relates to #

---

## Checklist

- [ ] I have referenced relevant issue numbers above
      updated the changelog
- [ ] I have added an AI assistance disclosure file (required!)
      in this PR

---

## AI Assistance Disclosure File

:warning: **Required for this PR:** You MUST include a disclosure
file at `.audit/<branch-name>.md` right in this PR. The
disclosure file must **follow the exact format and content** as
described below. _Your PR will not be accepted without a
disclosure file._

**Example 1** file contents of your disclosure file
`.audit/<branch-name>.md`:

```
## AI Assistance Disclosure

- [ ] I did **not** use any AI-assistance tools to help create this pull request.
- [x] I **did** use AI-assistance tools to *help* create this pull request.
- [x] I have read, understood and followed the projects' [AI Policy](https://github.com/crossbario/autobahn-python/blob/main/AI_POLICY.md) when creating code, documentation etc. for this pull request.

Submitted by: @your-github-username
Date: YYYY-MM-DD
Related issue(s): #issue-number
Branch: branch-name
```

**OR**

**Example 2** file contents of your disclosure file
`.audit/<branch-name>.md`:

```
## AI Assistance Disclosure

- [x] I did **not** use any AI-assistance tools to help create this pull request.
- [ ] I **did** use AI-assistance tools to *help* create this pull request.
- [x] I have read, understood and followed the projects' [AI Policy](https://github.com/crossbario/autobahn-python/blob/main/AI_POLICY.md) when creating code, documentation etc. for this pull request.

Submitted by: @your-github-username
Date: YYYY-MM-DD
Related issue(s): #issue-number
Branch: branch-name
```

---

Submitting code generated _primarily_ by AI, or for which you
cannot claim _human authorship_, is not permitted. See
[AI Policy](https://github.com/crossbario/autobahn-python/blob/main/AI_POLICY.md)
for details.

**Example 1 OR Example 2** show the only valid two variants. You
cannot have both or none of the first two marks checked, and you
must always have the last tick checked.

Well, "must" if you want your PR to be accepted and ultimately
merged that is. Of course you are always free to
`Go ahead! Fork my Day. (TM)` ;) This is Open-source.


oberstet:
Thanks for this! To be clear up front: **the problem you hit is real**, and the direction you took (don't second-guess an externally managed cross toolchain, which already injects its own `-march`/`-mtune`/`-mcpu`) is one I agree with.

I'm closing this PR, but not because the report (implied in the PR) is wrong — because reviewing it showed we can't yet say *what* it fixes. Let me explain, since I'd very much like to see this land properly eventually.

**On the referenced issue.** The audit file references #1930, which is the 26.7.1 *release* issue — already closed, and not a report of this problem. Our policy is that every change starts from an *open* issue describing the behaviour we consider wrong, so there's something concrete to agree on before code is written. (We're currently tightening this: the audit-file format and the open-issue requirement will be enforced in CI, and the contribution guide is being standardized across the WAMP Python projects. Our fault that this wasn't clearer — the guidance is genuinely inconsistent across our repos right now.)

**On what we actually promise.** Reviewing this raised a question I couldn't answer: **Does autobahn promise anything about cross-compilation?** It turns out the only statements are in a changelog entry and a docstring in `src/autobahn/nvx/_compile_args.py` — the latter claiming cross-compilation "works out of the box", with **no CI job that cross-compiles anything** behind it. That's an aspiration published as documentation, and it makes it impossible to tell whether a change like yours fixes a deviation from promised behaviour or introduces new behaviour.

So we've filed #1935 to write the promise down properly — in `README.md` and `docs/installation.rst`, with explicit support-status tags, and with a real cross-build CI target (**riscv64 via Buildroot, cross-built on x86-64**) to back it. Companion issues exist for zlmdb (crossbario/zlmdb#145) and Crossbar.io (crossbario/crossbar#2280).

**How we'd want the fix to land**, once the promise exists — we follow a two-phase test-driven workflow:

1. a test that **demonstrates the bug**, pushed first, and shown failing in CI (the "red" phase, documented in a PR comment linking the failing job);
2. the fix, making that same test pass in CI (the "green" phase, likewise documented).

That way the PR proves both that the behaviour was wrong *and* that it's now right, and the test stays in the suite so we don't regress. `test_compile_args.py` is the natural home for the unit-level part. As a real-world example, you can look at https://github.com/crossbario/autobahn-python/pull/1922 and the comments there.

**One thing that would really help**, and which only you can provide. I don't think the root cause is where this patch sits. If you got `-march=x86-64-v2` while targeting ARM, that means `sysconfig.get_platform()` returned the **host** (`linux-x86_64`) in your build — but that function is precisely the mechanism we rely on to detect the *target*. So the guarantee broke one layer below where this patch applies, and the same detection failure would affect anything else deriving the target from `sysconfig`.

Could you open an issue with:

- the output of `python3 -c "import sysconfig; print(sysconfig.get_platform())"` in your build environment;
- whether your recipe exports `_PYTHON_SYSCONFIGDATA_NAME` (Buildroot sets this so `sysconfig` reports the target; if OE isn't setting it in your setup, that would explain the whole thing);
- the `CC` / `CROSS_COMPILE` values in effect.

There's also a design question your patch answers implicitly and which we should settle explicitly: **what is authoritative about the target architecture** — `AUTOBAHN_ARCH_TARGET` > a cross-toolchain signal (`CC`/`CROSS_COMPILE`) >
`sysconfig`? That's recorded in #1935, and your input on it would be welcome given you're closer to the Yocto/OE side than we are.

Two smaller notes on the patch itself, for whenever it's revived:

- `if "-" in exe` also matches common *native* compilers — `gcc-13`, `clang-18`, Debian's `x86_64-linux-gnu-gcc` — so those builds would silently lose the `-march=x86-64-v2` baseline with no warning. It also *misses* wrapped cross
  toolchains like `CC="ccache aarch64-poky-linux-gcc"`.
- The check sits before the safe/default branch, so an explicit `AUTOBAHN_ARCH_TARGET=safe` would be silently overridden by autodetection. Explicit user intent should win.

Thanks again for taking the time — this genuinely improved things on our side, just not in the shape of a merge. If you'd like to pick it back up after #1935 lands, that would be very welcome.

Issue #1935, opened 2026-07-21
[DOCS] Write down the cross-compilation promise (README + installation docs), with explicit support-status tags


oberstet:
See also:

* https://github.com/crossbario/zlmdb/issues/145
* https://github.com/crossbario/crossbar/issues/2280

----

## Why

PR [#1934](https://github.com/crossbario/autobahn-python/pull/1934) proposed changing NVX compiler-flag selection for external cross toolchains (Yocto/OpenEmbedded, Buildroot). Reviewing it surfaced a more basic problem:

**we cannot say whether that PR fixes a deviation from promised behaviour, because we never wrote the promise down.**

Where cross-compilation is mentioned today:

- `docs/changelog.rst:70` (#1834/#1835) — *"The default architecture target is now the portable baseline for all build contexts (wheels, local source   installs, and cross-compilation). The target architecture is detected via   `sysconfig.get_platform()` so the correct baseline is chosen when cross-compiling."*
- `src/autobahn/nvx/_compile_args.py` module docstring — *"never hands a cross-compilation toolchain the host-only `-march=native` flag"* and, at  `:238-240`, *"Defaulting to 'safe' rather than `-march=native` is **what makes cross-compilation work out of the box**"*. This **is** published — via sphinx-autoapi, despite the leading underscore.

Where it is **not** mentioned: `README.md`, `docs/installation.rst`, or anywhere else a user or integrator would look. Grepping the docs tree returns hits only in the changelog, `_compile_args.py` and its test.

So an integrator has to discover our cross-build behaviour by hitting a compiler error. Meanwhile the docstring makes the strongest claim we have — *"works out of the box"* — with **no CI job that cross-compiles anything** behind it. That claim is aspiration, not fact, and it is published as documentation.

This issue is about fixing the *promise*, not the code.

## Support-status vocabulary (to be used identically in autobahn, zlmdb, crossbar)

- **Supported** — named targets, gated in CI: Tier 1 per-commit, Tier 2 nightly.
- **Best-effort** — expected to work, not gated in CI; bug reports welcome.
- **Roadmap** — intended direction, **not yet backed by tests**.

## Test tiers (vocabulary, for the docs; implementation is out of scope here)

- **Tier 1 — cross-build ("baking")**: the native extension cross-compiles for the target and the artifact is verified (e.g. `readelf -h` → expected machine, plus an import/functional smoke test of the built module). Fast; intended to
  run on every commit.
[...]

PR #1937, opened 2026-09-21
Add Windows ARM64 build

### LINKED ISSUE #1936: [FEATURE] Add Windows ARM64 wheels
## Feature Description

**Summary**: I would like to have native Windows ARM64 wheels available.

**Problem Statement**:
It's currently not easy to install autobahn on Windows ARM64 -- it builds from source, and that doesn't succeed unless you have several toolchains installed (Rust and MSVC.)

**Proposed Solution**:
In #1937.

## Use Cases

**Use Case 1**:
- **Actor**: Users who want to use autobahn natively on Windows ARM64 machines.
- **Goal**: (Easily) install and use autobahn using pip/uv.

## Alternatives Considered

**Alternative 1**: Install from source (sdist).
- Pros: Works.
- Cons: Slow process, requires the correct toolchains installed and configured.

**Why the proposed solution is better**: It's not easy to figure out the right toolchains and dependencies that are needed for an sdist install.

## Impact Assessment

**Breaking Changes**: No

**Affected Components**:
- [ ] Core library
- [ ] API
- [ ] Documentation
- [ ] Tests
- [x] CI/CD

**Estimated Complexity**: Low

## Additional Context

**Related Issues**:
- None

**External References**:
- None

**Mockups/Diagrams**:
N/A

## Checklist

- [x] I have searched existing issues to avoid duplicates
- [x] I have described the problem clearly
- [x] I have provided use cases
- [x] I have considered alternatives
- [x] I have assessed impact and breaking changes

ndabas:
## Description

As the title says. I have added a few extra checks to make sure that the correct wheels are built, because Windows ARM64 will silently emulate AMD64, and tools like `uv` currently default to installing the emulated Python on WoA rather than the native one.

There is also a temporary step that I added specifically for Windows ARM64, to install a native OpenSSL build -- this is so that the cryptography package (which is a transitive dependency) can be built from source, as they currently have no win_arm64 wheel published. That wheel should be coming sometime soon, as [the PR has already been merged](https://github.com/pyca/cryptography/pull/15350). We can remove that step when that happens.

---

## Related Issue(s)

Closes #1936.

---

## Checklist

- [x] I have added new or used existing tests that prove my fix
- [x] I have added necessary documentation (if appropriate) and
      updated the changelog
      in this PR


oberstet:
Thanks for this, @ndabas — and welcome! This is a well-scoped first contribution, and you've correctly identified the two subtle traps that make Windows-on-ARM builds go wrong, which most people miss:

- the **silent x64 emulation** on WoA (and `uv` defaulting to the emulated interpreter), and
- **cryptography** having no `win_arm64` wheel yet, forcing a source build (Rust + OpenSSL).

A few things I'd like to confirm or tighten already at this point — mostly about making the native-ARM64 guarantee *enforced* rather than best-effort, since the failure mode here is silent (mis-tagged wheels), which is the worst kind IMO:

**1. Make the native-arch check a hard failure.** The emulation guard is the crux of the whole PR. Please have the build assert the interpreter's `platform.machine()` is ARM64 — print it in the job log and fail loudly if it's AMD64/emulated — so a future runner-image change can never silently produce `amd64` bytes inside a `win_arm64` wheel. A green build that quietly built the wrong thing is the nightmare case.

**2. Confirm the wheels are actually *tested* on the native runner, not just built.** Building isn't proof. Please run at least an import smoke test on `windows-11-arm` that exercises the real moving parts — `autobahn`, the **NVX cffi accelerator**, `cryptography`, `txaio` — and ideally a slice of the suite. NVX compiling under MSVC/ARM64 and then actually loading is the thing most likely to break, and a pure build step won't catch a bad extension.

**3. Scope: CPython-only.** PyPy has no Windows/ARM64 build, so the `win_arm64` matrix should be CPython-only — please make sure nothing tries a PyPy `win_arm64` wheel.

**4. Track the temporary OpenSSL/cryptography step for removal.** Good that you flagged it as temporary — let's not lose it: gate it on the `cryptography` floor in `pyproject.toml`, and open a small follow-up issue ("remove win_arm64 OpenSSL bootstrap once the cryptography floor ships win_arm64 wheels", referencing the upstream PR) so the tech-debt is tracked rather than forgotten. If what currently resolves already ships the `win_arm64` wheel, the step may even be droppable now — worth checking what actually resolves.

**5. Update the stale note + the platform docs.** `wheels.yml` still carries a comment that *"GitHub does NOT provide Windows ARM64 hosted runners"* — this PR obsoletes it, so please remove/replace it (otherwise the file contradicts itself). And `.github/workflows/README.md` enumerates the full platform / arch / interpreter matrix — please add the `win_arm64` row there too, or the docs drift immediately.

**6. Wheel tag + repair.** Please confirm the emitted wheel is tagged `*-win_arm64` (not `win_amd64`), and that any wheel-repair/delocate step in the Windows path handles ARM64 correctly.

**7. Where it sits in our wheel tiers.** We're formalizing supported-vs-best-effort wheel tiers (#1935). Given a brand-new platform + the temporary crypto bootstrap + runner availability, I'd slot `win_arm64` as **best-effort (Tier 2)** for now — built and smoke-tested, not release-gating — and note that in the docs. Happy to place it once the above is settled.

CI is still running as I write this — I'll do a final pass once it's green, and in particular I want to see in the logs that the `win_arm64` job landed on a genuine ARM64 runner (per #1). Really nice first PR; the platform instincts here are exactly right.

(And thanks for including the AI-assistance disclosure per our `AI_POLICY.md` — appreciated.)

oberstet:
@ndabas — quick heads-up on the one red check: **"Code Quality Checks" is on us, not you.**

That job runs `ty` (we deliberately float it on the bleeding edge), and a recent `ty` update started flagging three pieces of **pre-existing dead code** in autobahn — two unused `_unique_list` helpers and one unreachable branch in the WebSocket opening-handshake path. None of it is related to your Windows ARM64 changes; your PR just happened to be the run that surfaced it.

I've opened #1938  to remove that dead code on `master` in a separate, fully-tested PR (keeping your PR focused on the wheels). Once that lands, please rebase on `master` — or I'll re-run CI — and this check goes green. Nothing for you to change here.

Thanks again for the contribution, and apologies for the red ✗ that isn't yours.

ndabas:
Thanks for the review, and the clarification on the failing CI job, @oberstet!

I don't think I've changed anything that should touch or cause the "wheels-arm64 / Build ARM64 wheels (pypy-3.11-bookworm-manylinux_2_36_aarch64" job to fail, but I'll double-check.

On the points you have noted, 3 and 5 are already done. For the rest I'll verify and add whatever is missing and let you know.

ndabas:
Thanks again for the great review. On the remaining points:

1. I already added a guard in `justfile`. It asserts on `sysconfig.get_platform()`, which is the actual value that determines the wheel tag. `platform.machine()` returns the actual CPU instead, even when running emulated, so using that would mean that we see `ARM64` as expected but the wheels would still end up being x86_64 when run with an emulated Python. I'll also add some comments and debug prints around this to make this clear in case someone feels like 'fixing' it later.
2. I'm adding the smoke tests now, and I see there is no existing smoke test for the built wheels on any platform currently in the CI. **So I'm looking at writing the tests in Bash so they run on all platforms (including Windows of course) that we build wheels for. Does that sound okay?**
3. I already added a condition in `justfile` so we don't try to build PyPy win_arm64 wheels at all.
4. As `cryptography` is not currently shipping a win_arm64 wheel, we cannot add a version floor because we don't know what they'll tag that release with. **I'll open an issue though to keep track of it.**
5. I updated that note and added platform docs already.
6. The Windows workflow (any platform) isn't currently doing any wheel repair, so there's nothing to do at the moment. If we ever add `delvewheel` for that purpose, it handles Windows ARM64 fine.
7. I have added `cpyNNN-win-arm64` to all three check-release-fileset target lists, **let me know if you would like me to remove those** in order to set this as a "Tier 2" target for now.

On the other failing job ("Build ARM64 wheels (pypy-3.11-bookworm-manylinux_2_36_aarch64)"), it looks like the current release of `auditwheel` requires a `patchelf` greater than what is already installed in the image, so the simple fix would be to `pip install patchelf auditwheel` instead of just installing `auditwheel` in `docker\Dockerfile.pypy-bookworm-manylinux-arm64`. **Not really part of this PR but I can open another PR for that if you'd like.**

oberstet:
Thanks @ndabas — these are great responses; you addressed every point, and a couple are better than what I asked for. Where I've landed:

**Please rebase on `master` first.** Two of the red checks are pre-existing issues we've since fixed on `master`, and rebasing clears them:
- **Code Quality (`ty`)** — fixed in #1939.
- **ARM64-Linux `auditwheel` / `patchelf`** — fixed in #1942. That's the *same* "unrelated Docker/auditwheel bug" you spotted in your bonus note — already resolved, so **no separate PR needed**; the rebase picks it up. Good catch, and sorry it cost you the investigation.

On the individual points:

1. **Native-arch check** — good call using `sysconfig.get_platform()` over `platform.machine()`; it's exactly the wheel-tag determinant, so an emulated amd64 interpreter is caught. One request: make it a **hard failure** (non-zero exit when it isn't `win-arm64`), not just debug output — the failure mode here (amd64 bytes silently landing in a `win_arm64` wheel) is the kind that must fail loudly.
2. **Smoke tests** — yes please, and you don't need to invent a harness: reuse the existing recipes — `just test-wheel-install <wheel>` plus `test-import` / `test-smoke`. Extending those keeps it consistent with how the rest of the matrix validates wheels (and avoids a bespoke Bash path).
3. **CPython-only** — 👍 thanks.
4. **OpenSSL/cryptography step** — agreed; a tracking issue is the right call (no `win_arm64` cryptography wheel to floor against yet). Same shape as #1941 — link it there for context.
5. **Docs / stale note** — 👍
6. **Wheel repair** — fine; Windows wheels don't need it today, `delvewheel` later if that changes.
7. **Wheel tiers** — let's classify `win_arm64` as **best-effort (Tier 2)** for now, since it depends on the temporary OpenSSL source-build. I'll take it out of the *required* release-fileset so it can't gate releases; keep building + smoke-testing it, just non-gating until cryptography ships a `win_arm64` wheel and the bootstrap step goes away.

After the rebase it should run green — with **one known exception**: the emulated `pypy-3.11-bookworm-manylinux_2_36_aarch64` job still has a *transient* QEMU segfault (exit 139) that's unrelated to your change and tracked separately in #1940. If it trips, a re-run usually clears it; we're making that job non-blocking as part of #1940, so don't worry if that single check flakes.

Really nice contribution — thanks for the care on the WoA emulation traps.

ndabas:
Thank you for the clear responses -- I have updated the code to match.

I used the existing recipes for the smoke tests, and added a couple of tests there. Let me know if that looks okay. CI is all green on my fork now, thanks for pushing those fixes so quickly.

On the wheel tiers, the `win_arm64` wheels are in the fileset lists at the moment, I thought about removing those but the process seems to clean up any extra wheels it finds. So those would never get released then. I did see the `allow-extra-wheels: true` option but that might allow a stray wheel through. Let me know how you would like to handle this.

On a side note, we could consolidate much of the Windows-specific/other platform-specific code since most of those are doing exactly the same thing on every platform. Both `pwsh` and `bash` are available on all three platforms. Probably something for another PR if you want that although I could roll some of the trivial ones into this PR, let me know what you think.

oberstet:
thanks a lot! "windows on arm64" coming. `\-/ / ;)`

side question: are you only after autobahn-python, or [zlmdb](https://github.com/crossbario/zlmdb) or [crossbar](https://github.com/crossbario/crossbar) as well for win/arm64?

also, I took the chance to open issues to improve docs as well now: https://github.com/crossbario/autobahn-python/issues/1946 (and https://github.com/crossbario/zlmdb/issues/147, and https://github.com/crossbario/crossbar/issues/2287)

oberstet:
image

https://github.com/crossbario/autobahn-python/actions/runs/35770259383/job/106890015630

oberstet:
@ndabas : https://github.com/crossbario/autobahn-python/releases/tag/master-202609230557

this now includes win_arm64 wheels! for cpython. if you have a chance to actually try these on your side, would be curious to know! I don't have any win machine ...

image

ndabas:
Excellent, thanks @oberstet! I tested the nightly on my Windows ARM64 laptop, works fine -- here's the script I used to do that here, in case someone wants to use it (because we need that OpenSSL build for now until `cryptography` releases `win_arm64` wheels): https://gist.github.com/ndabas/27f9797963c775f6bf7534d6bde168e2

[zlmdb](https://github.com/crossbario/zlmdb) and [crossbar](https://github.com/crossbario/crossbar) aren't on my list at the moment, but I would be happy to contribute if you're looking to get Windows ARM64 support for those as well!

ndabas:
[zlmdb](https://github.com/crossbario/zlmdb) and [crossbar](https://github.com/crossbario/crossbar) aren't on my list at the moment, but I would be happy to contribute if you're looking to get Windows ARM64 support for those as well!

oberstet:
@ndabas — the native `win_arm64` wheels are building and shipping in the nightly `master-*` GitHub
release now, and they'll land on PyPI with the next tagged release. Really solid work — the tricky
Windows-on-ARM parts (the native-arch guard so WoA doesn't silently emulate x64, the temporary
OpenSSL/cryptography bootstrap, CPython-only scoping) were all handled just right.

And yes — absolutely, welcome aboard! I'd love to get Windows ARM64 support across the rest of the
WAMP stack, and your offer to help with **zlmdb** and **crossbar** is very much appreciated. Let's do it.

A bit of framing so we pick good targets:

- **zlmdb** is the natural next step: like autobahn it has a native component (vendored LMDB + a
  FlatBuffers runtime) and it already uses the same shared `wamp-cicd` wheel/CI scaffolding, so the
  `win_arm64` job should port over closely to what you just did here. A WoA wheel there gets us most of
  the way to a fully installable stack on Windows ARM64.
- **crossbar** is a bit different — it's a pure-Python application (Twisted), so it isn't about
  compiling crossbar itself but about making sure the *whole dependency stack* has WoA wheels so
  `pip install crossbar` "just works" on Windows ARM64. That's really autobahn (done) + zlmdb (next) +
  the upstream natives (`cryptography` etc.). Same Tier-2 caveat likely applies until those upstream
  wheels exist — worth tracking the same way you did here.

Process-wise, we work issue-first per repo (each change starts from a GitHub issue the PR references).
Happy to open tracking issues on **zlmdb** and **crossbar** for the Windows ARM64 work so you've got a
clear home to PR against — I'll cc you. If you'd rather scope them yourself first, that's great too;
just say which you'd like to start with.

Thanks again for #1937, and welcome to the WAMP crew. 🚀

ndabas:
Great, thanks again @oberstet! Glad to see things moving for Windows ARM64 users.

If you could open tracking issues in these projects and cc me -- that would be great, I'll take a look and scope out the work needed.

